PR #1, opened 2018-03-23
[WIP] Upload files to the origin server

dongsupark:
Allow different types of remote repos where the images are uploaded. The type can be either `"gs"` (Google Storage, which is default), or `"sftp"` (file server that is accessible via sftp).

Explicitly set `--upload_type=sftp` for Jenkins scripts, instead of uploading to Google Storage. Doing that, we can first upload images to the flatcar origin server, and later we can sync repos with Google Storage buckets using `sync-rclone`.

**Work-in-progress. completely untested**

### DIFF
--- a/bootstrap_sdk
+++ b/bootstrap_sdk
@@ -60,7 +60,7 @@ GENTOO_MIRRORS="${GENTOO_MIRRORS//https:\/\//http://}"
 export GENTOO_MIRRORS
 
 catalyst_init "$@"
-check_gsutil_opts
+check_upload_opts
 
 if [[ "$STAGES" =~ stage4 ]]; then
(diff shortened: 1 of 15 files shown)


iaguis:
I think the consensus we reached in the last meeting was to upload the files to both the origin server and the google cloud bucket during the Jenkins build so that we don't need gcloud credentials in the origin server too (am I remembering correctly, @schu?)

Maybe we can modify this a bit to allow a list for `--upload_type=`:

```
--upload_type=sftp,gs
```

schu:
> I think the consensus we reached in the last meeting was to upload the files to both the origin server and the google cloud bucket during the Jenkins build so that we don't need gcloud credentials in the origin server too (am I remembering correctly, @schu?)

Yes.

PR #2, opened 2018-03-28
[WIP] Upload files to a specified server

dongsupark:
Allow different types of remote repos where the images are uploaded. The type can be either `"gs"` (Google Storage, which is default), or `"sftp"` (file server that is accessible via sftp).

Explicitly set `--upload_type=sftp` by default for Jenkins scripts, instead of uploading to Google Storage. Doing that, we can first upload images to the flatcar origin server, and later we can sync repos with Google Storage buckets using `sync-rclone`.

**Work-in-progress. completely untested**


dongsupark:
outdated

PR #5, opened 2018-10-15
signing: add an edge channel

dongsupark:
Add a new channel `edge` for experimental features.


dongsupark:
Hmm, you're right. `transfer.sh` is not used at all. Maybe it's just leftover.
Will close.

PR #6, opened 2018-12-04
sdk_lib: download Flatcar SDK in a graceful way

dongsupark:
If the given Flatcar SDK tarball is not available on the remote server, it should fall back to the next recent version available.

Doing that, we can deal with occasional download failures that occur when the given SDK was not properly built for some reasons.


vbatts:
this seems stale. Is it still needed?

dongsupark:
It is nice to have, but I don't care about it at the moment.
Closing.

PR #9, opened 2019-02-04
build_library: enlarge rootfs size to 6G for dev container

dongsupark:
Since the developer container image is only 3G in size for its rootfs, it's hard for users to do specific work there, for example, to build kernel modules.

Let's increase the size of rootfs to 6G.


dongsupark:
Yesterday I tested it by building a dev container image. It worked fine.

PR #21, opened 2019-09-10
build_library: whitelist 201909-08 for D-BUS

dongsupark:
Since we already backported a security patch to sys-apps/dbus 1.10, we should whitelist GLSA 201909-08 for now.

This PR should be merged together with https://github.com/flatcar-linux/coreos-overlay/pull/74.


dongsupark:
Rebased

PR #23, opened 2019-10-07
Support for ARM architecture

dongsupark:
This PR consists of most changes to support ARM architecture for Flatcar Linux.

A bunch of changes are about reverting old commits that dropped ARM architecture, back in mid 2018.

There could be still rooms for improvements.
So please review, test, and let us know.

Should be merged together with https://github.com/flatcar-linux/bootengine/pull/4 https://github.com/flatcar-linux/coreos-overlay/pull/88 https://github.com/flatcar-linux/mantle/pull/28.


pothos:
About the rust build failure because gcc can't find limits.h: it seems that the glibc headers are either not installed (yet?) or can't be found.

pothos:
Fixed another issue where the talloc package wanted to run some compiled binaries during configuration to test the compiler. This needed qemu-aarch64 to be set up with the correct library path.

dongsupark:
Kai, your fixes look good, according to my local ARM builds.
Let's merge this PR next week, when the release waves are gone.

PR #25, opened 2019-10-17
setup_board, update_chroot: introduce --binhost= option

dongsupark:
Add an option `--binhost=`, to make the scripts download binary packages from any arbitrary URL. The given URL will be then prepended to the list of the default URLs. For example:

```
./setup_board --binhost=https://storage.googleapis.com/flatcar-jenkins/sdk/amd64/2000.0.0/pkgs
```

This is needed to work around toolchain issues when building specific packages especially for ARM64. Though it will be anyway useful for general use cases.


dongsupark:
It was already merged via other PRs.

PR #29, opened 2019-11-15
build_library: replace coreos with flatcar in disk layout

dongsupark:
To make GPT partition type strings include `flatcar-` strings instead of `coreos-`, we need to change the disk layout file.

This PR should be merged after https://github.com/flatcar-linux/seismograph/pull/1 was merged.

### DIFF
--- a/build_library/disk_layout.json
+++ b/build_library/disk_layout.json
@@ -26,7 +26,7 @@
       "3":{
         "label":"USR-A",
         "uuid":"7130c94a-213a-4e5a-8e26-6cce9662f132",
-        "type":"coreos-rootfs",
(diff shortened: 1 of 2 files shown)


dongsupark:
Also fixed `disk_util`.

dongsupark:
I have just realized, this PR should not be merged until the next Flatcar release could be released with https://github.com/flatcar-linux/seismograph/pull/1.
That's because the current build always depends on the previous version of SDK, which does not include the fix for Flatcar GPT strings yet. As a result, `cgpt -t flatcar*` would simply fail.
So let's do not merge this PR yet.

PR #34, opened 2019-12-03
Merge flatcar-master to flatcar-master-edge 2019-12-03

dongsupark:
Merge flatcar-master to flatcar-master-edge, to include fixes for cross-builds.


dongsupark:
Excluded GPT-related commits, to avoid conflicts.
So this PR includes only cross-builds fixes.

PR #41, opened 2019-12-19
build_library/check_root: ignore some shared libs

dongsupark:
To be able to suppress unnecessary warning messages, we need to ignore some shared libraries needed by additional packages that were recently added.

### DIFF
--- a/build_library/check_root
+++ b/build_library/check_root
@@ -100,7 +100,23 @@ IGNORE_MISSING = {
                                  SonameAtom("x86_64", "libterm.so")],
 
     "sys-kernel/coreos-modules": [SonameAtom("x86_64", "libc.so.6"),
-                                  SonameAtom("x86_64", "libcrypto.so.1.0.0")],
+                                  SonameAtom("x86_64", "libcrypto.so.1.0.0"),
+                                  SonameAtom("x86_64", "libyaml-0.so.2")],
+
+    "app-emulation/cri-o":       [SonameAtom("x86_64", "libc.so.6"),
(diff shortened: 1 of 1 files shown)


marga-kinvolk:
Could you explain a bit why this file is needed? It seems rather weird to have to list all the libraries that need to be ignored... What's the benefit of this?

dongsupark:
I am not the original author of the `check_root` script, but let me try to explain.

After the `build_packages` script finished building individual packages, it [checks](https://github.com/flatcar-linux/scripts/blob/9d0e58dbe8f8c35598bd86f10d8432e93cbd50ba/build_packages#L248) if the generated packages are all ok, for example, if no GLSA issue is left, or if individual binary files have correct shared libraries. If anything is wrong, `build_packages` can fail, printing out a bunch of messages.

However, there are exceptional cases, where such strict restrictions are not relevant, or just noise messages. That's the case especially when it is not a pure Gentoo distro, but its derivative distros like ChromiumOS/CoreOS/Flatcar. Inside a Flatcar SDK environment, there is a distro outside of `/build/amd64-usr`. The outer distro comes from a Flatcar SDK of the previous version. On the other hand, the inner distro, inside `/build/amd64-usr` has to follow the current versions of individual packages. So there are cases where the two distro trees do not match with each other. 

That's why CoreOS maintainers introduced a list of ignored libraries for each package. The list allows us to work around some painful library mismatch issues.

The list was originally added by the commit https://github.com/flatcar-linux/scripts/commit/5985b1e3d6c5bc0c34c8cb3a6909ef06cd9a606b.

dongsupark:
outdated

PR #49, opened 2020-01-22
Use same APPID for arm64 and amd64

pothos:
Nebraska supports an additional arch
variable to distinguish both boards.


dongsupark:
I have just realized, this PR was made against `flatcar-build-2387.0.0` instead of `flatcar-build-2387`.
So I applied the fix also to https://github.com/flatcar-linux/scripts/tree/flatcar-build-2387.

pothos:
My intention was that the sync script will pick it up for the next release. But yes, for point releases this should also be on the other branch.

PR #51, opened 2020-01-29
build_image: Add prodtar command to build a tar ball

pothos:
Create a tar ball with the contents of the / and /usr partitions
to be used as follows with systemd-nspawn (via machinectl):
  machinectl import-tar flatcar-container.tar.gz flatcar-container
  machinectl start flatcar-container
  machinectl shell flatcar-container
or with docker by converting it to an OCI image:
  docker import -c "CMD /bin/bash" flatcar-container.tar.gz flatcar-container

Since the new "prodtar" command relies on the results of the "prod" command,
it bundles it so that "prod prodtar" and "prodtar" is the same.

Note: Document how to use the tars on the website. Automatically push docker images to a repository.

### DIFF
--- a/build_image
+++ b/build_image
@@ -56,12 +56,13 @@ FLAGS_HELP="USAGE: build_image [flags] [list of images to build].
 This script is used to build a CoreOS image. CoreOS comes in many
 different forms.  This scripts can be used to build the following:
 
-prod - Production image for CoreOS. This image is for booting.
+prod - Production image for CoreOS. This image is for booting (default if no argument is given).
+prodtar - Production container tar ball (implies prod). This can e.g. be used to run the Flatcar production image as a container (run machinectl import-tar or docker import).
 container - Developer image with single filesystem, bootable by nspawn.
 
 Examples:
 
-build_image --board=<board> [prod] [container] - builds developer and production images.
+build_image --board=<board> [prod] [prodtar] [container] - builds developer and production images/tars.
 ...
 "
 show_help_if_requested "$@"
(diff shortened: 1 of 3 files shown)


dongsupark:
I think in general it is ok to add "tarball" to the list of supported format.

However, if we start calling it "prodcontainer", then it gets hard to distinguish it from the existing "developer container". Both can be used for containers, but prod is a tarball, while dev is an image including partitions etc.? Users could get confused.

If we want to introduce a real generic format for containers, ideally we should create an OCI image, by using tools like buildah or umoci. Though in reality we cannot say either that the standard OCI images are widely used.

So I am not sure what the best way is.

pothos:
Ok, I've renamed it.

build_image
```diff
@@ -56,12 +56,13 @@ FLAGS_HELP="USAGE: build_image [flags] [list of images to build].
 This script is used to build a CoreOS image. CoreOS comes in many
 different forms.  This scripts can be used to build the following:
 
-prod - Production image for CoreOS. This image is for booting.
+prod - Production image for CoreOS. This image is for booting (default if no argument is given).
+prodtar - Production container tar ball (implies prod). Usable by machinectl import-tar or with docker import.
```
t-lo: + "This can e.g. be used to run the Flatcar production image in a container."

pothos: Updated it, but changed it to *as a container* because the image is run as a container and not inside one.

pothos:
I have added a commit for Jenkins to produce this tar ball by default.

build_image
t-lo: Good catch, thank you :smile:

PR #54, opened 2020-03-18
build_library/vm_image_util.sh: Generate iPXE script

pothos:
Write out an iPXE script file for Packet.
    The script uses relative URLs to refer to
    the other PXE files and thus can be copied
    along with the files to any server.
    This is useful because it saves the creation
    of an iPXE script for a release/channel on a
    third-party service. For CI testing it is
    also helpful because the script does not only
    end up on the release server but also already
    on the Google buckets, refering to unpublished
    PXE payloads.

Note: Pick for alpha and edge.

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -768,6 +768,15 @@ You can pass extra kernel parameters with -append, for example:
 When using -nographic or -serial you must also enable the serial console:
   ./$(basename "${script}") -nographic -append 'console=ttyS0,115200n8'
 EOF
+    local packetipxe="$(_dst_dir)/flatcar_production_packet.ipxe"
+    cat > "$packetipxe" <<EOF
+#!ipxe
+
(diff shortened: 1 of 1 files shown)


pothos:
For testing I copied the file to the current ARM64 folder: https://alpha.release.flatcar-linux.net/arm64-usr/current/packet.ipxe

pothos:
FYI: Here are the iPXE docs for https://ipxe.org/cmd/kernel and https://ipxe.org/cmd/imgfetch

build_library/vm_image_util.sh
```diff
@@ -768,6 +768,15 @@ You can pass extra kernel parameters with -append, for example:
 When using -nographic or -serial you must also enable the serial console:
   ./$(basename "${script}") -nographic -append 'console=ttyS0,115200n8'
 EOF
+    local packetipxe="$(_dst_dir)/packet.ipxe"
```
dongsupark: The ipxe script should have more clear name, like `flatcar_production_packet.ipxe`.

pothos: done

PR #66, opened 2020-05-19
[WIP] build_library: set correct SELinux contexts in final images

dongsupark:
_Work-in-progress: please do not merge_

We need to basically run `restorecon -R -v /` on all files in the final rootfs, before finalizing the image.
Without doing that, nearly every file will have `unlabeled_t` as its context.
Then with recent versions of selinux-base, some critical actions would not work correctly in SELinux enforcing mode, e.g., loading Kernel modules.

To be able to run `restorecon`, `/sys/fs/selinux` has to be mounted. Without that, it will silently skip relabelling.

This PR should be merged together with https://github.com/flatcar-linux/portage-stable/pull/66, https://github.com/flatcar-linux/coreos-overlay/pull/347.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -620,9 +620,34 @@ EOF
     sudo fstrim "${root_fs_dir}/usr" || true
   fi
 
-  # Build the selinux policy
+  # Build the selinux policy, and relabel files.
+  #
+  # We need to relabel all files in the final rootfs, before finalizing the
+  # image. Without doing that, nearly every file will have `unlabeled_t` as
+  # its context. Then with recent versions of selinux-base, some critical
+  # actions would not work correctly in SELinux enforcing mode, e.g.,
+  # loading Kernel modules. To run `restorecon`, `/sys/fs/selinux` has to
+  # be mounted. Without that, it will silently skip relabelling.
   if pkg_use_enabled coreos-base/coreos selinux; then
-      sudo chroot "${root_fs_dir}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+    local did_mount_sys=0, did_mount_sys_fs_selinux=0
+    if ! mountpoint -q "${root_fs_dir}/sys"; then
+      did_mount_sys=1
+      sudo mount -t sysfs none "${root_fs_dir}/sys"
+    fi
+    if ! mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
+      did_mount_sys_fs_selinux=1
+      sudo mount -t selinuxfs none "${root_fs_dir}/sys/fs/selinux"
+    fi
+
+    sudo chroot "${root_fs_dir}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+    sudo chroot "${root_fs_dir}" bash -c "find / -maxdepth 1 | egrep -v 'boot|proc|sys' | xargs restorecon -R || true"
+
+    if [[ ${did_mount_sys_fs_selinux} -eq 1 ]]; then
+      sudo umount "${root_fs_dir}/sys/fs/selinux"
+    fi
+    if [[ ${did_mount_sys} -eq 1 ]]; then
+      sudo umount "${root_fs_dir}/sys"
+    fi
   fi
 
   # Make the filesystem un-mountable as read-write and setup verity.


dongsupark:
Added comments in the script.

build_library/build_image_util.sh
```diff
@@ -515,9 +515,25 @@ EOF
...
+    if ! mountpoint -q "${root_fs_dir}/sys"; then
+      sudo mount -t sysfs none "${root_fs_dir}/sys"
+    fi
+    if ! mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
+      sudo mount -t selinuxfs none "${root_fs_dir}/sys/fs/selinux"
+    fi
+    sudo chroot "${root_fs_dir}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+    sudo chroot "${root_fs_dir}" bash -c "find / -maxdepth 1 | egrep -v 'boot|proc|sys' | xargs restorecon -R || true"
+    sudo umount "${root_fs_dir}/sys/fs/selinux"
+    sudo umount "${root_fs_dir}/sys"
```
krnowak, line 536: We unconditionally unmount `/sys/fs/selinux` and `/sys` here, while condtionally mounting them above. It leads me to two questions:

- Is it ever a case that `/sys` or `/sys/fs/selinux` were already mounted?
- Should unmounting be conditional too? So we unmount the filesystems only if we have just mounted them.

dongsupark: > Is it ever a case that /sys or /sys/fs/selinux were already mounted?

In an ideal world, Flatcar SDK will be created from scratch, without `/sys` inside the SDK environment being mounted. So we would not need to check for the mount, as you said.
However, there were some races around mountpoints or directories, especially it runs under Jenkins CI. (Never trust Jenkins about that!) I am afraid we cannot simply say the directories and mountpoints are always completely cleaned up. That's why I checked for the mountpoints, to be sure.

> Should unmounting be conditional too? So we unmount the filesystems only if we have just mounted them.

That sounds like a reasonable idea.
There might be race conditions, as we run `restorecon` for the whole rootfs, which could take a fair amount of time.

So how about checking for mountpoint like that?
```
    if mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
      sudo umount "${root_fs_dir}/sys/fs/selinux"
    fi
    if mountpoint -q "${root_fs_dir}/sys"; then
      sudo umount "${root_fs_dir}/sys"
    fi
```

krnowak: > > Should unmounting be conditional too? So we unmount the filesystems only if we have just mounted them.
> 
> That sounds like a reasonable idea.
> There might be race conditions, as we run `restorecon` for the whole rootfs, which could take a fair amount of time.
> 
> So how about checking for mountpoint like that?
> 
> ```
>     if mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
>       sudo umount "${root_fs_dir}/sys/fs/selinux"
>     fi
>     if mountpoint -q "${root_fs_dir}/sys"; then
>       sudo umount "${root_fs_dir}/sys"
>     fi
> ```

I was thinking more like following snippet below. But I suppose that if we have race conditions then all bets are off and stuff can break at any time.

```sh
local did_mount_sys=0, did_mount_sys_fs_selinux=0
if ! mountpoint -q "${root_fs_dir}/sys"; then
  did_mount_sys=1
  sudo mount -t sysfs none "${root_fs_dir}/sys"
fi
if ! mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
  did_mount_sys_fs_selinux=1
  sudo mount -t selinuxfs none "${root_fs_dir}/sys/fs/selinux"
fi

# do selinux business here

if [[ ${did_mount_sys_fs_selinux} -eq 1 ]]; then
  sudo umount "${root_fs_dir}/sys/fs/selinux"
fi
if [[ ${did_mount_sys} -eq 1 ]]; then
  sudo umount "${root_fs_dir}/sys"
fi
```

dongsupark: All right.
Updated the PR as you suggested.

pothos:
Needs to be rebased now to include the split SDK changes.

dongsupark:
Rebased

vbatts:
This looks ready. Just needing another quick review?

dongsupark:
No, it is still work-in-progress, as https://github.com/flatcar-linux/coreos-overlay/pull/347 is stuck with other issues.
Last when @krnowak looked into the set of SELinux updates, it somehow caused SELinux relabeling issues during Jenkins CI tests.
Not sure what the reason was.

dongsupark:
Probably not needed, please see https://github.com/kinvolk/portage-stable/pull/172 and https://github.com/kinvolk/coreos-overlay/pull/1048 .

PR #67, opened 2020-05-19
Extend OEM partition to 200Mb

samm-git:
This is needed for the https://github.com/flatcar-linux/coreos-overlay/pull/365

# Extend OEM partition to fit AWS SSM Agent

# How to use

AWS SSM agent disk space usage:

```
builder@localhost /build/amd64-usr/var/tmp/portage/app-emulation/amazon-ssm-agent-2.3.1205.0/image/usr/share/oem/bin $ ls -lh
total 139M
-rwxr-xr-x 1 root root 29M May 19 21:14 amazon-ssm-agent
-rwxr-xr-x 1 root root 24M May 19 21:14 ssm-cli
-rwxr-xr-x 1 root root 28M May 19 21:14 ssm-document-worker
-rwxr-xr-x 1 root root 11M May 19 21:14 ssm-session-logger
-rwxr-xr-x 1 root root 26M May 19 21:14 ssm-session-worker
-rwxr-xr-x 1 root root 24M May 19 21:14 ssm-updater
```
I also found that binaries are stripped already. 

# Testing done

- [x] Build with recent flatcar sdk

### DIFF
--- a/build_library/disk_layout.json
+++ b/build_library/disk_layout.json
@@ -50,7 +50,7 @@
         "label":"OEM",
         "fs_label":"OEM",
         "type":"data",
-        "blocks":"262144",
+        "blocks":"409600",
         "fs_type":"ext4",
         "mount":"/usr/share/oem"
       },


samm-git:
As a workaround - we can remove `ssm-updater`. It is not working on coreos yet anyway and is not part of the debian package as well. In this case it should be 118Mb, so should fit OEM.

cc: @pothos

dongsupark:
Yeah, I would prefer to first try to reduce size of OEM binaries.
Extending the partition in the disk layout means bigger partitions for all other images, not only AWS.

samm-git:
@dongsupark ok, will do. BTW, there are 2 options for it: 

1. To have a shared golang runtime (hacky)
1. Remove updater (which is not working anyway)

So will start with 2.

samm-git:
Done. Closing this for now

pothos:
I'm looking into using btrfs with compression for the /usr and the OEM partition. The /usr partition need to get some fixes rolled out first in a previous release but switching the OEM partition can be done already with the patches I prepared, will post PRs soon.

PR #68, opened 2020-05-22
Support the /boot/coreos/first_boot flag file

pothos:
If a user or old software creates the flag file on the old CoreOS location,
nothing would happen.
Check the old location, too, so that Ignition is rerun.

# How to use

Requires https://github.com/flatcar-linux/bootengine/pull/13 and https://github.com/flatcar-linux/update_engine/pull/5

Build available here:
```
$ wget 'https://storage.googleapis.com/flatcar-jenkins/developer/boards/amd64-usr/2345.3.1+dev-flatcar-master-449/flatcar_production_qemu.sh'
$ chmod +x flatcar_production_qemu.sh
$ wget 'https://storage.googleapis.com/flatcar-jenkins/developer/boards/amd64-usr/2345.3.1+dev-flatcar-master-449/flatcar_production_qemu_image.img.bz2'
```

# Testing done

1. Boot up without an Ignition config: `./flatcar_production_qemu.sh`
2. Create the flag file and shutdown via SSH `mkdir /boot/coreos; touch /boot/coreos/first_boot`
3. Boot up with an Ignition config `./flatcar_production_qemu.sh -i someconfig.ign`
4. Check that the config was applied

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -69,6 +69,9 @@ fi
 if [ -f "($root)/flatcar/first_boot" ]; then
     set first_boot="flatcar.first_boot=detected"
 fi
+if [ -f "($root)/coreos/first_boot" ]; then
+    set first_boot="flatcar.first_boot=detected"
+fi
 
 # Determine if the disk GUID needs to be randomized.
 search --no-floppy --set randomize_disk_guid \


build_library/grub.cfg
```diff
@@ -69,6 +69,9 @@ fi
 if [ -f "($root)/flatcar/first_boot" ]; then
     set first_boot="flatcar.first_boot=detected"
 fi
+if [ -f "($root)/coreos/first_boot" ]; then
+    set first_boot="coreos.first_boot=detected"
```
marga-kinvolk, line 73: Why set coreos.first_boot instead of flatcar.first_boot?

pothos: On one hand because it allows us to test this parameter and on the other because we could in the future give a deprecation warning for the old flag file location.

marga-kinvolk: The flag file location is separate from the grub command line parameter. I don't think it's good to add now a command line parameter that we don't intend to support.

pothos: We indent to support for now it because we aim to be a drop-in replacement.

pothos: But sure, I can remove it here as it's an internal thing.

PR #70, opened 2020-06-03
build_library: Enhance license info and store it on the image

pothos:
The license JSON file did only include the package names but not
any other metadata. Also since the file was not on the image itself,
it had to be downloaded.
Add more metadata to the license JSON and store it on the image.

# How to use

Assuming a VM was build from it:

```
$ less /usr/share/licenses/INFO
$ less /usr/share/licenses/licenses.json.bz2
$ ls /usr/share/licenses/common/
Apache-2.0.gz BSD-2.gz BSD-4.gz BSD-with-attribution.gz BSD.gz BZIP2.gz CC-BY-3.0.gz CC-BY-SA-3.0.gz CPL-1.0.gz FDL-1.3+.gz GPL-1.gz GPL-2+.gz GPL-2.gz GPL-3+.gz GPL-3.gz HPND.gz ISC.gz Info-ZIP.gz LGPL-2+.gz LGPL-2.1+.gz LGPL-2.1.gz LGPL-2.gz LGPL-3+.gz LGPL-3.gz MIT.gz MPL-2.0.gz NPL-1.1.gz OPENLDAP.gz PCRE.gz RSA.gz ZLIB.gz freedist.gz inner-net.gz lsof.gz metapackage.gz openafs-krb5-a.gz openssl.gz public-domain.gz rc.gz vim.gz
```

Opening the JSON release artifact `flatcar_production_image_licenses.json` will also show this enhanced metadata. Example: 
[flatcar_production_image_licenses.json.txt](https://github.com/flatcar-linux/scripts/files/4736319/flatcar_production_image_licenses.json.txt)

# Testing done

Run `./build_image` in a SDK with this branch checked out.
Then outside of the SDK in a new terminal:
```
$ gnome-disk-image-mounter sdk-folder/src/build/images/amd64-usr/developer-latest/flatcar_production_image.bin
$ less /run/media/"$USER"/*-*/share/licenses/INFO
$ less /run/media/"$USER"/*-*/share/licenses/licenses.json.bz2
$ ls /run/media/"$USER"/*-*/share/licenses/common/
$ gnome-disks # detach the image again
```

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -273,11 +273,80 @@ write_packages() {
     image_packages "$1" | sort > "$2"
 }
 
+# Get metadata $key for package $pkg installed under $prefix
+# The metadata is either read from the portage db folder or
+# via a portageq-BOARD invocation. In cases where SRC_URI is
+# not used for the package, fallback mechanisms are used to find
+# the source URL. Mirror names are replaced with the mirror URLs.
+get_metadata() {
+    local prefix="$1"
+    local pkg="$2"
+    local key="$3"
+    local path="${prefix}/var/db/pkg/${pkg%%:*}/${key}"
+    local val
+    if [[ -f "$path" ]]; then
+        val="$(< $path)"
+    else
+        # The package is not installed in $prefix or $key not exposed as file,
+        # so get the value from its ebuild
+        val=$(portageq-${BOARD} metadata "${BOARD_ROOT}" ebuild \
+                    "${pkg%%:*}" "${key}" 2>/dev/null ||:)
+    fi
+    # The value that portageq reports is not a valid URL because it uses a special mirror format.
+    # Also the value can be empty and fallback methods have to be used.
+    if [ "${key}" = "SRC_URI" ]; then
+        local package_name="$(echo "${pkg%%:*}" | cut -d / -f 2)"
+        local ebuild_path="${prefix}/var/db/pkg/${pkg%%:*}/${package_name}.ebuild"
+        # SRC_URI is empty for the special github.com/flatcar-linux projects
+        if [ -z "${val}" ]; then
+            # The grep invocation gives errors when the ebuild file is not present.
+            # This can happen if a "scripts" branch does not match the "coreos-overlay" branch
+            # or when the binary packages from ./build_packages are outdated.
+            val="$(grep "CROS_WORKON_PROJECT=" "${ebuild_path}" | cut -d '"' -f 2)"
+            if [ -n "${val}" ]; then
+                val="https://github.com/${val}"
+                # All github.com/flatcar-linux projects specify their commit
+                local commit=""
+                commit="$(grep "CROS_WORKON_COMMIT=" "${ebuild_path}" | cut -d '"' -f 2)"
+                if [ -n "${commit}" ]; then
+                    val="${val}/commit/${commit}"
+                fi
+            fi
+        fi
+        # During development "portageq-BOARD metadata ... ebuild ..." may result in the package not being found, fall back to a parameterized URL
+        if [ -z "${val}" ]; then
+            # Do not attempt to postprocess by resolving ${P} and friends because it does not affect production images
+            val="$(cat "${ebuild_path}" | tr '\n' ' ' | grep -P -o 'SRC_URI=".*?"' | cut -d '"' -f 2)"
+        fi
+        # Some packages use nothing from the above but EGIT_REPO_URI (currently only app-crypt/go-tspi)
+        if [ -z "${val}" ]; then
+            val="$(grep "EGIT_REPO_URI=" "${ebuild_path}" | cut -d '"' -f 2)"
+        fi
+        # Replace all mirror://MIRRORNAME/ parts with the actual URL prefix of the mirror
+        new_val=""
+        for v in ${val}; do
+            local mirror="$(echo "${v}" | grep mirror:// | cut -d '/' -f 3)"
+            if [ -n "${mirror}" ]; then
+                # Take only first mirror, those not working should be removed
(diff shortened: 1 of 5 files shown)


build_library/build_image_util.sh
```diff
@@ -273,6 +273,58 @@ write_packages() {
     image_packages "$1" | sort > "$2"
 }
 
+get_metadata() {
```
marga-kinvolk, line 276: Can you please add a few comments regarding what this function does and how it should be invoked?

build_library/build_image_util.sh
```diff
@@ -273,6 +273,58 @@ write_packages() {
...
+    local val
+    if [[ -f "$path" ]]; then
+        val="$(< $path)"
+    else
+        # The package is not installed in $prefix or $key not exposed as file,
+        # so get the value from its ebuild
+        val=$(portageq-${BOARD} metadata "${BOARD_ROOT}" ebuild \
+                    "${pkg%%:*}" "${key}" 2>/dev/null ||:)
+    fi
+    if [ "${key}" = SRC_URI ]; then
```
marga-kinvolk, line 290: Can you please add a comment about this conditional? i.e. Why is it needed, what it's doing, etc. It's pretty long, so I think it's worth explaining it a bit more.

Also, not that it matters, but I think it would be clearer if SRC_URI was between quotes.

build_library/build_image_util.sh
```diff
@@ -273,6 +273,58 @@ write_packages() {
...
 }
 
+get_metadata() {
+    local prefix="$1"
+    local pkg="$2"
+    local key="$3"
+    local path="${prefix}/var/db/pkg/${pkg%%:*}/${key}"
+    local val
+    if [[ -f "$path" ]]; then
+        val="$(< $path)"
```
marga-kinvolk, line 283: What is val supposed to be? Could you give it a more meaningful name?

build_library/build_image_util.sh
```diff
@@ -273,6 +273,58 @@ write_packages() {
...
+    else
+        # The package is not installed in $prefix or $key not exposed as file,
+        # so get the value from its ebuild
+        val=$(portageq-${BOARD} metadata "${BOARD_ROOT}" ebuild \
```
marga-kinvolk, line 287: Below you say that the portage wrapper is slow. Is this the command you're talking about?  You're invoking this command several times per package (one per key). If it's slow, it might be worth it to split this function in two to avoid calling the wrapper over and over.

build_library/build_image_util.sh
```diff
@@ -345,12 +394,41 @@ write_licenses() {
...
+    echo "The following list of projects shows their licenses, how to obtain the source code," | sudo_append "${output}"
+    echo "and which binary files in Flatcar Container Linux were created from it." | sudo_append "${output}"
+    echo "The license texts are available under /usr/share/licenses/common/ (read with \"less NAME.gz\")." | sudo_append "${output}"
+    echo "Build system files and patches used to build these projects are located at:" | sudo_append "${output}"
+    echo "https://github.com/flatcar-linux/coreos-overlay/" | sudo_append "${output}"
+    echo "https://github.com/flatcar-linux/portage-stable/" | sudo_append "${output}"
+    echo "https://github.com/flatcar-linux/scripts/" | sudo_append "${output}"
+    echo "Information on how to build Flatcar Container Linux can be found under:" | sudo_append "${output}"
+    echo "https://docs.flatcar-linux.org/os/sdk-modifying-flatcar/" | sudo_append "${output}"
+    echo | sudo_append "${output}"
```
marga-kinvolk, line 421: Why are you calling sudo_append after each echo? It makes this very hard to read... Any reason why you're not using a heredoc?

build_library/build_image_util.sh
pothos: There is `key` and `val`. I can write `value` but I thought it was a common abbreviation :)

build_library/build_image_util.sh
pothos: I moved this code. It already tries to read the value from a file if it exists but that doesn't work for all packages and it's not working for SRC_URI values.

build_library/build_image_util.sh
pothos: Yeah, heredoc would also work.

build_library/build_image_util.sh
pothos: Done

blixtra:
I'm really glad to see this information all in one place. I see this as being very useful. But a few questions...

How are other distros doing this? Is the format that we're using something that is done in other distros? I'd prefer that we did not have a "new" way of doing this. Although I understand we're in a different situation from most distros that install such files from individual packages and we may need to do it our own way.

What are the contents of `/usr/share/licenses/common/`? Could you `ls -l` that directory and post the results?

What's the reason for adding the include files section to the license file? I think that may be more data than is necessary.

Now that this file has grown in scope, I'm questioning the reason for having this as a text file and maybe it's enough to simply include the bzipped json file.

marga-kinvolk:
In Debian/Ubuntu, there's a /usr/share/doc/package/copyright file, that includes the licenses that apply to the code and the files covered by each license.  The copyright file is prepared by the maintainers of the package, which is not the case in Gentoo.

There's also a command (dpkg -L package) that lists which files belong to which package. In Flatcar, it's not easily possible to know which file belongs to which package, because all that information has been removed from the image.

Having this as a text file allows us to include the explanation at the beginning as well as the pointers to the github repos that contain our patches.

Should we schedule a meeting to discuss any further details?

build_library/build_image_util.sh
marga-kinvolk: Ok, I didn't realize that this was the value associated with the key... I guess I just find this "expose key-value pairs as files" rather confusing... Anyway, I guess it's fine.

pothos:
I added the commit IDs for URLs to https://github.com/flatcar-linux/ repositories.

build_library/build_image_util.sh
pothos: Done

PR #72, opened 2020-06-10
build_library/catalyst: Update virtual/rust during catalyst stage1 for alpha

dongsupark:
Install `virtual/rust` to avoid version conflicts that happen in case of rust versions in the SDK being different from those in the new ebuilds.

`/usr/share/catalyst/targets/stage1/stage1-chroot.sh` installs gcc and its dependencies, including `dev-lang/rust`, while `virtual/rust` does not get updated. That results in version conflicts between `virtual/rust` and `dev-lang/rust`.

To avoid such an issue, we should update also `virtual/rust` when building stage1. Since `virtual/rust` automatically pulls in `dev-lang/rust`, we do not need to explicitly specify `dev-lang/rust` here.

Also deal with both virtual/rust and dev-lang/rust.
Toolchain utils have installed only `dev-lang/rust`. It could result in version mismatch between `virtual/rust` and `dev-lang/rust`, because `dev-lang/rust` does not automatically pull in `virtual/rust`.
So install `virtual/rust` instead of `dev-lang/rust`.

## How to use

```
sudo FLATCAR_DEV_BUILDS_SDK=https://storage.googleapis.com/flatcar-jenkins/sdk \
  /mnt/host/source/src/scripts/bootstrap_sdk

sudo /mnt/host/source/src/scripts/build_toolchains
```

### DIFF
--- a/build_library/catalyst.sh
+++ b/build_library/catalyst.sh
@@ -121,6 +121,15 @@ target: stage1
 # stage1 packages aren't published, save in tmp
 pkgcache_path: ${TEMPDIR}/stage1-${ARCH}-packages
 update_seed: yes
+# Install virtual/rust to avoid version conflicts that happen in case of
+# rust versions in the SDK being different from those in the new ebuilds.
+# /usr/share/catalyst/targets/stage1/stage1-chroot.sh installs gcc and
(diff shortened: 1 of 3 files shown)


dongsupark:
Cherry-picked to edge.

PR #79, opened 2020-07-13
common: Support reading the package URL env vars from version.txt

pothos:
The environment variables FLATCAR_DEV_BUILDS and FLATCAR_DEV_BUILDS_SDK
define where the base URL for the binary package store of the board
packages and the SDK packages. To set it, they were either exported in
the chroot or passed each time as parameter but this was only available
for the SDK packages in the tricky order
"update_chroot --dev_builds_sdk URL" (or --binhost) and then
"./build_packages --skip_chroot_upgrade".
The defining information for binary package URL comes from version.txt
but it lacks the base URL for a convenient use. Add the base URLs for
SDK and board packages there to be able to get binary packages without
manual tricks and errors. Any changes to
~/trunk/.repo/manifests/version.txt will directly be picked up
from ./build_packages and friends to set the correct URLs in
/etc/portage/make.conf and /build/BOARD/etc/portage/make.conf
so that even a manual "emerge(-BOARD) --usepkg --getbinpkg" uses them.
A helper script "./set_version" is provided to ease modifying
~/trunk/.repo/manifests/version.txt by resolving the latest nightly
version.
Later this functionality can also be used to simplify the Jenkins code
which currently sets these variables (but the special case of the
Release Base download being different from the upload holds).

# How to use

```
cork enter
./set_version --dev-board --board-version amd64-usr/stable-nightly
./build_packages
```

# Testing done

I tested that building an image with the local SDK uses the correct binary package URLs (emerge downloads them) and that the script modifies the files correctly.

### DIFF
--- a/set_version
+++ b/set_version
@@ -1,0 +1,91 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DEFAULT_BASE_URL="https://storage.googleapis.com/flatcar-jenkins"
+DEV_BOARD_URL="${DEFAULT_BASE_URL}/developer"
+DEV_SDK_URL="${DEFAULT_BASE_URL}/developer/sdk"
+FILE=~/trunk/.repo/manifests/version.txt
+FLATCAR_DEV_BUILDS_SDK=""
+FLATCAR_SDK_VERSION=""
+FLATCAR_DEV_BUILDS=""
+FLATCAR_VERSION=""
+FLATCAR_VERSION_ID=""
+
+# To keep this script usable outside of the SDK chroot, do not include common.sh here.
+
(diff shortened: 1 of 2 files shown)


set_version
```diff
@@ -0,0 +1,91 @@
...
+      shift
+      ;;
+    *)
+      echo "Unkown argument ${ARG}"
+      exit 1
+      ;;
+  esac
+done
+
+if [[ "${FILE}" != "/dev/stdout" ]] && [[ "${FILE}" != "/dev/stdout" ]]; then
```
krnowak, line 77: The `[[ "${FILE}" != "/dev/stdout" ]]` condition seems to be duplicated here.

pothos: Thanks, this was supposed to be `stderr`

PR #87, opened 2020-08-13
build_library: cgroupv2 only

vbatts:
# build_library: cgroupv2 only

currently [edge] has cgroup v2 available via hybrid hierarchy, which is
_fine_, but that is not the goal. Testing a real next step for linux
will be to run unified hierarchy _only_.

This will certainly not work for all container runtimes currently, but
having access to the default behavior is needed for better testing and
exposure.

Signed-off-by: Vincent Batts <vbatts@kinvolk.io>

# How to use

you can currently already test the behaviour of this today by adding this kernel parameter. 
In `/usr/share/oem/grub.cfg`:
```
set linux_append="systemd.unified_cgroup_hierarchy=true"
```
and reboot your flatcar host. Then poke around for what breaks and which upstream project to file issues with or get their newer release.

# Testing done

tested that the image builds and runs. 

Granted, things like `docker` don't currently work, but that is part of the upstream testing and feedback needed.

### DIFF
--- a/build_library/grub.cfg
+++ b/build_library/grub.cfg
@@ -19,7 +19,7 @@ set randomize_disk_guid=""
 set oem_id=""
 
 # Anything else the OEM adds should use this variable.
-set linux_append="systemd.unified_cgroup_hierarchy=false systemd.legacy_systemd_cgroup_controller=false"
+set linux_append="systemd.unified_cgroup_hierarchy=true"
 
 set secure_boot="0"
 


dongsupark:
Agreed in general that the pure cgroupv2 is the right direction.
Though I suppose first of all we would need to sort out Edge branches to make an Edge release, before making code changes.

build_library/grub.cfg
```diff
@@ -19,7 +19,7 @@ set randomize_disk_guid=""
 set oem_id=""
 
 # Anything else the OEM adds should use this variable.
-set linux_append="systemd.unified_cgroup_hierarchy=false systemd.legacy_systemd_cgroup_controller=false"
+set linux_append="systemd.unified_cgroup_hierarchy=true"
```
pothos, line 22: Instead of changing this here, can we remove this parameter and only change the flag in the systemd ebuild file?

vbatts: fair

vbatts:
> Agreed in general that the pure cgroupv2 is the right direction.
> Though I suppose first of all we would need to sort out Edge branches to make an Edge release, before making code changes.

an updated release is good too. I'm not forcing that point here though.
Just giving a place to test.

vbatts:
closing for now, as edge already has hybrid cgroupv2

PR #88, opened 2020-08-25
lsb_release: remove codename

vbatts:
reading through the LSB specs
(https://refspecs.linuxfoundation.org/lsb.shtml), the `DISTRIB_CODENAME`
is not a required field. And if not present, then "n/a" is the value
returned from the `lsb_release` command (which is not even provided in
flatcar ...).

Further this Codename has no bearing on releases or series or versions.
Historically
(https://github.com/flatcar-linux/scripts/commits/main/build_library/set_lsb_release)
the codename was changed sporadically.
This leads users to associate a range of releases as a particular
series, when this may not be the case.

At best, the up-side of some arbitrary codename if it is memorable as
landmark in memory.

Signed-off-by: Vincent Batts <vbatts@kinvolk.io>


build_library/set_lsb_release
```diff
@@ -26,10 +26,9 @@ ROOT_FS_DIR="$FLAGS_root"
 [ -d "$ROOT_FS_DIR" ] || die "Root FS does not exist? ($ROOT_FS_DIR)"
 
 OS_NAME="Flatcar Container Linux by Kinvolk"
-OS_CODENAME="Oklo"
 OS_ID="flatcar"
 OS_ID_LIKE="coreos"
-OS_PRETTY_NAME="$OS_NAME $FLATCAR_VERSION (${OS_CODENAME})"
+OS_PRETTY_NAME="$OS_NAME $FLATCAR_VERSION"
```
pothos, line 31: Maybe make it possible to have one but not mandatory
```
if [ "${OS_CODENAME}" = "" ]; then
  CODENAME_OPTION=""
else
  CODENAME_OPTION=" (${OS_CODENAME})"
fi
OS_PRETTY_NAME="$OS_NAME $FLATCAR_VERSION${CODENAME_OPTION}"
```

vbatts: after looking at this further, I'm thinking it'll be confusing to some for CODENAME to just disappear. And then again confusing for it to possibly re-appear?

vbatts:
putting this as a draft, until we agree on the name.

PR #90, opened 2020-09-01
contrib: allow newer ssh keys, and user-provided template

vbatts:
When making an ISO locally, there are unit files I would like to specify without editing this script everytime. :smile_cat: 
And ran this script through `shellcheck` for linting.

### DIFF
--- a/contrib/create-basic-configdrive
+++ b/contrib/create-basic-configdrive
@@ -43,7 +43,7 @@ ssh_authorized_keys:
   - <SSH_KEY>
 hostname: <HOSTNAME>
 "
-REGEX_SSH_FILE="^ssh-(rsa|dss|ed25519) [-A-Za-z0-9+\/]+[=]{0,2} .+"
+REGEX_SSH_FILE="^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp521) [-A-Za-z0-9+\/]+[=]{0,2} .+"
 
 while getopts "d:e:c:H:i:n:p:S:t:l:u:h" OPTION
 do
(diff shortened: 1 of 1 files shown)


contrib/create-basic-configdrive
```diff
@@ -42,13 +42,14 @@ ssh_authorized_keys:
...
-REGEX_SSH_FILE="^ssh-(rsa|dss|ed25519) [-A-Za-z0-9+\/]+[=]{0,2} .+"
+REGEX_SSH_FILE="^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp521) [-A-Za-z0-9+\/]+[=]{0,2} .+"
 
-while getopts "d:e:H:i:n:p:S:t:l:u:h" OPTION
+while getopts "d:e:c:H:i:n:p:S:t:l:u:h" OPTION
 do
     case $OPTION in
         d) ETCD_DISCOVERY="$OPTARG" ;;
         e) ETCD_ADDR="$OPTARG" ;;
+        c) CLOUD_CONFIG_FILE="${OPTARG}" ;;
```
dongsupark, line 52: We should probably also update `USAGE` in the same file, to add the `-c` option.

dongsupark (approved):
Looks good in general.

BTW this script looks like a really old legacy that almost no one has touched.
For example, it still starts `fleet`. :facepalm: 
Anyway we can clean it up later.

contrib/create-basic-configdrive
vbatts: ah yeah, i forgot the usage info. 
and yes, i was not yet going to axe all that stuff.

PR #93, opened 2020-09-17
Add Azure OEM and selective private upload

sayanchowdhury:
# Add Azure OEM and selective private upload

This PR adds the Azure Premium OEM and selective upload the premium image to GCS as a private object

This commit would upload the image to GCS as a private object and then mantle would take over the process.

### DIFF
--- a/jenkins/vm.sh
+++ b/jenkins/vm.sh
@@ -43,6 +43,13 @@ img=src/flatcar_production_image.bin
 [[ "${img}.bz2" -nt "${img}" ]] &&
 enter lbunzip2 -k -f "/mnt/host/source/${img}.bz2"
 
+PRIVATE_UPLOAD_OPT=""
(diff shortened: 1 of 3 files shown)


jenkins/vm.sh
```diff
@@ -43,6 +43,12 @@ img=src/flatcar_production_image.bin
 [[ "${img}.bz2" -nt "${img}" ]] &&
 enter lbunzip2 -k -f "/mnt/host/source/${img}.bz2"
 
+PRIVATE_UPLOAD_OPT=""
+if [[ ${FORMAT} == 'azure_premium' ]]
```
pothos, line 47: Maybe quoting is better: "${FORMAT}"

sayanchowdhury: Done

marga-kinvolk (approved):
LGTM

PR #94, opened 2020-09-21
build_library: Rename the images to use pro instead of premium

sayanchowdhury:
# build_library: Rename the images to use pro instead of premium

Signed-off-by: Sayan Chowdhury <sayan@kinvolk.io>


marga-kinvolk (approved):
LGTM

PR #95, opened 2020-09-24
build_image: Allow a two-part version value leaving out FLATCAR_PATCH

pothos:
When a version scheme with x.y is detected, don't include the
FLATCAR_PATCH field in the version.txt file.

# How to use

Build an LTS image with a two-part version scheme.

# Testing done

Image built, checked the contents of /etc/os-release /usr/share/flatcar/release and the version.txt file.


build_image
```diff
@@ -196,7 +196,11 @@ split_ver "${FLATCAR_VERSION_ID}" SPLIT
 tee "${BUILD_DIR}/version.txt" <<EOF
 FLATCAR_BUILD=${SPLIT[0]}
 FLATCAR_BRANCH=${SPLIT[1]}
-FLATCAR_PATCH=${SPLIT[2]}
+EOF
+if [[ ${#SPLIT[@]} -eq 3 ]]; then
```
vbatts, line 200: probably could end up in common.sh

vbatts:
Seems like a good starting point. I'm sure we'll find various places that expect xx.yy.zz, so this will be iterative

build_image
pothos: The whole file creation part here? The version.txt is only created once for an image.

vbatts: sorry, was thinking about the version comparisons.

marga-kinvolk (approved):
LGTM assuming tests pass :)

PR #98, opened 2020-09-30
jenkins: Generate and upload update payloads too

krnowak:
This may come handy for testing the updates. This change should be
enough, since the `generate_update` function also uploads the
generated payloads if the `--upload` flag was passed. Which this
script is doing.

# How to use

Nothing else is needed to do, just run the build. The uploading of the update payload should happen as a part of the image matrix part.

# Testing done

Not finished yet: http://localhost:9091/job/os/job/manifest/1137/

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -94,7 +94,7 @@ generate_update() {
       -new_kernel "${image_kernel}" \
       -out_file "${update}.gz"
 
-  upload_image -d "${update}.DIGESTS" "${update}".{bin,gz,zip}
+  upload_image -d "${update}.DIGESTS" "${update}".{bin,gz}
 }
 
 # ldconfig cannot generate caches for non-native arches.
--- a/jenkins/images.sh
+++ b/jenkins/images.sh
@@ -67,6 +67,7 @@ done < <(jq -r '.value.packages[] | . as $p | .name as $n | $p.versions[] | [.ca
 script build_image \
     --board="${BOARD}" \
     --group="${GROUP}" \
+    --generate_update \
     --getbinpkg \
     --getbinpkgver="${FLATCAR_VERSION}" \
     --sign="${SIGNING_USER}" \


krnowak:
Updated. Apparently, we generate `flatcar_production_update.bin`, `flatcar_production_update.gz`, `flatcar_production_update.bz2` and `flatcar_production_update.zip`. Confusingly, the last one is not an update payload, but rather a bunch of tools for applying the update?

Anyway, will need to run another test build to check it.

pothos:
Correct, the final file is the signed `flatcar_production_update.gz` payload. To sign it you can invoke `signing/sign.sh` but the expected key is mismatching of course since we sign the update payloads with the real key in a manual process and it would only use the `signing/devel.key.pem` key. That's good for testing if we either change the developer image builds to embed the matching dummy key or if we document how to download and bind-mount the dummy key into the developer image builds with something like `sudo mount --bind /tmp/key /usr/share/update_engine/update-payload-key.pub.pem`.

krnowak:
Test build running: http://localhost:9091/job/os/job/manifest/1145/

krnowak:
So I'm not sure if I should be pursuing this PR. From what I have discovered, we actually generate some update payloads. I haven't noticed it before because google cloud platform browser has just a terrible UI - it shows only 50 files in a directory, and you need to press some arrow at the bottom of the screen to see the rest (you need to scroll the file listing to the very bottom, not very discoverable). So from I have seen, we actually have `flatcar_production_update.bin.bz2` and `flatcar_production_update.zip` (together with signatures, digests and whatnot).

I don't yet understand what is the difference between `generate_update` and `extract_update` in https://github.com/flatcar-linux/scripts/blob/86800378db508e1ab11b7bdd3825970f97052e7e/build_image#L178-L182, but both generate the `flatcar_production_update.bin.bz2` file (`generate_update` also creates `flatcar_production_update.gz`). So, without this PR, we hit only the `extract_update` path (because `FLAGS_extract_update` is `${FLAGS_TRUE}` by default), and with this PR, we hit only the `generate_update` path (because we pass `--generate_update` flag explicitly).

pothos:
The final file needed is `flatcar_production_update.gz` signed by the key that is embedded under `/usr/share/update_engine/update-payload-key.pub.pem`.

pothos:
I chose a different approach in https://github.com/kinvolk/flatcar-scripts/pull/130 to keep the current behavior but generating an update payload with a custom name since it's only for testing.

krnowak:
I think we can just close this, since we actually generated the payloads. And I don't remember any more why I needed them anyway. :) So not sure I'll be doing any follow-up of it.

pothos:
Yes, thanks for pioneering!

PR #100, opened 2020-10-26
jenkins: do not configure ccache variables

dongsupark:
Setting the invalid `CCACHE_` variables resulted in strange failure in projects depending on meson, newer version like 0.55.3.
For example systemd build fails like the following errors:

```
 * ACCESS DENIED:  utimes:       /mnt/host/source/ccache
 * ACCESS DENIED:  utimes:       /mnt/host/source/ccache

F: utimes
S: deny
P: /mnt/host/source/ccache
A: /mnt/host/source/ccache
R: /mnt/host/source/ccache
C: ccache cc /build/amd64-usr/var/tmp/portage/sys-apps/systemd-246/work/systemd-246-abi_x86_64.amd64/meson-private/sanitycheckc.c -o /build/amd64-usr/var/tmp/portage/sys-apps/systemd-246/work/systemd-246-abi_x86_64.amd64/meson-private/sanitycheckc.exe -O1 -pipe -pipe -D_FILE_OFFSET_BITS=64
```

We should not set up ccache at all, as it has been already disabled in coreos-overlay repo.

## Testing done

CI passed

### DIFF
--- a/build_library/catalyst.sh
+++ b/build_library/catalyst.sh
@@ -236,12 +236,9 @@ catalyst_init() {
 }
 
 write_configs() {
-    # No catalyst config option, so defined via environment
-    export CCACHE_DIR="$TEMPDIR/ccache"
-
     info "Creating output directories..."
-    mkdir -m 775 -p "$TEMPDIR/portage/repos.conf" "$DISTDIR" "$CCACHE_DIR"
-    chown portage:portage "$DISTDIR" "$CCACHE_DIR"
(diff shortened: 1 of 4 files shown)


krnowak:
I'm going to pick that up for the systemd update to see if it does the trick. I had the similar failure there.

dongsupark:
Test builds passed.
Merging for unblocking Alpha builds.

dongsupark:
cherry-picked to flatcar-2671

PR #102, opened 2020-11-18
build_packages, build_image_util.sh: fix up liblto symlink

t-lo:
This change fixes an issue with builds that use a gcc-9.3.0 SDK. `gcc-config` in that SDK version creates a symlink in binutils' libs directory to gcc's `liblto` since some packages expect binutils to ship that library. The wrong symlink is created for chroot environments however, which makes a health check fail in the `build_packages` and `build_image` script. We fix this by correcting the symlink with a new custom function in `common.sh`.

This change is required in order to build the packages and image with the gcc-9.3.0 SDK.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -966,3 +966,23 @@ clean_qemu_static() {
     *) die "Unsupported arch" ;;
   esac
 }
+
+# Fix up liblto softlink created by gcc-config
+fixup_liblto_softlinks() {
+  local root="$1"
+
+  info "fixup_liblto_softlinks: Looking for broken softlinks in '${root}'"
+
+  local link
+  local target
+  # check both native (/usr/CHOST/) as well as cross compile (/usr/CHOST/CTARGET) paths
+  { ls -l "${root}/"usr/*/binutils-bin/lib/bfd-plugins/liblto_plugin.so 2>/dev/null;
+    ls -l "${root}/"usr/*/*/binutils-bin/lib/bfd-plugins/liblto_plugin.so 2>/dev/null; } \
+    | sed 's:.* \([^[:space:]]\+\) -> \([^[:space:]]\+\):\1 \2:' \
+    | while read link target; do
+              local newtarget=$(echo "$target" | sed "s:${root}:/:")
+              info "   Fixing up broken $link -> $target"
+              info "   sudo ln -sf $newtarget $link"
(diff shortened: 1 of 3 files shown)


common.sh
```diff
@@ -966,3 +966,23 @@ clean_qemu_static() {
...
+
+# Fix up liblto softlink created by gcc-config
+fixup_liblto_softlinks() {
+  local root="$1"
+
+  info "fixup_liblto_softlinks: Looking for broken softlinks in '${root}'"
+
+  local link
+  local target
+  # check both native (/usr/CHOST/) as well as cross compile (/usr/CHOST/CTARGET) paths
```
pothos, line 978: This links from `/build/arm64-usr/` to `//usr/` – is the target link supposed to work in the `/build/arm64-usr` chroot, i.e., to a path `/build/arm64-usr/usr/libexec/gcc/aarch64-cros-linux-gnu/9.3.0/liblto_plugin.so` from the perspective of the top chroot or is it supposed to point to a path in the top chroot's `/usr`?

pothos: In the old SDK this was simply a relative link from `liblto_plugin.so` to `liblto_plugin.so.0.0.0` which I think should be done here, too, or?

t-lo: The link is supposed to work from within the `/build/arm64-usr/` chroot. The soft link is sanity-checked by [build_library/test_image_contents.sh::test_image_content()](https://github.com/kinvolk/flatcar-scripts/blob/main/build_library/test_image_content.sh#L34) in the [symlink check](sudo ROOT="$root" "$check_root" symlink), which calls [check_symlink in build_library/check_root](https://github.com/kinvolk/flatcar-scripts/blob/main/build_library/check_root#L287).

That sanity check will validate softlinks within the target chroot. It is used by build_packages and build_image, and it would have made the build fail if the "fixed" softlinks would have been dangling.

pothos: Ok, good, I didn't notice that this is not the same link we already had in the SDK because it is in a second one from another folder. Then it makes sense to use absolute paths.

In https://storage.googleapis.com/flatcar-jenkins/developer/developer/boards/amd64-usr/2020.11.17+dev-flatcar-master-1482/flatcar_developer_container_contents.txt it looks good, just the `//` prefix is a bit confusing but still valid.

pothos (approved):

PR #105, opened 2020-12-03
image_to_vm: export vendor partition contents with vm image build

t-lo:
This change to `image_to_vm.sh` generates a vendor partition tarball if the file`/usr/share/oem/flatcar-oem-release` is present. `flatcar-oem-release` is populated with the current release version as well as with the name of the vendor partition tarball.
The vendor partition tarball is then added to the list of files to be uploaded, so it can be published on the update server.

TODO:
- change our publishing process to sign the tarball with the update key and also publish the vendor tools OEM tarball alongside the update payload


t-lo:
Closing in favour of the sysext initiative.

PR #106, opened 2020-12-09
Add AWS Pro OEM type

pothos:
This is the initial creation of an AWS Pro image coming with support,
and features added on top of it.

*Note:* Pick for 2605 and 2705

# How to use

# Testing done

Built image and ran kola tests


sayanchowdhury (approved):
Looks good to me

PR #107, opened 2020-12-14
dev_container_util.sh: Use the shflags group instead of the default flag

sayanchowdhury:
# dev_container_util.sh: Use the shflags group instead of the default flag

The build_image script invokes the create_dev_container function and
passes the `FLAGS_group` as param. Use the param, to generate the
binhost URL instead of using the DEFAULT_GROUP which stays as developer
always.

Fixes: kinvolk/Flatcar#298
Signed-off-by: Sayan Chowdhury <sayan@kinvolk.io>

Note: Pick the patch to `flatcar-2605` and `flatcar-2705`

# How to use

```
./build_image --group=stable container
```

# Testing done

Built a test container locally with the `--group` parameter and verified the `/etc/portage/make.conf`

### DIFF
--- a/build_library/dev_container_util.sh
+++ b/build_library/dev_container_util.sh
@@ -3,10 +3,12 @@
 # found in the LICENSE file.
 
 get_binhost_url() {
-	if [ "${DEFAULT_GROUP}" == "developer" ]; then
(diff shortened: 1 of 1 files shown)


pothos (approved):

sayanchowdhury:
Picked to `flatcar-2605` and flatcar`2705` branch.

PR #108, opened 2020-12-29
Add GCE Pro OEM type

marga-kinvolk:
# Add GCE Pro OEM type

This adds gce_pro as one of the valid formats, currently it installs the `oem-gce` package.

# How to use / Testing done
Built image and run kola tests (vm-matrix 1896, kola/gce 366)

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -18,6 +18,7 @@ VALID_IMG_TYPES=(
     digitalocean
     exoscale
     gce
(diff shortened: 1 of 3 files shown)


pothos (approved):

sayanchowdhury (approved):

marga-kinvolk:
I've cherry-picked this to flatcar-2605 and flatcar-2705, so that we can automatically generate stable and beta images with the next release.

PR #109, opened 2021-01-19
build_library: Fix git URLs

krnowak:
The repos were moved to kinvolk. Redirects still work, but it's better
not to rely on them.


marga-kinvolk (approved):
LGTM

PR #110, opened 2021-01-20
Fall back to source repository license files if not in portage

pothos:
When a license file is newly added, portage may not yet have it in the
shared folder and the license inclusion step fails.
Fall back to the source repositories and look for the license file
there, too. Print a warning if not found instead of failing to build.

[Test](http://localhost:9091/job/os/job/manifest/1848/)

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -452,8 +452,23 @@ EOF
     sudo lbzip2 -9 "${root_fs_dir}"/usr/share/licenses/licenses.json
     # Copy all needed licenses to a "common" subdirectory and compress them
     local license_list # define before assignment because it would mask any error
-    license_list="$(jq -r '.[] | "/var/gentoo/repos/gentoo/licenses/\(.licenses | .[])"' "${json_input}" | sort | uniq)"
-    sudo cp ${license_list} "${root_fs_dir}"/usr/share/licenses/common/
+    license_list="$(jq -r '.[] | "\(.licenses | .[])"' "${json_input}" | sort | uniq)"
+    local license_dirs=(
+        "/mnt/host/source/src/third_party/coreos-overlay/licenses/"
+        "/mnt/host/source/src/third_party/portage-stable/"
+        "/var/gentoo/repos/gentoo/licenses/"
+        "none"
+    )
+    for license_file in ${license_list}; do
+        for license_dir in ${license_dirs[*]}; do
+            if [ "${license_dir}" = "none" ]; then
+                warn "The license file \"${license_file}\" was not found"
+            elif [ -f "${license_dir}${license_file}" ]; then
+                sudo cp "${license_dir}${license_file}" "${root_fs_dir}"/usr/share/licenses/common/
+                break
+            fi
+        done
+    done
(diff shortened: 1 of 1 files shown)


build_library/build_image_util.sh
```diff
@@ -452,8 +452,18 @@ EOF
     sudo lbzip2 -9 "${root_fs_dir}"/usr/share/licenses/licenses.json
     # Copy all needed licenses to a "common" subdirectory and compress them
     local license_list # define before assignment because it would mask any error
-    license_list="$(jq -r '.[] | "/var/gentoo/repos/gentoo/licenses/\(.licenses | .[])"' "${json_input}" | sort | uniq)"
-    sudo cp ${license_list} "${root_fs_dir}"/usr/share/licenses/common/
+    license_list="$(jq -r '.[] | "\(.licenses | .[])"' "${json_input}" | sort | uniq)"
+    for license_file in ${license_list}; do
+      if [ -f "/var/gentoo/repos/gentoo/licenses/${license_file}" ]; then
```
marga-kinvolk, line 457: I think a nested loop that iterates over possible directories here would make it a bit more readable (less repeated code, in particular less repeated routes):

```
LICENSE_DIRS=(
"/var/gentoo/repos/gentoo/licenses",
"/mnt/host/source/src/third_party/portage-stable/licenses",
"/mnt/host/source/src/third_party/coreos-overlay/licenses"
)
for license_file in ${license_list}; do
  for license_dir in ${LICENSE_DIRS}; do
    if [ -f "$license_dir/$license_file" ]; then
      cp "$license_dir/$license_file" "${root_fs_dir}"/usr/share/licenses/common/
    fi
  done
done
```

pothos: Yes, would work when I add a dummy entry as end of the list to print the warning when no file was found

pothos: Updated it (Alternatives would be "break N" or some state variable)

build_library/build_image_util.sh
```diff
@@ -452,8 +452,23 @@ EOF
...
+        "/mnt/host/source/src/third_party/portage-stable/"
+        "/var/gentoo/repos/gentoo/licenses/"
+        "none"
+    )
+    for license_file in ${license_list}; do
+        for license_dir in ${license_dirs}; do
+            if [ "${license_dir}" = "none" ]; then
+                warn "The license file \"${license_file}\" was not found"
+            elif [ -f "${license_dir}${license_file}" ]; then
+                sudo cp "${license_dir}${license_list}" "${root_fs_dir}"/usr/share/licenses/common/
```
marga-kinvolk, line 467: This says list where it should say file.

pothos: Thanks

marga-kinvolk (approved):
LGTM assuming you tested this :)

pothos:
I'm getting unrelated build errors in the environment where I don't have the portage license dir up-to-date:

```
/usr/bin/smb2-quota: (env)/python does not exist
/usr/share/nghttp2/fetch-ocsp-response: (env)/python does not exist
```
I can try again on Jenkins but that doesn't test if it actually falls back.

PR #111, opened 2021-01-22
vm_image_util: Use new aws_pro flag for ec2-compat

marga-kinvolk:
# Use new aws_pro flag for ec2-compat package

When building the ec2-compat package for the ami_vmdk_pro image format, use the new aws_pro flag instead of the existing one.

It should be merged together with https://github.com/kinvolk/coreos-overlay/pull/794


krnowak (approved):

marga-kinvolk:
Cherry picked to flatcar-2605 and flatcar-2705

PR #112, opened 2021-02-03
Improve the dependency cycle breaking logic

krnowak:
The `break_dep_loop` function is updated to take pairs of package name and multiple USE flags to be disabled, and makes sure all the packages in pairs are rebuilt. That way it's more robust against package build ordering.

In our case the issue was that we had a dependency cycle like `sys-apps/systemd` -> `sys-fs/cryptsetup` -> `virtual/udev` -> `sys-apps/systemd`. We were breaking the cycle at the first arrow, thus not building `sys-fs/cryptsetup` in dependency-cycle-breaking emerge as it was not pulled by anything else as a dependency. When doing the rest-of-the-packages-building emerge, `sys-fs/cryptsetup` was built and `sys-apps/systemd` was rebuilt, in that order. But after updating portage to a newer version, the order was reversed and `sys-apps/systemd` failed during the configure phase, because it depends on a yet unbuilt `sys-fs/cryptsetup`. I think that the building order of packages that are in the dependency cycle is arbitrary, so to guard against such a scenario we build all the packages in the dependency cycle. That way, when rebuilding them in any order we are sure that their dependencies are already installed either by the dependency-cycle-breaking emerge or the rest-of-the-packages-building emerge.

This change has moved updating portage forward.

Test build: http://localhost:9091/job/os/job/manifest/1950/cldsv/ (arm64 build failure is because some nvidia drivers are missing the keyword changes).

### DIFF
--- a/build_packages
+++ b/build_packages
@@ -266,11 +266,18 @@ break_dep_loop() {
 }
 
 if [[ "${FLAGS_usepkgonly}" -eq "${FLAGS_FALSE}" ]]; then
+  # Breaking the following loops here:
+  #
   # util-linux[udev] -> virtual->udev -> systemd -> util-linux
-  break_dep_loop sys-apps/util-linux udev,systemd sys-apps/systemd cryptsetup
-
-  # systemd[cryptsetup] -> cryptsetup -> lvm2 -> virtual/udev -> systemd
-  break_dep_loop sys-apps/systemd cryptsetup
+  # util-linux[systemd] -> systemd -> util-linux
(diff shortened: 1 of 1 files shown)


krnowak:
Started a new test build: http://localhost:9091/job/os/job/manifest/1991/

krnowak:
CI passed, package-matrix job on arm64 took a bit over 2 hours instead of 11, so it's back to normal. qemu-uefi job failed because of usual soft-lockups.

t-lo (approved):
LGTM, thank you!

PR #117, opened 2021-02-05
bootstrap_sdk: Put image, contents and digests paths into variables

krnowak:
It makes it a bit easier to read and clearer what the script is doing
without all that repetition. No functional changes.


marga-kinvolk (approved):
LGTM

PR #123, opened 2021-03-04
Fix SDK bootstrap in bootstrap_sdk: no package updates in stage 1

t-lo:
# SDK bootstrap related fixes

This change updates the stage1 SDK bootstrap build to use local
("known good") package ebuilds only, preventing updated package ebuilds
to apply in stage 1. This fixes SDK build breakage we observed when
upgrading core libraries like readline.

The change also removes the seed update from stage 1 as it should not
be needed anymore now that we postpone any package updates to stage 2.

The following package ebuild repos are used for stage 1:
- for portage-stable, we simply copy /var/gentoo/repos/gentoo
   from the SDK root.
- coreos-overlay is more complicated since ebuilds are missing from
   the SDK. So we grok the version the SDK was built with from
     /mnt/host/source/.repo/manifests/default.xml
   and then we create a local stage 1 clone of
     https://github.com/kinvolk/coreos-overlay.git
   in which we then check out the revision noted in the default mnifest.

# How to use

Needed in combination with
* portage-stable: https://github.com/kinvolk/portage-stable/pull/152
* coreos-overlay: https://github.com/kinvolk/coreos-overlay/pull/874

```
$ cork enter
$ sudo ./bootstrap_sdk
```

# Testing done

SDK CI build (manifest #2158) passed.

### DIFF
--- a/bootstrap_sdk
+++ b/bootstrap_sdk
@@ -10,20 +10,32 @@
 # 
 # For reference the procedure it performs is this:
 # 
-# 1. snapshot: Grab a snapshot of portage-stable. Note that overalys are
-#    not snapshotted.
+# 1. snapshot: Grab a snapshot of the portage-stable repo from
+#    third_party/... . coreoy-overlay is used as-is from the same source.
+#
 # 2. stage1: Using a "seed" tarball as a build environment, build a
 #    minimal root file system into a clean directory using ROOT=...
 #    and USE=-* The restricted USE flags are key be small and avoid
-#    circular dependencies.
+#    circular dependencies. Tarball is built from the current SDK's
+#    portage-stable and coreos-overlay instead of the ones from 1
+#    since stage1 must not contain updated ebuilds (see build_stage1 below).
+#    This stage uses:
+#    - portage-stable from the SDK's /var/lib/gentoo/repos/gentoo
+#    - a fresh check-out of coreos-overlay at the revision used to build the SDK
+#
 # 3. stage2: Run portage-stable/scripts/bootstrap.sh
-#    This rebuilds the toolchain. Probably not strictly necessary most of
-#    the time but does super-duper-promise that the toolchain isn't linked
+#    This rebuilds the toolchain using Gentoo bootstrapping, ensuring it's not linked
 #    to or otherwise influenced by whatever was in the "seed" tarball.
-# 4. stage3: Run emerge -e system to rebuild everything using the fresh
-#    toolchain using the normal USE flags provided by the profile. This
+#    The toolchain rebuild may contain updates package ebuilds from
+#    third_party/(portage-stable|coreos-overlay).
+#    This and all following stages use portage-stable and coreos-overlay
+#    from third_party/... (see 1.)
+#
+# 4. stage3: Run emerge -e system to rebuild everything using the fresh updated
+#    toolchain from 3., using the normal USE flags provided by the profile. This
 #    will also pull in assorted base system packages that weren't included
 #    in the minimal environment stage1 created.
+#
 # 5. stage4: Install any extra packages or other desired tweaks. For the
 #    sdk we just install all the packages normally make_chroot.sh does.
 #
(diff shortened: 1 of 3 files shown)


bootstrap_sdk
```diff
@@ -83,6 +83,54 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
 chmod 1777 "${ROOT_OVERLAY}/tmp"
 cp "${BUILD_LIBRARY_DIR}/toolchain_util.sh" "${ROOT_OVERLAY}/tmp"
 
+
+# Stage 1 uses a different ebuild repo snapshot as well as a different portage_overlay 
+# the ones from the SDK where ./bootstrap_sdk is being executed -
```
krnowak, line 88:
```suggestion
# than the ones from the SDK where ./bootstrap_sdk is being executed -
```

bootstrap_sdk
```diff
@@ -83,6 +83,54 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
...
+# to buils a known-good stage 1 (see stages description at top
```
krnowak, line 89:
```suggestion
# to build a known-good stage 1 (see stages description at top
```

bootstrap_sdk
```diff
@@ -83,6 +83,54 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
...
+# of file). This prevents package upgrades from updated ebuild files in the portage/coreos
+# to apply in stage 1.
+# Stage 1 lacks proper isolation and will link all packages built against its own
+# seed libraries instead of against libraries installed to /tmp/stage1root.
+build_stage1() {
+    local stage1_repos="$TEMPDIR/stage1-ebuild-repos"
+
+    write_configs
+
+    # use known-good gentoo base repo from SDK, coreos-overlay from manifest
```
krnowak, line 99: Isn't this gentoo base repo being our portage-stable?

t-lo: Yes; it is generally referred to as the "gentoo base" repo in the SDK scripts (and also in the gentoo catalyst / bootstrap tooling).

marga-kinvolk: I don't understand the code... Why is coreos-overlay called out explicitly, but portage-base not?

bootstrap_sdk
```diff
@@ -83,6 +83,54 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
...
+
+    local overlay_revision=$(
+        grep 'name="kinvolk/coreos-overlay"' /mnt/host/source/.repo/manifests/default.xml \
+        | sed 's/.*revision="refs\/\(heads\/\)\{0,1\}\([^"]\+\)".*/\2/' )
+
+    info "Using coreos-overlay revision '$overlay_revision'"
+
+    mkdir "$stage1_repos/coreos-overlay"
+    ( cd "$stage1_repos/coreos-overlay" \
+         && git clone https://github.com/kinvolk/coreos-overlay.git . \
```
marga-kinvolk, line 113: Why is this clone needed? Shouldn't it already be available in the build chroot?

bootstrap_sdk
```diff
@@ -83,6 +83,54 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
...
+         && git checkout "$overlay_revision" )
+
+    catalyst_conf > "$TEMPDIR/catalyst-stage1.conf"
+    sed -i "s:^portdir.*:portdir=\"$stage1_repos/gentoo\":" \
```
marga-kinvolk, line 117: Please add a comment explaining why this is needed

bootstrap_sdk
```diff
@@ -83,6 +83,54 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
...
+         "$TEMPDIR/catalyst-stage1.conf"
+
+    catalyst $DEBUG -c "$TEMPDIR/catalyst-stage1.conf" -s "$FLAGS_version-stage1"
+
+    sed -i -e "s/^snapshot:.*/snapshot: $FLAGS_version-stage1/" \
```
marga-kinvolk, line 122: Please add a comment explaining why this is needed

marga-kinvolk (commented):
This is very complex and has required a lot of time spent on this, please add a lot more comments explaining why things are needed and what's going on behind the scenes.  Thanks!

bootstrap_sdk
t-lo: I agree it should, but it currently is not. I'm planning to address this in the near future - it's a super low-hanging fruit - but currently, coreos-overlay is not included in the chroot, only portage-stable is (in `/var/lib/gentoo/repos/gentoo`).

bootstrap_sdk
t-lo: Done, thank you for the feedback!

bootstrap_sdk
t-lo: Done - pardon for creating confusion.

t-lo:
> This is very complex and has required a lot of time spent on this, please add a lot more comments explaining why things are needed and what's going on behind the scenes. Thanks!

Thank you very much for your review @marga-kinvolk ; I've added more comments and slightly restructured `build_stage1` for clarity.

bootstrap_sdk
```diff
@@ -83,6 +83,87 @@ mkdir -p "${ROOT_OVERLAY}/tmp"
...
+# - the coreos-overlay repo version listed in the current SDK's manifest.
+#    Unfortunately coreos-overlay is not included in the SDK chroot at all; this
+#    will be addressed in the future.
+#
+# "Current SDK" refers to the SDK we entered with 'cork enter', i.e. the SDK
+#   we run ./bootstrap_sdk in.
+#
+# Using ebuilds from the above mentioned sources will ensure that stage 1 builds
+#  a minimal stage 2 from known-good ebuild versions - the same ebuild versions
+#  were used to build the very SDK ware currently in.
```
marga-kinvolk, line 102:
```suggestion
#  were used to build the very SDK we're currently in.
```

But even after fixing the typo this is confusing, "we" are not in an SDK. Maybe "were used to build the SDK that is used to run the bootstrap_sdk script"?

bootstrap_sdk
marga-kinvolk: Ok, it's not in /var/lib/gentoo/repos, but it is in `/mnt/host/source`, no?

marga-kinvolk (commented):
Thanks, it's clearer now, although I still don't get why you need to git clone coreos-overlay

bootstrap_sdk
t-lo: Not necessarily. Consider the following:
- User creates local SDK by means of `cork create` / `cork enter`
- User updates packages in `third_party/coreos-overlay`
- User rebuilds SDK
In the above case, updated packages would "leak" into the SDK build. It's not perfect - and it does not cover all cases - but it's the best we can do until we add a copy of coreos-overlay to the SDK tarball.

bootstrap_sdk
```diff
@@ -10,20 +10,32 @@
...
+#    - a fresh check-out of coreos-overlay at the revision used to build the SDK
+#
 # 3. stage2: Run portage-stable/scripts/bootstrap.sh
-#    This rebuilds the toolchain. Probably not strictly necessary most of
-#    the time but does super-duper-promise that the toolchain isn't linked
+#    This rebuilds the toolchain using Gentoo bootstrapping, ensuring it's not linked
 #    to or otherwise influenced by whatever was in the "seed" tarball.
-# 4. stage3: Run emerge -e system to rebuild everything using the fresh
-#    toolchain using the normal USE flags provided by the profile. This
+#    The toolchain rebuild may contain updates package ebuilds from
```
marga-kinvolk, line 29:
```suggestion
#    The toolchain rebuild may contain updated package ebuilds from
```

bootstrap_sdk
marga-kinvolk: I see, I'm not sure if this interferes with CI builds, I think from what you're saying that it doesn't, it's only for local user builds, but ok.

t-lo:
Added a commit to save a snapshot of stage4's coreos-overlay to the SDK tarball.
This will allow us - once a new SDK has officially been released - to use both `/var/gentoo/repos/gentoo` (aka portage-stable) and `/var/gentoo/repos/coreos-overlay` from seed tarballs in stage1, in a future change/PR.

sayanchowdhury (commented):
Changes looks good to me.

t-lo:
We now have some initial high-level documentation available on the SDK build process: https://kinvolk.io/docs/flatcar-container-linux/latest/reference/developer-guides/sdk-bootstrapping/

dongsupark (approved):

sayanchowdhury (approved):

PR #124, opened 2021-04-22
bootstrap_sdk: use local SDK ebuilds for stage1 (part 2)

t-lo:
This change uses portage-stable and coreos-overlay from the local SDK chroot (from /var/lib/gentoo/repos) in the stage 1 SDK bootstrap build. This is part 2 of the SDK bootstrap stage 1 fix (part 1 is covered in #123 ), which ensures stage 1 does
not introduce any changes in its ebuilds over the seed SDK.

The change also introduces an option to consciously divert from the above enforcement (i.e. **do** introduce a change in the stage1 ebuilds) by use of command line parameters:
```
--stage1_overlay_ref <gitref> will check out coreos-overlay and use <gitref> for stage 1 instead of the
                                local SDK's /var/lib/gentoo/repos/coreos-overlay

--stage1_portage_ref <gitref> will check out portage-stable and use <gitref> for stage 1 instead of the
                                local SDK's /var/lib/gentoo/repos/gentoo
```

If any of the two options is set then an update is forced in the stage1 seed before stage 1 is built. These options come in handy to force-upgrade a package in the stage1 seed, e.g. when fixing a flaw in the seed or to work around un-published sources (like openssl-1.1.1j recently).

# How to use

```shell
$ sudo ./bootstrap_sdk
```

# Testing done

- Built SDK locally (all stages, both arm64 and amd64)
- CI build running.

### DIFF
--- a/bootstrap_sdk
+++ b/bootstrap_sdk
@@ -11,33 +11,36 @@
 # For reference the procedure it performs is this:
 # 
 # 1. snapshot: Grab a snapshot of the portage-stable repo from
-#    third_party/... . coreoy-overlay is used as-is from the same source.
+#     the current SDK's /var/lib/gentoo/repos/gentoo.
+#     Alternatively, check out a git ref specified via --portage-ref.
 #
 # 2. stage1: Using a "seed" tarball as a build environment, build a
-#    minimal root file system into a clean directory using ROOT=...
-#    and USE=-* The restricted USE flags are key be small and avoid
-#    circular dependencies. Tarball is built from the current SDK's
-#    portage-stable and coreos-overlay instead of the ones from stage1
-#    since stage1 must not contain updated ebuilds (see build_stage1 below).
-#    This stage uses:
-#    - portage-stable from the SDK's /var/lib/gentoo/repos/gentoo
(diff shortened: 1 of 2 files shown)


bootstrap_sdk
```diff
@@ -99,41 +109,63 @@ cp "${BUILD_LIBRARY_DIR}/toolchain_util.sh" "${ROOT_OVERLAY}/tmp"
...
+#  a git ref known to work for stage1. In that case the stage1 seed (i.e. the seed SDK)
+#  will be updated prior to starting to build stage 2.
+
+stage_repo() {
+    local repo="$1"
+    local gitref="$2"
+    local dest="$3"
+    local gitname="$repo"
+
+    [ "$gitname" = "gentoo" ] && gitname="portage-stable"
```
pothos, line 139: an `if` statement clearer. In case someone adds `set -e` and `gitname` is not `gentoo`, the execution will stop at this command because the return code is not 0

pothos (approved):

bootstrap_sdk
t-lo: Fixed, good catch!

PR #125, opened 2021-06-25
build_library/break_dep_loop: add to library

t-lo:
This change moves `break_dep_loop` from `build_packages` into its own library script so it can be used from `build_packages` and `build_toolchains` / `toolchain_util.sh`.

`toolchain_util.sh` now calls break_dep_loop to break a dependency loop between util-linux and systemd which was introduced by the glibc-2.33 update. toolchain_util.sh further modifies USE flags to work around build breakage; both is done in install_cross_libs.

This PR is based on @dongsupark 's good work, here: https://github.com/kinvolk/flatcar-scripts/compare/dongsu/libpcre-8.44

### DIFF
--- a/build_library/break_dep_loop.sh
+++ b/build_library/break_dep_loop.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Based on work (c) 2011 The Chromium OS Authors.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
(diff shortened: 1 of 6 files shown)


t-lo:
There might be a much easier way to fix the breakage while also removing unneeded work from the toolchains build (significantly improving the overall build time); see [t-lo/update-to-glibc-2.33-simple](https://github.com/kinvolk/flatcar-scripts/tree/t-lo/update-to-glibc-2.33-simple)

t-lo:
Closed inf favour of the much simpler https://github.com/kinvolk/flatcar-scripts/pull/126.

PR #133, opened 2021-07-16
disk_layout: devcontainer: use architecture agnostic part uuid

jepio:
# disk_layout: devcontainer: use architecture agnostic part uuid

The previously used partition uuid for the rootfs partition of the devcontainer is x86 specific.
Change the partition uuid to one that works for both arches.

## How to use

```
./build_image --board=arm64-usr container
```

## Testing done

Ran the developer container on aarch64.
```
systemd-nspawn --image=flatcar_developer_container.bin
```

### DIFF
--- a/build_library/disk_layout.json
+++ b/build_library/disk_layout.json
@@ -130,7 +130,7 @@
       "9":{
         "label":"ROOT",
         "fs_label":"ROOT",
-        "type":"4f68bce3-e8cd-4db1-96e7-fbcaf984b709",
+        "type":"0fc63daf-8483-4772-8e79-3d69d8477de4",
         "blocks":"12582912"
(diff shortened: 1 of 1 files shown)


jepio:
@tormath1 is there any CI jobs for the dev container that I should run?

sayanchowdhury:
You could do a manifest job run to run the [kola/dev-container.sh](https://github.com/kinvolk/flatcar-scripts/blob/main/jenkins/kola/dev-container.sh) script

jepio:
CI passed at last 👍 http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/dev-container/2134/

dongsupark (approved):

PR #135, opened 2021-07-26
jenkins/kola/qemu(_uefi): run update test from previous release

pothos:
- jenkins/kola: share a single qemu script file
- jenkins/kola/qemu(_uefi): run update test from previous release
    The newly enabled update test performs an update from the built image
    to itself. This is useful to test that the update mechanism didn't
    break but it doesn't say if the built image will be accepted as update
    from the previous official release.
    
    Introduce an additional kola run that begins from the previous official
    release and tests to update to the built image. Since the test does two
    updates it also covers the case of updating from the built image to the
    built image. Thus, we can skip the test in the normal run.
    This new kola run is done first to keep the qemu-latest symlink valid
    for the main test suite.

## How to use/testing done

The test runs successfully and the TAP reports and slack reports work as usual, there is just an additional TAP test table for the separate run.
http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3129/cldsv/


tormath1:
Awesome, I just have one concern: now we have a proper way to test update from _n-1_ to _n_ we can maybe skip the current `cl.update.payload` who's testing from _n_ to _n_ ? https://github.com/kinvolk/flatcar-scripts/blob/aa6c6c6a3f2750957253cb321d37ec22f2cb6712/jenkins/kola/qemu_common.sh#L101

pothos:
Both are important, also the `n` to version `n` test because it ensures that the new image has a working mechanism to process the new payload. The way the kola test is currently implemented actually includes updating twice, and we could indeed skip the test that is part of the whole test suite, so yes, skipping the current test would work.
Maybe it's still useful to include it, e.g., when someone runs kola locally and only chooses to run it from the built image? (But we could achieve this by adding a comment, too.)

pothos:
Updated it to skip the existing test.

tormath1 (approved):
LGTM, thanks !

pothos:
Will backport it to all channels

PR #137, opened 2021-07-28
build_image_util.sh: delete vmlinuz file from /usr partition

pothos:
The vmlinuz kernel image gets installed to /usr/boot/ but isn't usable
for dm-verity until it gets copied over to /boot/flatcar/ and the hash
gets embedded at a particular offset. The file in /usr/boot/ uses space
while it's not having a real purpose as long as dm-verity is used.

Delete the vmlinuz file under /usr/boot/ to free up space.

## How to use

Build an image and run all tests

→ 203 MB free on `/usr`

## Testing done

http://localhost:9091/job/os/job/manifest/3150/cldsv/

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -597,10 +597,13 @@ finish_image() {
     *) disable_read_write=${FLAGS_FALSE} ;;
   esac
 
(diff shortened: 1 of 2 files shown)


dongsupark (approved):
Good catch, Thanks!

PR #141, opened 2021-08-12
build_library: support multi-arch in generate_au_zip

dongsupark:
To be able to support arm64 native SDK without cross builds, we should make `generate_au_zip` support both architectures, amd64 and arm64.

Without doing that, `build_image` fails with `ERROR : Required WHITE_LIST items ld-linux-x86-64.so.2 not found!!!`, because the script recognizes only amd64 libs in WHITE_LIST.

We should first determine the architecture in build_image, before running `generate_au_zip`, and pass the architecture, either amd64 or arm64.
Also add `allow_list` and `ld_linux` parameters to necessary functions.
Replace `black list` with `deny list`, `white list` with `allow list`.

## How to use

```
...
./build_image
```

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3281/cldsv

### DIFF
--- a/build_library/generate_au_zip.py
+++ b/build_library/generate_au_zip.py
@@ -32,6 +32,9 @@ DYNAMIC_EXECUTABLES = ['/usr/bin/delta_generator',
                        '/usr/bin/bsdiff',
                        '/usr/bin/bspatch']
 
+LD_LINUX_AMD64 = 'ld-linux-x86-64.so.2'
+LD_LINUX_ARM64 = 'ld-linux-aarch64.so.1'
+
 # These files will be ignored when present in the dependancy list.
 DENY_LIST = [
     # This library does not exist on disk, but is inserted into the
@@ -40,9 +43,14 @@ DENY_LIST = [
     ]
 
 # These files MUST be present in the dependancy list.
-ALLOW_LIST = [
-    # Update WrapExecutableFiles if this file changes names
-    'ld-linux-x86-64.so.2',
+# Update WrapExecutableFiles if this file changes names.
+# Each architecture requires different allow list libs.
+ALLOW_LIST_AMD64 = [
+    LD_LINUX_AMD64,
+    ]
+
+ALLOW_LIST_ARM64 = [
+    LD_LINUX_ARM64,
     ]
 
 LIB_DIR = 'lib.so'
@@ -93,7 +101,7 @@ def _SplitAndStrip(data):
   return return_list
 
 
-def DepsToCopy(ldd_files):
+def DepsToCopy(ldd_files, allow_list):
   """Returns a list of deps for a given dynamic executables list.
     Args:
       ldd_files: List of dynamic files that needs to have the deps evaluated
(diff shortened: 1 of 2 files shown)


build_library/generate_au_zip.py
```diff
@@ -335,6 +343,8 @@ def main():
                     default='au-generator.zip', help='Name of the zip file')
   parser.add_option('-k', '--keep-temp', dest='keep_temp', default=False,
                     action='store_true', help='Keep the temp files...',)
+  parser.add_option('-a', '--arch', dest='arch',
+                     default='amd64', help='Arch Default: amd64',)
```
pothos, line 347:
```suggestion
                     default='amd64', help='Arch amd64/arm64. Default: amd64',)
```
Maybe optparse can do choices? Anyway it helps the user to know the possible values.

pothos (approved):

build_library/generate_au_zip.py
dongsupark: Changed so

dongsupark:
CI passed

build_library/generate_au_zip.py
```diff
@@ -93,11 +101,11 @@ def _SplitAndStrip(data):
   return return_list
 
 
-def DepsToCopy(ldd_files):
+def DepsToCopy(ldd_files, allow_list):
   """Returns a list of deps for a given dynamic executables list.
     Args:
       ldd_files: List of dynamic files that needs to have the deps evaluated
-      black_list: List of files that we should ignore
+      deny_list: List of files that we should ignore
```
krnowak, line 108: This looks wrong - it should be talking about `allow_list`. I suppose that in the past the function has a `black_list` parameter that was later dropped in favor of `BLACK_LIST` global variable, but the docstring wasn't updated.

build_library/generate_au_zip.py
```diff
@@ -129,12 +137,12 @@ def DepsToCopy(ldd_files):
...
 
-  result = _ExcludeBlacklist(list(libs), BLACK_LIST)
-  _EnforceWhiteList(list(libs), WHITE_LIST)
+  result = _ExcludeDenylist(list(libs), DENY_LIST)
+  _EnforceAllowList(list(libs), allow_list=allow_list)
   return result
 
 
-def CopyRequiredFiles(dest_files_root):
+def CopyRequiredFiles(dest_files_root, allow_list):
```
krnowak, line 145: Could you add `allow_list` to the Args section in the docstring below?

build_library/generate_au_zip.py
```diff
@@ -277,17 +285,17 @@ def _ExcludeBlacklist(library_list, black_list=[]):
...
+def _EnforceAllowList(library_list, allow_list=[]):
+  """Deletes the set of files from deny_list from the library_list
     Args:
-      library_list: List of the library names to filter through black_list
-      black_list: List of the black listed names to filter
+      library_list: List of the library names to filter through deny_list
+      deny_list: List of the deny listed names to filter
     Returns:
       Filtered library_list
   """
```
krnowak, line 295: This whole doc string is a copy-pasta from `_ExcludeDenylist` and not even updated for `_EnforceAllowList`, should be more like (can't do a suggestion for the deleted lines, github tells me):
```python
  """Ensures that library_list contains all the items from allow_list
    Args:
      library_list: List of the library names to check
      allow_list: List of the items that ought to be in the library_list
  """
```

krnowak (commented):
I only spotted some docstring issues.

build_library/generate_au_zip.py
dongsupark: Done

build_library/generate_au_zip.py
dongsupark: Done

build_library/generate_au_zip.py
dongsupark: Done

krnowak (approved):

PR #142, opened 2021-08-18
Port some stuff to python3

krnowak:
This PR drops the `check_out_of_date.py` script - it was already broken (a python2 script importing a python3 portage module). And I have one less script to port.

All the python scripts in `build_library` are ported to python3. The scripts in `lib` are left intact for now - they are only imported by `coreos-dev/cros-devutils` with which I'm not entirely sure what to do yet.

CI: http://localhost:9091/job/os/job/manifest/3302/cldsv/ (this uses a different branch, where I dropped some files from `lib`, before I realized that those are actually used, but the ported scripts are the same).


krnowak:
CI passed.

build_library/generate_au_zip.py
```diff
@@ -153,7 +155,7 @@ def CopyRequiredFiles(dest_files_root, allow_list):
     sys.exit(1)
 
   all_files = DYNAMIC_EXECUTABLES + STATIC_FILES
-  all_files = map(os.path.expanduser, all_files)
+  all_files = list(map(os.path.expanduser, all_files))
```
jepio, line 158: in line 166 (can't select in the review): is that a lonely `all_files` statement?

jepio (approved):
Looks good

build_library/generate_au_zip.py
krnowak: Huh, seems to be. I'll drop it.

sayanchowdhury: Yeah! seems like so!

PR #143, opened 2021-08-19
*: Use the internal copy of cros_workon

krnowak:
The scripts that invoked `cros_workon` without specifying a path to
the script were not calling the internal `cros_workon` directly, but
rather a copy installed in `/usr/bin/cros_workon`.
`/usr/bin/cros_workon` comes from the `coreos-base/cros-devutil` and
is a wrapper script that sources `common.sh` file to figure the
location of the `scripts` and finally invokes the internal
`cros_workon`. Curious thing is that the sourced `common.sh` comes
from the `/usr/lib/crosutils` directory and contents of the directory
come from the `dev-util/crosutils` package. And that `common.sh` is
different from the one in the scripts directory, but fortunately the
part that detects the path to the `scripts` directory is the same. I'm
not sure where where exactly the copy of `common.sh` in
`/usr/lib/crosutils` comes from - likely from somewhere in
`https://chromium.googlesource.com/chromiumos/platform/crosutils`.

Just cut the middle layers and call the internal copy of `cros_workon`
directly.

CI: http://localhost:9091/job/os/job/manifest/3332/cldsv/

### DIFF
--- a/build_packages
+++ b/build_packages
@@ -166,7 +166,7 @@ fi
 # Build cros_workon packages when they are changed.
 CROS_WORKON_PKGS=()
 if [ "${FLAGS_workon}" -eq "${FLAGS_TRUE}" ]; then
-  CROS_WORKON_PKGS+=( $(cros_workon list --board=${FLAGS_board}) )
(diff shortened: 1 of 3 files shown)


krnowak:
CI passed (os/kola/qemu_uefi tests failed, but those doesn't seem to be related).

PR #147, opened 2021-08-27
SDK container

t-lo:
This is the start of work aiming to create an SDK container and to make it usable as Flatcar's default SDK.
(also for CI builds etc.)

Usage
- `./bootstrap_sdk_container 3005.0.0` - bootstraps a new SDK tarball with the SDK version in `version.txt`, using seed SDK container 3005.0.0
- `./build_sdk_container_image <tarball>` - builds an SDK container image from `<tarball>`
- `./run_sdk_container  -v 2969.0.0`  - creates an SDK container from image 2969.0.0. The container will stop when the user exits and can subsequently be restarted using the same command.

Goals
- Generate SDK container straight from an SDK tarball w/o intermediate steps
- Allow work on host repos flatcar-scripts, coreos-overlay, and portage-stable via container
- ease versioning, remove need for manifest
- remove need for `repo` tools

Design choices:
- The SDK version is bound to the flatcar-scripts repo
  - `sdk_container/.repo/manifests/version.txt` contains the current SDK version
  - coreos-overlay and portage-stable repos are git modules of flatcar-scripts
  - aims to ease versioning as a simple check-out of flatcar-scripts will ensure the version string matches the ebuild repos used to build that version
- local `__build__`directory which is bind-mounted into container, will contain OS images and bootstrapped SDK tarballs
  - `__build__` is in `.gitignore` 

Features:
- Stand-alone (w/o host repos) and integrated (w/ "scripts") host repo operation modes
- docker image imitates "cork enter", user is placed in `~/trunk/src/scripts/`; gpg and ssh agent forwarding are supported
- SDK container can build packages, OS images, and bootstrap SDKs (i.e. create new tarballs)
- Lightweight integration with CI via scripts in `ci-automation/` sub-folder

Notable divergence from current SDK (chroot):
- The container does not use the `repo` tool. All `cros`-managed repos pull from github instead of local repos.
- The container does not use manifest.xml

# TODOs from design review
- ~[ ] Push versioned commits to branches, don't just push tags.​~
- [x] Add documentation for submodules, and for updating local tags​
- [x] Accept "main-" as version prefix just like alpha-, beta-, stable- etc.
- [ ] Publish the nightly SDK on GHCR.IO​
- [x] Run containerised build side-by-side regular nightly​


sdk_lib/sdk_container_common.sh
```diff
@@ -0,0 +1,68 @@
...
+# Check out ebuild repos tag/branch/commit-ish.
+# Must be called from the script root.
+
+function checkout_ebuild_repos() {
+    local version="$1"
+
+    yell "Checking out portage-stable, coreos-overlay tags"
+    for repo in portage-stable coreos-overlay; do
+		cd "sdk_container/src/third_party/$repo"
+		git fetch --all
```
jepio, line 56: The git-submodule part is the one thing that I don't like about this. Can we try to avoid them please, they don't work well to track dependencies at all.
We need the repos during container build and then later during usage. At container build we can definitely clone, during usage I would continue relying on a slimmed down repo manifest.

t-lo: I understand your concern - submodules are often rather painful to use. Let's put this up for discussion for the time being.

The motivation here is to make `flatcar-scripts` the authoritative (top-level) repo for any given SDK and OS image version - removing the need for manifests altogether.
A usage example would be to clone `flatcar-scripts`, check out a specific version tag (say, 2969.0.0), and automatically get the respective `portage-stable`and `coreos-overlay` states.
Submodules help with automating this since the `flatcar-scripts` tag will have submodules point to the respective commits of `portage-stable`and `coreos-overlay`.
An alternative (though softer) version of the above would be to also tag both `portage-stable`and `coreos-overlay` repos with 2969.0.0 at release time.

t-lo: I've updated the "Simplify SDK sources" roadmap item with a design proposal to better define the goals: https://github.com/kinvolk/Flatcar/issues/440

sdk_lib/sdk_container_common.sh
```diff
@@ -0,0 +1,104 @@
...
+
+		tag=$(git tag -l | grep "${version}" || true)
+		if [ -z "$tag" ]; then
+			echo "ERROR: Unable to find tag for version '$version' in repo '$repo'"
+			exit 1
+		fi
+
+	    yell "$repo: $tag"
+		git checkout "$tag"
```
pothos, line 101: I think this should use the submodule commit. Would it work to make a symlink to the submodule checkout instead of checking out another version?

sdk_lib/sdk_container_common.sh
pothos: I think submodules are ok as long as the scripting takes care of using the right commit and not a stale checkout. We could implement the same again with text files but I don't think this helps much.
We need a single place to tie the commits of the three repos together. If we do it here as submodules, it allows us to start automatic builds from PRs.

run_sdk_container
```diff
@@ -0,0 +1,94 @@
...
+
+stat="$(docker ps --all --no-trunc --filter name="$name" --format '{{.Status}}'\
+        | cut -f1 -d' ')"
+
+if [ -z "$stat" ] ; then
+    yell "Creating a new container '$name'"
+    exec docker run -ti \
+       -v /dev:/dev \
+       -v "$(pwd)/sdk_container:/mnt/host/source/" \
+       -v "$(pwd)/__build__/images:/mnt/host/source/src/build" \
```
tormath1, line 71: For `__build__/images` we can create this directory automatically (or warn the user that this directory does not exist):
```
+ exec docker run -ti -v /dev:/dev -v /home/mathieu/github/kinvolk/flatcar-scripts/sdk_container:/mnt/host/source/ -v /home/mathieu/github/kinvolk/flatcar-scripts/__build__/images:/mnt/host/source/src/build -v /home/mathieu/github/kinvolk/flatcar-scripts:/mnt/host/source/src/scripts --privileged --network host -e SDK_USER_ID=1000 -e SDK_GROUP_ID=1001 --name=flatcar-sdk-all-2969.0.0 --hostname flatcar-sdk-all-2969.0.0 flatcar-sdk-all:2969.0.0 /mnt/host/source/sdk_entry.sh
Error: statfs /home/mathieu/github/kinvolk/flatcar-scripts/__build__/images: no such file or directory
```

sdk_lib/sdk_container_common.sh
t-lo: Hmmm, good point. We would need to check out the respective `flatcar-scripts` tag, then update the submodules.

run_sdk_container
t-lo: Good catch! This exists in my local workspace but I forgot to validate with a fresh checkout of `flatcar-scripts`. Will add.

tormath1: Thanks ! I wanted to test compatibility with `podman` too - in case we were using some `docker` edge cases command / options and, so far, so good !

t-lo:
I've addressed the comments by @pothos and @tormath1 , and added 2 features:
- upload containers to ghcr.io (assumes docker is authenticated with github)
- add a script to initialise a self-contained SDK container, i.e. without flatcar-scripts repo. The script clones the correct versions of scripts and ebuild repos directly into the container. While this is not the preferred use it arguably adds some freedom.

t-lo:
**NOTE** because of the recent change of using [submodules for version pinning](https://github.com/kinvolk/flatcar-scripts/pull/147/commits/0ba5ded822dd419052a0ee7a35d2873d9b1605e4) you might run into issues when trying to test this branch.
As a workaround you need to trick the container scripts into thinking this branch's tip is the `alpha-2969.0.0` release:
```
$ git checkout t-lo/sdk-container
$ git tag --force alpha-2969.0.0
```
Please make sure to not push your tags to origin after that ;-)

t-lo:
I've updated the download section of `build_sdk_container_image` to download to `sdk_container/.cache/sdks` so the container can re-use the tarball (e.g. when bootstrapping a new SDK).
The download is skipped if the file exists.
Also, checksum and signature are downloaded now, too, and the tarball is verified.

build_sdk_container_image
```diff
@@ -0,0 +1,145 @@
...
+             .
+
+for a in all arm64 amd64; do
+    yell "Creating '$a' arch SDK image"
+    rmarch=""
+    case $a in
+        arm64) rmarch="amd64-usr";;
+        amd64) rmarch="arm64-usr";;
+    esac
+    docker build -t "$sdk_container_common_registry/flatcar-sdk-${a}:${version}" \
```
tormath1, line 174: Out of curiosity, is there something preventing us to merge the following Dockerfiles: `sdk_lib/Dockerfile.sdk-container` and `sdk_lib/Dockerfile.lean-arch` ?

t-lo: I was thinking about pipelining / paralellising the tasks later : the 3 docker build commands producing the final images can run in parallel, while the first 2 (importing the tarball and installing the BSPs) cannot.

t-lo: Actually, let me add a check for whether the intermediate containers exist already and then skip the respective steps conditionally.

t-lo:
Rebased to Alpha-2983.0.0.

run_sdk_container
```diff
@@ -0,0 +1,95 @@
...
+fi
+
+stat="$(docker ps --all --no-trunc --filter name="$name" --format '{{.Status}}'\
+        | cut -f1 -d' ')"
+
+if [ -z "$stat" ] ; then
+    yell "Creating a new container '$name'"
+    mkdir -p "__build__/images"
+    exec docker run -ti \
+       -v /dev:/dev \
```
tormath1, line 70: On my system, `docker` is aliased with `podman` since most of the `docker` commands / flags are supported by `podman`.

Running `./run_sdk_container`, I got the following:
```
+ exec docker run -ti -v /dev:/dev -v /home/mathieu/github/kinvolk/flatcar-scripts/sdk_container:/mnt/host/source/ -v /home/mathieu/github/kinvolk/flatcar-scripts/__build__/images:/mnt/host/source/src/build -v /home/mathieu/github/kinvolk/flatcar-scripts:/mnt/host/source/src/scripts --privileged --network host -e SDK_USER_ID=1000 -e SDK_GROUP_ID=1001 --name=flatcar-sdk-all-2983.0.0 --hostname flatcar-sdk-all-2983.0.0 ghcr.io/flatcar-linux/flatcar-sdk-all:2983.0.0 /mnt/host/source/sdk_entry.sh
Error: chown `/dev/pts/2`: Operation not permitted: OCI permission denied
```

Since `podman` is rootless, we need to pass the `/dev` as a `--device`:
```diff
     exec docker run -ti \
-       -v /dev:/dev \
+       --device /dev/:/dev/ \
```

pothos: I think this should even run as root, or? Maybe try `sudo docker`.

pothos: But maybe it really works to run in a user namespace and the root user gets mapped to your user ID and still can set up loopback devices and create network namespaces, haven't tried this.

pothos: But the goal here was to use the host's `/dev`, is this required?

pothos: I can't get `mknod /dev/loop0 b 7 0` work, I think `sudo docker` is required to have the same result in Podman and Docker. For Docker it makes no difference anyway.

By the way, with `sudo podman` `--privileged` implies `-v /dev:/dev` and I think that's also the case for Docker and it can be dropped, or?

t-lo: > But the goal here was to use the host's `/dev`, is this required?

It is for `build_image` (and likely `image_to_vm.sh`) because creating / modifying the OS image uses loopback mounts / devices. Without `--privileged`, access to `/dev/loop[x]` is prohibited, and without `-v /dev:/dev` partition re-scanning after modifications to the block device file does not work in the container, so individual OS image partitions cannot be accessed. A work-around for the latter would be to run udev in the container to update the container device nodes, something I'd rather like to avoid...

pothos: Ok, `sudo podman -v /dev:/dev --privileged` should be the same as `docker -v /dev:/dev --privileged`.

tormath1: Yes, it seems to work as expected with `sudo` :)

t-lo:
I found an issue with switching to the `sdk` user via the `USER` command in `Dockerfile.lean-arch`: `USER` only allows to specify a single group, but `sdk` needs to be in group `sdk` (mirroring the host user's group, for bind-mounting) and in group `portage`.

I've added an ENTRYPOINT trampoline script to take care of switching to the `sdk` user correctly.

run_sdk_container
```diff
@@ -0,0 +1,90 @@
...
+version="$(get_version_from_versionfile)"
+arch="all"
+checkout="true"
+create_only="false"
+name=""
+
+usage() {
+    echo "  $0 - Start an SDK container of a given SDK release version."
+    echo "       This will create the container if it does not exist, otherwise it will just start it."
+    echo "       If the container is already running then it will exec into the container."
```
pothos, line 21: Having to start and stop the container will make it a bit complicated to use in scripts and using a fixed container name even forbids concurrent usage. I think this needs to be reworked to be useful in the build pipeline with the goals for easy automatic cleanup of containers and concurrent usage.

pothos (changes requested):

sdk_lib/sdk_container_common.sh
```diff
@@ -0,0 +1,102 @@
...
+function yell() {
+    echo -e "\n###### $@ ######"
+}
+# --
+
+# Prevent the sourcetree from being used by multiple containers with different
+#  SDK release versions by use of a version lock.
+# The lock is self-cleaning and does not need to be removed in scripts calling
+# this function.
+function lock_version() {
```
pothos, line 24: I don't understand for which use cases the lock should primarily be working but it's not race free

run_sdk_container
t-lo: That's a good point - the script is biased towards interactive use. I'll add a `--oneshot` flag to make it better suited for scripting use.

sdk_lib/sdk_container_common.sh
t-lo: The lock is used to prevent changing the git tree to a different version when a container is running (since the git tree is volume-mounted into the container). It's probably too complex and can be simplified.
It should not be racy, though - could you please elaborate?

build_sdk_container_image
```diff
@@ -0,0 +1,202 @@
...
+# --
+
+if [ -z "$version" ] ; then
+    echo "ERROR: Flatcar SDK version to build container for cannot be autodetected and was not provided."
+    exit 1
+fi
+
+yell "\n######\n###### Building SDK container for version $version"
+
+if $checkout; then
```
pothos, line 71: I think this should always ensure that the submodule commits are matching because this is the tricky part of using submodules. Maybe this will require to do a local commit in flatcar-scripts but I think it's worth to be as strict as possible?

sdk_lib/sdk_container_common.sh
pothos: Two executions of this function can run in parallel up to the end where both will create the lock files

pothos: Even one can complete totally and the other is still at the `if [ -z "$lock_owner_active" ] ; then` branch and will remove the lock again before it writes its own lock

pothos: It also seems to be able to have two executions with the same version succeed to run and then there is only one owner, but a third execution will not know whether both have finished

build_sdk_container_image
t-lo: I'm currently reworking version handling so the git tree version is authoritative. The version file will always be generated.
If the commit has a tag, then I'll use the tag (`git tag --points-at HEAD`). If there's no tag then I'll use `git describe` output.

run_sdk_container
pothos: The Jenkins scripts make use of entering the SDK multiple times. Maybe we can simplify this but it would be nice to keep the behavior of being able to exec into the container but using it's temporary name (use the one allocated by the container runtime?). E.g., have the container being started with `docker run --rm` in the background so that killing the process will also remove the container (we already have a way of killing a job's processes in Jenkins). By saving the temporary container name to a local file we could know how to exec into it on the next run.

run_sdk_container
```diff
@@ -0,0 +1,90 @@
...
+       -v "$(pwd)/sdk_container:/mnt/host/source/" \
+       -v "$(pwd)/__build__/images:/mnt/host/source/src/build" \
+       -v "$(pwd):/mnt/host/source/src/scripts" \
+       --privileged \
+       --network host \
+       -e SDK_USER_ID="$(id -u)" \
+       -e SDK_GROUP_ID="$(id -g)" \
+       --name="$name" \
+       --hostname "flatcar-sdk-${arch}-${version}" \
+       "$sdk_container_common_registry"/flatcar-sdk-"${arch}:${version}"
```
pothos, line 80: We need a way to forward a list of environment variables, both here and in `/etc/sudoers.d/90_env_keep` (see https://github.com/kinvolk/mantle/blob/flatcar-master/sdk/create.go#L72)

pothos: There is also some mounting for ssh agent and gpg done: https://github.com/kinvolk/mantle/blob/flatcar-master/sdk/enter.go#L190

pothos: `CopyGoogleCreds` is probably also relevant to upload binary packages

pothos:
For reference, the `ApplyPatch` function is in https://github.com/kinvolk/mantle/blob/flatcar-master/sdk/repo.go#L173

sdk_lib/sdk_container_common.sh
t-lo: The directory is the lock, not the files.

The files give information about lock ownership (to figure out if the lock can safely be cleared)  to determine if the lock can safely be cleared. Ownership information is written only after acquiring the lock succeeded (in the worst of cases, a lock that could be cleared is not cleared).

pothos: The directory doesn't really function as lock because it's created once and not in an atomic way that also creates the files that the if branches check for

pothos: How about adding a comment that this is aimed for human usage and not race-free for automation. We can avoid running this in Jenkins as long as we kill the processes and remove any existing folders.

run_sdk_container
t-lo: We can provide a custom name via `-n` and make sure that's random. However, I feel Jenkins should not re-use containers but instead produce container images to then be re-used... feels cleaner.
Thinking about it, we might even add yet another Dockerfile for building binary packages and then publish a super-fat SDK which ships all the board packages precompiled.
Thanks for bringing up the `--rm` flag, that was missing from my (ongoing) oneshot implementation!

sdk_lib/sdk_container_common.sh
t-lo: Giving up for now on the multiple-containers-with-the-same-version feature and will make the locking simpler so it is safe :)

run_sdk_container
t-lo: Most of these can be passed via the `sdk_container`  volume mount (which ends up at `/mnt/host/source` in the container). Environment variables are trickier since these need to survive `sudo su ...` in the entrypoint. I'll come up with something.

run_sdk_container
pothos: The Jenkins jobs need to run set up an SDK container (=`cork create`) and then execute some commands in it (=`cork enter`). You already have this implemented here which is ok, but it needs to be improved a bit.

pothos: Killing the docker client doesn't actually remove the container with `run --rm`, something else is needed…

pothos: When there are Docker containers left over when canceling a Jenkins job, the build process in the Jenkins job will continue to write to the folder which can corrupt the state of the next job depending on how the folders are reused (and wastes RAM/CPU resources).

t-lo:
I've updated the sdk container logic to address most of @pothos ' comments:
- Simplified container locking to eliminate races
- reworked version handling so:
  - for SDK container builds, the git tag is authoritative for SDK version and OS version
  - for container runs, the git tag is authoritative for the OS version; SDK version is taken from the version file after checkout
- added code to pass environment variables to container runs via `sdk_container/.env`, added  COREOS_OFFICIAL passing to SDK build Dockerfile, and added `/etc/sudoers.d/90_env_keep` file to container

Wrt. the last item, I feel it's worth checking /iterating on what else from the remaining items `cork` passes to the chroot  we actually need.

sdk_lib/sdk_container_common.sh
```diff
@@ -17,40 +17,42 @@ function yell() {
...
+    fi
+
+    lock_owner="$(cat sdk_container/.lock/owner || true)"
+    lock_owner_active="$(docker ps --no-trunc \
+                              --filter name="$lock_owner" \
+                              --format '{{.Status}}')"
+
+    if [ -z "$lock_owner_active" ] ; then
+        yell "Removing stale SDK lock (owner: '$lock_owner')"
+        rm -rf sdk_container/.lock/
```
pothos, line 45: nit: maybe add a comment here that this is not race free when executed in parallel (two executions want to clear the lock), because this `rm` line is able to remove the newly created lock from in case the other execution already deleted it and recreated it again

run_sdk_container
pothos: My suggestion is that we start a background process `sleep infinity &`, remember the PID, and start the container with `docker run --rm -v /proc/$PID:/watchdir sh -c 'while ls /watchdir > /dev/null; do sleep 5; done'`
This will ensure that the container exits when the `sleep` process is killed.

sdk_lib/sdk_container_common.sh
t-lo: Nice catch! Looks like I need to lock the owner check, too.

t-lo:
> We need a way to forward a list of environment variables, both here and in /etc/sudoers.d/90_env_keep (see https://github.com/kinvolk/mantle/blob/flatcar-master/sdk/create.go#L72)
> There is also some mounting for ssh agent and gpg done: https://github.com/kinvolk/mantle/blob/flatcar-master/sdk/enter.go#L190

Finally found some time to get back to this - I've added respective features to `run_sdk_container`.
 
> CopyGoogleCreds is probably also relevant to upload binary packages
This is still up to do since I need to figure out what this actually does.

t-lo:
Addressed @pothos ' final comment:
> CopyGoogleCreds is probably also relevant to upload binary packages

Added a respective feature to `run_sdk_container`.

t-lo:
I'm in the process of uploading a new SDK container based on 3005.0.0 (`ghcr.io/flatcar-linux/flatcar-sdk-all:3005.0.0`) to our repo at https://github.com/orgs/flatcar-linux/packages/container/package/flatcar-sdk-all ; should be available in a few hours (uploading 7GiB takes a while) so you can test the PR.

dongsupark:
`build_sdk_container_image` sometimes fails with the following error, when `update-chroot` tries to rebuild `dev-lang/go:1.17.2` inside the Docker container.

```
Building Go cmd/dist using /usr/lib/go-bootstrap. (go1.5.3 linux/amd64)
 * /var/tmp/portage/sys-apps/sandbox-2.12/work/sandbox-2.12/libsandbox/trace.c:do_peekstr():125: failure (Operation not permitted):
 * ISE:do_peekstr:process_vm_readv(105779, 0x00007ffc7464c5b0{0x00007f2d7be07010, 0x580}, 1, 0x00007ffc7464c5c0{0x000000c820012a80, 0x580}, 1, 0) failed: Operation not permitted
```

It looks like the same issue as https://github.com/gentoo/gentoo-docker-images/issues/98.
A [recommented way](https://github.com/moby/moby/issues/1916#issuecomment-530278444) to resolve this issue is to run `docker buildx build --allow security.insecure`.

t-lo:
So, do folks think we can get this merged? I realise there are open TODOs but having it in main it would be helpful to set up a nightly `main` build. Changes in this branch should not affect the "traditional" build, things would co-exist nicely.

t-lo:
Rebased on latest main.

t-lo:
I've built an SDK container for 2905.2.6 and pushed it to ghcr.io just in case someone wants to test this with stable. I've built the container from a 2905.2.6 SDK tarball but using this PR branch for `scripts` (which is effectively `main` with my changes on top). We can consider cherry-picking my changes to `flatcar-2905` or branch off a new `sdk-2905` from `main` (after PR merge), shouldn't matter much.

run_sdk_container
t-lo: Implemented + tested in Jenkins.

sdk_lib/sdk_container_common.sh
t-lo: Removed locking entirely since it was more of a hindrance than a help.

run_sdk_container
t-lo: Added creds.

build_sdk_container_image
```diff
@@ -0,0 +1,245 @@
...
+
+    # remove sdk tarball from scripts root so it's not part of the SDK container build context
+    if [ -f "${tarball_copied}" ] ; then
+        rm "${tarball_copied}"
+    fi
+
+    docker container rm "${toolchains_container}"
+
+    host_ip="$(ip addr show docker0 | grep -Po 'inet \K[\d.]+')"
+    binhost_port="$((1000 + (RANDOM % 55000) ))"
```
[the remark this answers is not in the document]
t-lo: Leaving it on random for now, will improve if we experience CI issue.

t-lo:
When merging, this PR should be cherry-picked to
* flatcar-2905
* flatcar-2983
* flatcar-3033
Respective adjustments need to be made to the git submodules in these branches to pull in the correct version, and to `sdk_container/.repo/manifests/version.txt` to reflect the correct Flatcar and SDK versions.

This will enable us to support containerised builds for all current release branches. Respective SDK container image versions have already been made available on https://github.com/orgs/flatcar-linux/packages.

.gitignore
```diff
@@ -8,3 +8,16 @@
...
+
+# SDK container env passing helpers
+sdk_container/.env
+sdk_container/.sdkenv
+
+# build cache / artefacts directories
+__build__/
+sdk_container/.cache
+sdk_container/.config
+sdk_container/.lock
```
krnowak, line 23: This is probably not necessary any more.

README.md
```diff
@@ -2,4 +2,134 @@
 
 Welcome to the scripts repo, your starting place for most things here in the Flatcar Container Linux SDK. To get started you can find our documentation on [the Flatcar docs website][flatcar-docs].
 
+The SDK can be used to
+* patch or update applications or libraries included in the Flatcar OS image
```
krnowak, line 6:
```suggestion
* Patch or update applications or libraries included in the Flatcar OS image
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+* add or remove applications and / or libraries
```
krnowak, line 7:
```suggestion
* Add or remove applications and / or libraries
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+* Modify the kernel configuration and add or remove kernel modules included with Flatcar
+* Build OS images for a variety of targets (qemu, bare metal, AWS, Azure, VMWare, etc.)
+* And lastly, the SDK can be used to upgrade SDK packages and to build new SDKs
+
 [flatcar-docs]: https://docs.flatcar-linux.org/os/sdk-modifying-flatcar/
+
+# Using the scripts repository: submodules and tags
+
+The repository is meant to be the entry point for Flatcar builds and development.
+For building pakages, there are 2 addigional repositories, [coreos-overlay](https://github.com/flatcar-linux/) and [portage-stable](https://github.com/flatcar-linux/portage-stable), which contain all packages' `ebuild` (build configuration) files.
```
krnowak, line 17:
```suggestion
For building packages, there are 2 additional repositories, [coreos-overlay](https://github.com/flatcar-linux/) and [portage-stable](https://github.com/flatcar-linux/portage-stable), which contain all packages' `ebuild` (build configuration) files.
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+                             +------coreos-overlay
+                             +------portage-stable
+```
+
+When working with the scripts repo always make sure to initialise and to update these submodules; otherwise builds will break because build configuration is missing:
+```bash
+$ git clone https://github.com/flatcar-linux/scripts.git
+$ cd scripts
+$ git submodule init
+$ git submodule update
```
krnowak, line 34: I suppose this probably could be shortened to:
```suggestion
$ git clone --recurse-submodules https://github.com/flatcar-linux/scripts.git
```

krnowak: Or maybe even:
```suggestion
$ git clone --recurse-submodules --shallow-submodules https://github.com/flatcar-linux/scripts.git
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+
+# Using the SDK container
+
+We provide a containerised SDK via https://github.com/orgs/flatcar-linux/packages. The container comes in 3 flavours:
+* Full SDK initialised with both architectures supported by Flatcar (amd64 and arm64). This is the largest container, it's about 8GB in size (~3 GB compressed).
+* AMD64 SDK initialised for building AMD64 OS images. About 6GB in size (2GB compressed).
+* ARM64 SDK initialised for building ARM64 OS images on AMD64 hosts. Also about 6GB in size.
+While work on a native ARM64 native SDK is ongoing, it's unfortunately not ready yet. If you want to help, patches are welcome!
+
+The container can be run in one of two ways - "standalone", or integrated with the [Scripts](https://github.com/flatcar-linux/scripts) repo:
```
krnowak, line 58: I'd write "scripts" instead of "Scripts". Use of the capitalization is not consistent later.

```suggestion
The container can be run in one of two ways - "standalone", or integrated with the [scripts](https://github.com/flatcar-linux/scripts) repo:
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+* Standalone mode will use no host volumes and will allow you to play with the SDK in a sandboxed throw-away environment. In standalone mode, you interface with Docker directly to use the SDK container.
+* Integrated mode will closely integrate with the Scripts directory and bind-mount it as well as the portage-stable and coreos-overlay gitmodules into the container. Integrated mode uses wrapper scripts to interact with the SDK container. This is the recommended way for developing patches for Flatcar.
```
krnowak, line 60: Same here with capitalization.

```suggestion
* Integrated mode will closely integrate with the scripts repo directory and bind-mount it as well as the portage-stable and coreos-overlay gitmodules into the container. Integrated mode uses wrapper scripts to interact with the SDK container. This is the recommended way for developing patches for Flatcar.
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+Both the host's scripts repo as well as its submodules (portage-stable and coreos-overlay) are made available in the container, allowing for work on these repos directly.
+The wrapper scripts will re-use existing containers instead of creating new ones to preserve your work in the container, enabling consistency.
+
+To clone the scripts repo and pick a version:
+* Clone the scripts repo: `git clone https://github.com/flatcar-linux/scripts.git`
+  * Make sure to initialise and fetch git submodules - Flatcar's ebuilds are in 2 separate repositories, connected to `scripts` via submodules.
+  * `git submodule init; git submodule update`
+* Optionally, check out a release tag to base your work on
+  * list releases (e.g. all Alpha releases): `git tag -l alpha-*`
+  * check out the release version, e.g. `3033.0.0`: `git checkout 3033.0.0`
```
krnowak, line 99: I think it should more or less be:
- Clone the scripts repo
- Optionally, check our a release tag
- Initialize submodules

Initialization of submodules should happen after picking a tag as a base, since it could have different submodule checkouts.

README.md
```diff
@@ -2,4 +2,134 @@
...
+The local sourcetree can also be used with an entirely custom SDK container image. Users must ensure that the image is either fetch-able or present locally. The custom image can be specified using `-C <custom-image>`. This option is useful e.g. for building the local sourcetree with different SDK versions.
+
+Check out `./run_sdk_container -h` for more information on command line options.
+
+# Building a new SDK container
+
+Building an SDK container is done using `./build_sdk_container_image <tarball>`.
+The tarball input is the result of an SDK bootstrap (see below). Version information for both OS as well as for the SDK will be extracted from the tarball name.
+The version file will be updated accordingly before the SDK container is built.
+During the build, toolchain packages will be built and installed into the SDK container image. Both supported boards (`amd64-usr` and `arm64-usr`) will be initialisedin the container image.
```
krnowak, line 127:
```suggestion
During the build, toolchain packages will be built and installed into the SDK container image. Both supported boards (`amd64-usr` and `arm64-usr`) will be initialised in the container image.
```

README.md
```diff
@@ -2,4 +2,134 @@
...
+
+# Bootstrapping a new SDK tarball using the SDK container
+
+The script `./bootstrap_sdk_container` bootstraps a new SDK tarball using an existing SDK container and seed tarball. Specifying the seed version as well as the designated new SDK version is required for this script.
+
+# Automation stubs for continuous integration
+
+Script stubs for various build stages can be found in the [ci-automation](ci-automation) folder. THese are helpful for glueing Flatcar Container Linux builds to a continuous integration system.
```
krnowak, line 135:
```suggestion
Script stubs for various build stages can be found in the [ci-automation](ci-automation) folder. These are helpful for gluing Flatcar Container Linux builds to a continuous integration system.
```

krnowak (changes requested):
For now I only reviewed the docs.

sdk_lib/sdk_init_selfcontained.sh
```diff
@@ -0,0 +1,38 @@
...
+    local dest="$2"
+    local version="$3"
+
+    git clone https://github.com/flatcar-linux/$repo "$dest"
+    (
+        cd "$dest"
+        git fetch --all
+        local tag=$(git tag -l | grep "${version}")
+        git checkout "$tag"
+    )
```
krnowak, line 27: Nit: instead of a subshell, you could do `git -C "$dest" …`.

krnowak (commented):
Items that could be done later:

- maybe use `set -euo pipefail`
- introduce long options and use them in scripts

.gitignore
t-lo: Nice catch, I'll remove `.lock`.

bootstrap_sdk_container
```diff
@@ -0,0 +1,76 @@
...
+
+seed_version=""
+target_version=""
+vernum="$(strip_version_prefix "$target_version")"
+if is_official "$vernum" ; then
+    official="true"
+else
+    official="false"
+fi
+cleanup=""
```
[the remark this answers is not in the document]
t-lo: Very nice catch, this way we preserve whitespaces!

ci-automation/image.sh
```diff
@@ -0,0 +1,68 @@
...
+#        pushed to buildcache.
+#   2. "./ci-cleanup.sh" with commands to clean up temporary build resources,
+#        to be run after this step finishes / when this step is aborted.
+
+set -eu
+
+function image_build() {
+    local arch="$1"
+
+    source ci-automation/ci_automation_common.sh
```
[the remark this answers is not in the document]
t-lo: Yupp, I'm aware of the global scope here. Changing this to sourcing outside the function would make it easier to override things set in the sourced files - i.e., source image.sh, then re-set some of the variables in common or the versionfile, then call image_build(). I really would like to avoid giving people the opportunity to hack around like that :)

ci-automation/ci_automation_common.sh
```diff
@@ -0,0 +1,208 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
t-lo: We already have a "common.sh" in the scripts root and I really would like to prevent confusion.

ci-automation/vms.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# >>> This file is supposed to be SOURCED from the repository ROOT. <<<
+#
+# packages_build() should be called w/ the positional INPUT parameters below.
```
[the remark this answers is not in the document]
t-lo: LOL 😅

sdk_container/.repo/manifests/os-release
```diff
@@ -0,0 +1,6 @@
+NAME="Flatcar Container Linux by Kinvolk"
```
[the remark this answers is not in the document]
t-lo: The string is actually parsed by several 3rd party automation tools to detect what OS is used, unfortunately there is no easy way to change this.

sdk_container/.repo/manifests/version.txt
```diff
@@ -0,0 +1,4 @@
+FLATCAR_VERSION=3033.0.0
+FLATCAR_VERSION_ID=3033.0.0
+FLATCAR_BUILD_ID=""
+FLATCAR_SDK_VERSION=3033.0.0
```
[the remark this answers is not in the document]
t-lo: Yes, this file is updated by a number of scripts.

sdk_lib/Dockerfile.sdk-import
```diff
@@ -0,0 +1,52 @@
+ARG VERSION
+FROM flatcar-sdk-tarball:${VERSION}
+
+RUN touch /etc/debian_chroot
```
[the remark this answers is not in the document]
t-lo: It's a hack to make the "old" chroot detection work. Will add a comment.

sdk_lib/sdk_container_common.sh
```diff
@@ -0,0 +1,239 @@
...
+    ( source "$sdk_container_common_versionfile"; echo "$FLATCAR_SDK_VERSION"; )
+}
+# --
+
+# return true if a given version number is an official build
+#
+function is_official() {
+    local vernum="$1"
+
+    official="$(echo "$vernum" | sed -n 's/^[0-9]\+\.[0-9]\+\.[0-9]\+$/true/p')"
```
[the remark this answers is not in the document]
t-lo: Yes, absolutely!

sdk_lib/sdk_container_common.sh
```diff
@@ -0,0 +1,239 @@
...
+#
+function vernum_from_version() {
+    local version="$1"
+
+    # support vernums and versions ("alpha-"... is optional)
+    echo "${version}" | sed -n 's/^\([a-z]\+-\)\?\([0-9.]\+\).*/\2/p'
+}
+# --
+
+# Strip prefix from version string if present ("alpha-3233.0.0[-...]" => "3233.0.0[-...]")
```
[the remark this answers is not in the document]
t-lo: I'll add a comment. The (traditional) build scripts use `version`+`build` for non-official (non-release) builds, this is mimic'ed here.

sdk_lib/90_env_keep
```diff
@@ -0,0 +1,8 @@
+Defaults env_keep += "FLATCAR_BUILD_ID COREOS_OFFICIAL \
```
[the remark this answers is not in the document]
t-lo: No, I don't think it does (I believe I tested comments in this file and things broke).

ci-automation/sdk_bootstrap.sh
```diff
@@ -0,0 +1,90 @@
...
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# >>> This file is supposed to be SOURCED from the repository ROOT. <<<
+#
+# sdk_bootstrap() should be called w/ the positional INPUT parameters below.
+
+# Boostrap SDK build automation stub.
```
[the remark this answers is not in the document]
t-lo: Oh wow that typo is gorgeous. Nice catch!

t-lo:
@krnowak Thank you for the feedback! I've addressed the comments and committed your suggestions.

README.md
```diff
@@ -2,4 +2,131 @@
...
+In order to build OS images (via `./build_image` and `./image_to_vm`) the SDK tooling requires privileged access to `/dev`.
+This is necessary because the SDK currently employs loop devices to create and to partition OS images.
+
+To start a container in privileged mode with `/dev` available use:
+* `docker run -ti  --privileged -v /dev:/dev ghcr.io/flatcar-linux/flatcar-sdk-all:3033.0.0`
+
+## Integrated mode
+
+This is the preferred mode of working with the SDK.
+Interaction with the container happens via wrapper scripts from the Scripts repository.
```
krnowak, line 86:
```suggestion
Interaction with the container happens via wrapper scripts from the scripts repository.
```

krnowak (approved):
The PR gets unwieldy to review (it's just a lot of code to pore over). Found one nit. I see some issues are not yet resolved (podman?), but the PR looks good from my side.

README.md
t-lo: I think podman support should be a follow-up PR (you're rightfully mentioned this one is growing big).
One key benefit of this work is that it coexists with the traditional SDK - improvements / features can be added incrementally.
Thanks for reviewing!

t-lo:
Rebased on the latest main (scripts, coreos-overlay, portage-stable).

pothos:
> The PR gets unwieldy to review (it's just a lot of code to pore over). Found one nit. I see some issues are not yet resolved (podman?), but the PR looks good from my side.

We can merge it as is and add Podman support later (I can do it since I would use Podman anyway on Fedora)

jepio:
+1 for merging as well. I love that we can run it side by side as we transition and have something to fall back on.

t-lo:
🤩 Will merge & create separate, squashed PRs for the release branches flatcar-(2983,3033,3066).

PR #150, opened 2021-09-03
jenkins: add script to run kola arm64 tests under docker

jepio:
# jenkins: add script to run kola arm64 tests under docker

This PR introduces a script that runs qemu_uefi arm64 tests through a docker container instead of in a Flatcar SDK. The entrypoint uses the jenins paramter 'NATIVE_ARM64' to select whether to use the normal path, or the docker path.

### DIFF
--- a/jenkins/kola/Dockerfile.kola-test-runner
+++ b/jenkins/kola/Dockerfile.kola-test-runner
@@ -1,0 +1,4 @@
+FROM debian:11
+
+RUN apt-get update && \
+  apt-get install -y qemu-system-aarch64 qemu-efi-aarch64 lbzip2 sudo dnsmasq gnupg2 git curl iptables
--- a/jenkins/kola/qemu_common.sh
+++ b/jenkins/kola/qemu_common.sh
@@ -12,11 +12,35 @@ else
   echo "Unknown platform: \"${PLATFORM}\""
 fi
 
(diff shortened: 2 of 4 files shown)


jepio:
This is not pretty but it now works: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu_uefi/4567/console.
Triggering a full build.

pothos:
Was confused because `enter` is still used until I saw it's a no-op

pothos:
To get it running completely, you can try to _Rebuild_ http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu_uefi/4559/

pothos (commented):
Can we move the changes in the `qemu_common_arm64.sh` file into the existing file with if-else cases? E.g., whether `enter` is a no-op and whether the special fodler setup needs to be done.

jepio:
> Can we move the changes in the `qemu_common_arm64.sh` file into the existing file with if-else cases? E.g., whether `enter` is a no-op and whether the special fodler setup needs to be done.

OK, did that now. Works nicely: http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/qemu_uefi/4590/console. I think we could switch to it by default.

pothos (approved):

PR #153, opened 2021-09-28
jenkins: run simple network test on different hardware

pothos:
The Linux 5.10 stable kernel introduced a regression that we didn't
catch because we only run kola on one hardware type in Equinix Metal.
Validate that a simple network test works on various instance types of
the current hardware generation.

## How to use
Port to all active channel branches

## Testing done
[Ongoing](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3689/cldsv/)

### DIFF
--- a/jenkins/kola/packet.sh
+++ b/jenkins/kola/packet.sh
@@ -31,6 +31,36 @@ if [[ "${KOLA_TESTS}" == "" ]]; then
   KOLA_TESTS="*"
 fi
 
+# cl.basic includes cl.internet which is run on multiple instance types
+cl_basic_included="$(bin/kola list --platform=packet --filter "${KOLA_TESTS}" | grep cl.basic)"
+if [[ "${BOARD}" == "amd64-usr" ]] && [[ "${cl_basic_included}" != ""  ]]; then
+  for INSTANCE in c3.small.x86 c3.medium.x86 m3.large.x86 s3.xlarge.x86 n2.xlarge.x86; do
+    (
+    OUTPUT=$(timeout --signal=SIGQUIT "${timeout}" bin/kola run \
+    --basename="${NAME}" \
+    --board="${BOARD}" \
+    --channel="${GROUP}" \
+    --gce-json-key="${UPLOAD_CREDS}" \
+    --packet-api-key="${PACKET_API_KEY}" \
+    --packet-facility="${PACKET_REGION}" \
+    --packet-image-url="${IMAGE_URL}" \
+    --packet-installer-image-kernel-url="${KERNEL_URL}" \
+    --packet-installer-image-cpio-url="${CPIO_URL}" \
+    --packet-project="${PACKET_PROJECT}" \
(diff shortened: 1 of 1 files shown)


jenkins/kola/packet.sh
```diff
@@ -27,6 +27,29 @@ else
   CPIO_URL="http://${BASE_PATH}/flatcar_production_pxe_image.cpio.gz"
 fi
 
+if [[ "${BOARD}" == "amd64-usr" ]]; then
```
t-lo, line 30:
```suggestion
if [[ "${KOLA_TESTS}" == "" ]]; then
  KOLA_TESTS="*"
fi

# cl.basic includes cl.internet which is run on multiple instance types
cl_basic_included="$(bin/kola list --filter "${KOLA_TESTS}" | grep cl.basic)"
if [[ "${BOARD}" == "amd64-usr" ]] && [[ "${cl_basic_included}" != ""  ]]; then
```
Should we maybe skip this if there's a filter set for kola which does not include cl.basic?

jenkins/kola/packet.sh
```diff
@@ -27,6 +27,29 @@ else
...
+    --packet-project="${PACKET_PROJECT}" \
+    --packet-storage-url="${UPLOAD_ROOT}/mantle/packet" \
+    --packet-plan="${INSTANCE}" \
+    --parallel="${PARALLEL_TESTS}" \
+    --platform=packet \
+    --tapfile="${JOB_NAME##*/}_validate_${INSTANCE}.tap" \
+    --torcx-manifest=torcx_manifest.json \
+    cl.internet || true
+  done
+fi
```
t-lo, line 51:
```suggestion
    cl.internet || true &
  done
fi

wait
```
We can kick these off in parallel if we don't care about the result (`|| true`) anyway.

jenkins/kola/packet.sh
```diff
@@ -27,6 +27,29 @@ else
...
+
 if [[ "${KOLA_TESTS}" == "" ]]; then
   KOLA_TESTS="*"
 fi
```
t-lo, line 55:
```suggestion
```
Suggest removal if we ingest my first suggestion above (which moves this up above the cl.internet test).

t-lo (approved):
Thank you! Looks good in general, suggested a few improvements.

jenkins/kola/packet.sh
pothos: Yes, I just think that it makes the stdout logs worthless if they are mangled

jenkins/kola/packet.sh
pothos: Done, also added the platform option for the filter (default is qemu)

jenkins/kola/packet.sh
pothos: I pushed a change that starts all in parallel but collects the output to print it in one go. It can still become mangled but less.

dongsupark (approved):

PR #154, opened 2021-09-28
build_library: Ignore more categories in write_licenses

krnowak:
acct-user and acct-group categories aren't licensed too.

CI: http://localhost:9091/job/os/job/manifest/3693/cldsv/


krnowak:
CI passed, the warnings are not printed any more.

sayanchowdhury (approved):
LGTM

PR #156, opened 2021-09-30
jenkins/kola/packet.sh: fix check for skipping machine type tests

pothos:
The cl.basic and cl.internet tests are different tests which wasn't
    clear before. Also, the grep process returns an exit code of 1 if it
    didn't find a match, causing the job to cancel. The list of tests is
    space separated and should not be quoted but on the other hand, we
    do have to handle a literal *.
    Look for the right test and handle the grep exit code, and disable
    globs for the subshell for preserving a literal *.

## How to use

Pick to all channels

## Testing done

Local execution of the command pipe with cl.internet as argument and without, and checking that it has a 0 exit code.

### DIFF
--- a/jenkins/kola/packet.sh
+++ b/jenkins/kola/packet.sh
@@ -31,9 +31,9 @@ if [[ "${KOLA_TESTS}" == "" ]]; then
   KOLA_TESTS="*"
 fi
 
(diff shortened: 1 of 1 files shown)


sayanchowdhury (approved):

pothos:
Found one more problem, the globbing.

PR #157, opened 2021-10-01
build_packages: Add another cycle to break

krnowak:
With the update of the sys-apps/util-linux package, another cycle has
reared its ugly head. Looks like the package got an optional
dependency on sys-fs/cryptsetup, so break it.

For the util-linux update, see https://github.com/flatcar-linux/portage-stable/pull/222 and https://github.com/flatcar-linux/coreos-overlay/pull/1320.


dongsupark (approved):
Thanks for the fix.

Manually tested.
I can confirm that this PR fixes the cyclic dependency [issue](https://github.com/flatcar-linux/portage-stable/pull/222#issuecomment-932327990) of util-linux and cryptsetup.

PR #158, opened 2021-10-05
jenkins: add `--json-key` to `cork update` commands

tormath1:
it pulls https://github.com/flatcar-linux/mantle/pull/239 to be able to
use `--json-key` in order to access private GCS bucket

## Note for reviewers

* I have a doubt regarding the availability of `GOOGLE_APPLICATION_CREDENTIALS` from each script
* in `jenkins/vm.sh` we mix `GS_DEVEL_CREDS` and `GOOGLE_APPLICATION_CREDENTIALS` - I think they target the same file


jenkins/images.sh
```diff
@@ -32,6 +32,7 @@ fi
 bin/cork update \
     --create --downgrade-replace --verify --verify-signature --verbose \
     --sdk-url-path "${SDK_URL_PATH}" \
+    --json-key "${GOOGLE_APPLICATION_CREDENTIALS}" \
```
jepio, line 35: Below the following is used:
```
sudo ln -f "${GS_DEVEL_CREDS}" chroot/etc/portage/gangue.json
...
--json-key=/etc/portage/gangue.json $verify_key
```
So should it be `GS_DEVEL CREDS`?

tormath1: right, like with `jenkins/vm.sh` - I missed this one, thanks !

dongsupark (approved):
Looks ok.
A test CI run would be great.

tormath1:
@dongsupark I started a build: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/3742/cldsv/

tormath1:
@krnowak thanks for the explanation - I really had a doubt regarding this duality `GOOGLE_APPLICATION_CREDENTIALS` / `GS_DEVEL_CREDS`.

I think we're good using `GS_DEVEL_CREDS` we don't want to upload but to download a SDK (from `flatcar-jenkins`) and we can see that the `enter` function does use this creds too in the fetch command, so that "makes sense" but it's a bit confusing.

PR #159, opened 2021-10-05
setup_board, update_chroot, dev container: use new bincache mirror

t-lo:
This updates the default settings in build scripts to use https://mirror.release.flatcar-linux.net/ instead of the google storage bucket if no binhost or FLATCAR_DEV_BUILDS is specified.

Defaults are updated for
* update_chroot (runs at SDK initialisation time)
* setup_board (creates /boards/[ARCH]/) chroots
* the development container
* set_version

# Testing done
- [x] run `./update_chroot` with and without `FLATCAR_DEV_BUILDS` set, verified that `https://mirror.release.flatcar-linux.net/` is used by default, the URL in `FLATCAR_DEV_BUILDS` otherwise
- [x] run `./setup_board`  with and without `FLATCAR_DEV_BUILDS` set, same as above
- [x] run `./build_packages` and verify `https://mirror.release.flatcar-linux.net/` is used as bincache
- [ ] run `./build image prodtar container` and verify dev container uses `https://mirror.release.flatcar-linux.net/`
- [ ] run full CI build (incl. SDK)


set_version
```diff
@@ -41,10 +41,28 @@ Usage: $0 FLAGS...
...
+for arg in ${@} ; do
+    if $custom_binhost; then
+        echo "Using custom binhost '${arg}'"
+        DEFAULT_BASE_URL="${arg}"
+        DEV_BOARD_URL="${DEFAULT_BASE_URL}/developer"
+        DEFAULT_SDK_URL="${DEFAULT_BASE_URL}/sdk"
+        DEV_SDK_URL="${DEFAULT_BASE_URL}/developer/sdk"
+        break
+    fi
+    if [ "$arg" = "--binhost" ] ; then
```
krnowak, line 61: You will need to add this to `case` below, otherwise won't you get the typoed error `"Unkown argument --binhost"`?

setup_board
```diff
@@ -27,8 +27,10 @@ DEFINE_boolean getbinpkg "${FLAGS_TRUE}" \
   "Download binary packages from remote repository."
 DEFINE_string getbinpkgver "" \
   "Use binary packages from a specific version."
+
+get_board_and_variant $FLAGS_board $FLAGS_variant
```
krnowak, line 31: `FLAGS_board` and `FLAGS_variant` are not yet evaluated at this point, no? It happens below, after the line with `# Parse command line flags`, I think.

set_version
t-lo: Good catch! Will fix.

setup_board
t-lo: True. Currently it only outputs the defaults. Since `get_board_and_variant` is only called to populate `BOARD` for the succeeding help text I'll just change the help text and remove `get_board_and_variant`.

setup_board
```diff
@@ -149,7 +149,6 @@ if [[ "${FLAGS_usepkgonly}" -eq "${FLAGS_TRUE}" ]]; then
   done
 fi
 
-get_board_and_variant $FLAGS_board $FLAGS_variant
```
krnowak, line 152: But this should stay, I think. Otherwise BOARD, VARIANT and BOARD_VARIANT will not be initialized.

setup_board
krnowak: The BOARD variable is used below also for other things, so I don't think `get_board_and_variant` call should be completely removed. Just called after the flags are evaluated, like it used to be.

setup_board
t-lo: Oh no I removed it entirely instead of moving it back where it was :-/ Will fix.

krnowak (approved):

build_library/dev_container_util.sh
```diff
@@ -6,9 +6,9 @@ get_binhost_url() {
 	local image_group=$1
 	local image_path=$2
 	if [ "${image_group}" == "developer" ]; then
-		echo "https://storage.googleapis.com/flatcar-jenkins/${image_group}/boards/${BOARD}/${FLATCAR_VERSION}/${image_path}"
+		echo "${FLATCAR_DEV_BUILDS}/${image_group}/boards/${BOARD}/${FLATCAR_VERSION}/${image_path}"
 	else
-		echo "https://storage.googleapis.com/flatcar-jenkins/boards/${BOARD}/${FLATCAR_VERSION_ID}/${image_path}"
+		echo "${FLATCAR_DEV_BUILDS}/boards/${BOARD}/${FLATCAR_VERSION_ID}/${image_path}"
 	fi
```
t-lo, line 12: Thinking about it, this won't work in the CI. The CI will set `FLATCAR_DEV_BUILDS` to a custom value (the private google bucket) so the official mirror will not be used when the dev container is built via CI.
I'll introduce a command line option to `build_image` to set the dev container's binhost.

t-lo:
OK, I think we're good to go now.

PR #170, opened 2021-10-15
jenkins: use the SDK_URL_PATH path for DOWNLOAD_ROOT_SDK

pothos:
The SDK can either be a release SDK or a dev build SDK which are stored
in different paths. DOWNLOAD_ROOT_SDK should be based on the
SDK_URL_PATH value which indicates whether it's a release or dev build
path.

## How to use
Pick for all channels.


pothos:
Not the full nightly but I rebuilt one manifest job of it and toolchains runs the change correctly.

PR #173, opened 2021-10-21
do: Update the base URL to point to bucket.release.f-ln

sayanchowdhury:
This needs to be picked to the release branches `flatcar-3033`, `flatcar-2983`, `flatcar-2905`. 

Signed-off-by: Sayan Chowdhury <schowdhury@microsoft.com>


pothos:
For consistency it would also be good to have it in `flatcar-lts-2605`

PR #181, opened 2021-11-22
kola/packet: override EM region for ARM64 server

tormath1:
Equinix Metal ARM server are not yet hourly available in the default `sv15` region
so we override the `PACKET_REGION` to `Dallas` since it's available in this region.
We do not override `PACKET_REGION` for both board on top level because we need to keep proximity
for PXE booting.

To be merged with https://github.com/flatcar-linux/mantle/pull/257


pothos (approved):
Thanks

tormath1:
cherry-picked to:
* flatcar-3033
* flatcar-3066

PR #183, opened 2021-11-24
jenkins: override PARALLEL_TESTS for ARM servers in da11

dongsupark:
We override `PARALLEL_TESTS`, because kola run with `PARALLEL_TESTS` >= 4 causes the tests to provision >= 12 ARM servers at the same time.
As the da11 region does not have that many free ARM servers, the whole tests will fail.
With `PARALLEL_TESTS=2` the total number of servers stays < 10.

In addition, we override `timeout` to 10 hours, because it takes more than 8 hours to run all tests only with 2 tests in parallel.

### DIFF
--- a/jenkins/kola/packet.sh
+++ b/jenkins/kola/packet.sh
@@ -35,7 +35,17 @@ fi
 # so we override the `PACKET_REGION` to `Dallas` since it's available in this region.
 # We do not override `PACKET_REGION` for both board on top level because we need to keep proximity
 # for PXE booting.
-[[ "${BOARD}" == "arm64-usr" ]] && PACKET_REGION="da11"
+# We override `PARALLEL_TESTS`, because kola run with PARALLEL_TESTS >= 4 causes the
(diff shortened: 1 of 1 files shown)


jenkins/kola/packet.sh
```diff
@@ -35,7 +35,14 @@ fi
 # so we override the `PACKET_REGION` to `Dallas` since it's available in this region.
 # We do not override `PACKET_REGION` for both board on top level because we need to keep proximity
 # for PXE booting.
```
tormath1, line 37: Maybe we can also keep the previous comment ?

dongsupark: I did not delete previous comments..?

tormath1: Haha indeed - I don't know why I thought they were gone.

tormath1 (approved):

dongsupark:
cherry-picked to flatcar-3033, flatcar-3066.

PR #189, opened 2021-12-10
jenkins: make release.sh publish arm64 for Stable

dongsupark:
From flatcar-3033 on, arm64 images are available for Stable channel, we need to also add arm64 to Stable boards.


jepio (approved):

dongsupark:
cherry-picked to flatcar-3033, flatcar-3066.

PR #190, opened 2021-12-15
New release: stable-3033.2.0

t-lo:
SDK container bump to the new stable release. Versionfile, coreos-overlay, and portage-stable have been updated to reflect the correct OS and SDK versions, and to point to the correct submodule commits respectively. This change only affects the SDK container logic.

Please note that after merging, I'd like to **force-push the tag stable-3033.2.0** to point to this commit, so when we check out the "stable-3033.2.0" tag we get the correct versionfile and submodule commits.


t-lo:
I think we can easily automate this at some point; for now the SDK container is experimental and I don't mind doing this manually.

sayanchowdhury (approved):

PR #194, opened 2021-12-21
*: Replace kinvolk github org with flatcar-linux

krnowak:
Stop relying on github redirects, they are a mixed blessing and using
them broke emerge-gitclone inside dev-container in silent way. The
script could not find a desired revision of portage-stable or
coreos-overlay, because it tried to pull from kinvolk instead of
flatcar-linux github org. The redirects seem to hinder fetching a
specific commit, so the script pulled something else (HEAD or main?).

CI: http://localhost:9091/job/os/job/manifest/4445/cldsv/

Fixes https://github.com/flatcar-linux/Flatcar/issues/577

### DIFF
--- a/bootstrap_sdk
+++ b/bootstrap_sdk
@@ -151,7 +151,7 @@ stage_repo() {
         info "Using git ref '$gitref' for ebuild repo '$repo' ('$gitname') in stage 1."
         mkdir "$dest/$repo"
         (   cd "$dest/$repo" \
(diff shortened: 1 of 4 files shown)


krnowak:
CI passed.

krnowak:
Cherry-picked to flatcar 3033, 3066 and 3087. LTS didn't need the fix.

PR #196, opened 2022-01-05
run_sdk_container: use exact match for existing containers

t-lo:
`run_sdk_container` uses the sourcetree version to decide whether to re-use existing containers or create new ones. However, containers were not matched by exact name - instead, plain `--filter name="..."` was used, leading to prefix matching. This change updates `name="..."` to use regular expressions for exact matching.

This change should be cherry-picked into the release branches flatcar-3033, flatcar-3066, and into the upcoming Alpha release branch.

### DIFF
--- a/run_sdk_container
+++ b/run_sdk_container
@@ -77,7 +77,7 @@ if [ -z "$name" ] ; then
     name="flatcar-sdk-${arch}-${docker_sdk_vernum}_os-${docker_os_vernum}"
 fi
 
-stat="$(docker ps --all --no-trunc --filter name="$name" --format '{{.Status}}'\
+stat="$(docker ps --all --no-trunc --filter name="^/$name\$" --format '{{.Status}}'\
         | cut -f1 -d' ')"
 
 # pass SDK related environment variables and gcloud auth


run_sdk_container
```diff
@@ -77,7 +77,7 @@ if [ -z "$name" ] ; then
     name="flatcar-sdk-${arch}-${docker_sdk_vernum}_os-${docker_os_vernum}"
 fi
 
-stat="$(docker ps --all --no-trunc --filter name="$name" --format '{{.Status}}'\
+stat="$(docker ps --all --no-trunc --filter name="^/$name\$" --format '{{.Status}}'\
```
krnowak, line 80: Docker docs are just terrible. I found some [message on docker forum](https://forums.docker.com/t/how-to-filter-docker-ps-by-exact-name/2880/11) about this regexp stuff (the slash after `^` is what confused me).

krnowak (approved):

run_sdk_container
t-lo: It's likely the same message I found :rofl: 
What's even more confusing is that `--filter name="..."` does not do an exact match by default. I feel this violates [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment), but maybe that's just me.

PR #199, opened 2022-01-05
run/build_sdk_container: support Podman

pothos:
When the docker wrapper script for Podman is used, we need to
explicitly create a root user container with "sudo podman".
Podman also has its own bridge for root user containers which we need
to detect.
Add a "$docker" variable that uses sudo as needed, and also check which
bridge interface to use.

## How to use

Port to all active branches (except flatcar-lts-2605)

## Testing done

Used it for both regular container starts and the building of the SDK container (still running, though)

### DIFF
--- a/build_sdk_container_image
+++ b/build_sdk_container_image
@@ -111,16 +111,16 @@ create_versionfile "$version" "${os_version}"
 
 docker_vernum="$(vernum_to_docker_image_version "${version}")"
 import_tarball="flatcar-sdk-tarball:${docker_vernum}"
-image_present="$(docker image ls "$import_tarball" --format '{{.Repository}}:{{.Tag}}')"
+image_present="$($docker image ls "$import_tarball" --format '{{.Repository}}:{{.Tag}}')"
 
 if [ "${image_present}" = "${import_tarball}" ] ; then
     yell "Using existing SDK tarball image '${import_tarball}'"
 else
     yell "Importing SDK tarball"
     if [ -n "$cleanup" ] ; then
-        echo "docker image rm -f '${import_tarball}'" >> "$cleanup"
+        echo "$docker image rm -f '${import_tarball}'" >> "$cleanup"
     fi
-    docker import "${tarball}" "${import_tarball}"
+    $docker import "${tarball}" "${import_tarball}"
 fi
 
 # --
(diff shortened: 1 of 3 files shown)


build_sdk_container_image
```diff
@@ -178,19 +178,25 @@ else
         rm "${tarball_copied}"
     fi
 
-    docker container rm "${toolchains_container}"
+    $docker container rm "${toolchains_container}"
 
-    host_ip="$(ip addr show docker0 | grep -Po 'inet \K[\d.]+')"
+    docker_interface="docker0"
+    if docker --version | grep -q -i podman; then
```
tormath1, line 184: both scripts `run_sdk_container` and `build_sdk_container_image` source `sdk_lib/sdk_container_common.sh` so we could define a common helper in this file rather than doing `docker --version | grep ...` twice.

What do you think ? :)

sdk_lib/sdk_container_common.sh
```diff
@@ -11,6 +11,11 @@ sdk_container_common_versionfile="sdk_container/.repo/manifests/version.txt"
 sdk_container_common_registry="ghcr.io/flatcar-linux"
 sdk_container_common_env_file="sdk_container/.sdkenv"
 
+docker="docker"
+if docker --version | grep -q -i podman; then
```
tormath1, line 15: I think we could extend the test in case the user has not aliased `docker` with `podman` ?

build_sdk_container_image
pothos: ok, done

sdk_lib/sdk_container_common.sh
pothos: An alias from `.bashrc` doesn't matter for scripts

tormath1: Ok - but then, this command will just fail if there is no `docker` on the system.

On my machine:
```
./run_sdk_container -t
sdk_lib/sdk_container_common.sh: line 15: docker: command not found

###### Writing versionfile 'sdk_container/.repo/manifests/version.txt' to SDK '3087.0.0', OS '3066.0.0+56-ga4864fd9'. ######
./run_sdk_container: line 85: docker: command not found

###### Creating a new container 'flatcar-sdk-all-3087.0.0_os-alpha-3066.0.0-56-ga4864fd9' ######
./run_sdk_container: line 107: docker: command not found
```

pothos: Misunderstood it, will change to check for the podman binary and use it preferably

pothos: Done

tormath1 (approved):
`podman` users say "thank you"

pothos:
Pushed one more change to use the full container image reference for caddy

PR #200, opened 2022-01-05
sdk_lib/sdk_entry.sh: use a login shell to source /etc/profile

pothos:
For execution of the compiled binaries in /build/arm64-usr we rely on
    qemu-user binfmt emulation and have to tell it where the root is with
    QEMU_LD_PREFIX because build systems don't chroot into /build/arm64-usr
    themselves (which also works just by chance on amd64 because we have
    similar glibc versions and so on). The env var setup was done in
    /etc/profile.d/qemu-aarch64.sh but is now not read anymore since the
    container runs the shell not as login shell.
    
    Add the login options to the bash and su calls when starting the
    container.

## How to use

Check that `env` really prints out `QEMU_LD_PREFIX=/build/arm64-usr/`

Backport to all active branches

## Testing done

Manual edit of the file

### DIFF
--- a/sdk_lib/sdk_entry.sh
+++ b/sdk_lib/sdk_entry.sh
@@ -22,7 +22,7 @@ chown -R sdk:sdk /home/sdk
 #    our quotes for su -c "<cmd>" already.
 if [ $# -gt 0 ] ; then
     cmd="/home/sdk/.cmd"
-    echo -n "exec bash -i -c '" >"$cmd"
+    echo -n "exec bash -l -i -c '" >"$cmd"
     for arg in "$@"; do
         echo -n "\"$arg\" " >>"$cmd"
     done
@@ -33,5 +33,5 @@ if [ $# -gt 0 ] ; then
     rm -f "$cmd"
     exit $rc
 else
-    exec sudo su sdk
+    exec sudo su -l sdk
 fi


sdk_lib/Dockerfile.sdk-import
```diff
@@ -38,6 +38,7 @@ RUN rm /home/sdk/.bashrc
 RUN cp /etc/skel/.bashrc /home/sdk
 RUN echo "cd /home/sdk/trunk/src/scripts" >> /home/sdk/.bashrc
 RUN echo 'export PATH="$PATH:/usr/local/bin:/usr/local/sbin"' >> /home/sdk/.bashrc
+RUN echo "if [-f /etc/profile.d/qemu-aarch64.sh ]; then source /etc/profile.d/qemu-aarch64.sh; fi" >> /home/sdk/.bashrc
```
jepio, line 41: Is this the only file we anticipate being there? If not then a for loop over the files in `/etc/profile.d` makes sense.

pothos: We could source `/etc/profile` which already loops over the files in `/etc/profile.d` but maybe we get a circle…? No idea why `/etc/profile` is skipped now.

pothos: As found out it, we need `-l` for bash and su

t-lo (approved):
LGTM, thank you for the swift turn-around!

t-lo:
@pothos please don't forget to cherry-pick into flatcar-3033 and flatcar-3066 (we can skip flatcar-3087 I guess since it never got released and we're now planning to release an entirely new Alpha).

pothos:
Thanks, I did it but hadn't pushed it until now

PR #201, opened 2022-01-05
sdk_entry.sh: ensure currect binpkg host

t-lo:
This change ensures the binpkg host is updated if the board (OS) version differs from the SDK version.
This is to ensure /build/[arch] uses the correct binary package cache.

The change aims to address an [issue](https://github.com/flatcar-linux/flatcar-docs/pull/195#discussion_r778978471) uncovered by @pothos while reviewing documentation changes.

Also, a minor one-line change in `run_sdk_container` reduces the shutdown time of the SDK container, fixing a 10s lag when exiting the SDK.

### DIFF
--- a/sdk_lib/sdk_entry.sh
+++ b/sdk_lib/sdk_entry.sh
@@ -9,6 +9,31 @@ fi
 
 chown -R sdk:sdk /home/sdk
 
+# Check if the OS image version we're working on is newer than
+#  the SDK container version and if it is, update the boards
+#  chroot portage conf to point to the correct binhost.
+(
+    source /etc/lsb-release # SDK version in DISTRIB_RELEASE
+    source /mnt/host/source/.repo/manifests/version.txt # OS image version in FLATCAR_VERSION_ID
+
+    if [ "${FLATCAR_VERSION_ID}" != "${DISTRIB_RELEASE}" ] ; then
+        for target in amd64-usr arm64-usr; do
+            if [ -f "/build/$target/etc/target-version.txt" ] ; then
+                source "/build/$target/etc/target-version.txt"
+                if [ "${TARGET_FLATCAR_VERSION_ID}" = "${FLATCAR_VERSION_ID}" ] ; then
+                    continue # already updated
+                fi
+            fi
+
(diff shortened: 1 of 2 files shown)


sdk_lib/sdk_entry.sh
```diff
@@ -9,6 +9,31 @@ fi
...
 chown -R sdk:sdk /home/sdk
 
+# Check if the OS image version we're working on is newer than
+#  the SDK container version and if it is, update the boards
+#  chroot portage conf to point to the correct binhost.
+(
+    source /etc/lsb-release # SDK version in DISTRIB_RELEASE
+    source /mnt/host/source/.repo/manifests/version.txt # OS image version in FLATCAR_VERSION_ID
+
+    if [ "${FLATCAR_VERSION_ID}" != "${DISTRIB_RELEASE}" ] ; then
```
pothos, line 19: Maybe for extra safety this outer if block could be removed to ensure that /build/$target/etc/target-version.txt is always correct?

pothos: I can imaging someone starting the container and changing the version files afterwards and working on the right branches, entering the container again forcefully with `- NAME` despite the version mismatch, assuming all is still safe, but then ending up using wrong binary packages.

t-lo: I would argue that in this case, the changes are deliberate - i.e. the user *wants* to use a specific "underlay" (versionfile + branches) instead of the default one - particularly when forcing a custom name. It's pretty hard to end up in that situation without intention.

While initially developing the SDK container's version matching, I was far more rigid - and found that I had painted myself into a corner more often than not when I needed to (consciously) diverge from the default path.

I think when folks decide to mess with the versionfile while the container is running (instead of using `run_sdk_container -v` or `-V`) then they probably know what they are doing.

pothos (approved):

t-lo:
Will merge and cherry-pick to flatcar-3033, flatcar-3066.

t-lo:
Cherry-picked to flatcar-3033, flatcar-3066.

PR #202, opened 2022-01-06
checkout: helper script for checking out branches and tags

t-lo:
Helper script to check out branches and tags and keeping submodules in sync.

Usage:
`./checkout <tag-or-branch>`

The tool will guess whether a tag or branch was specified, with tags taking preference over branches of the same name.
For branches, the tool will check out the branch in both submodules too, and by default pull / rebase changes from origin.
Tags will be checked put in `scripts` only, and submodules will be updated (i.e. a tag is expected to have the correct submodule pinnings).

This should be cherry-picked to flatcar-3033 and flatcar-3066, too.

### DIFF
--- a/checkout
+++ b/checkout
@@ -1,0 +1,171 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -eu
+
+force_tag="false"
+force_branch="false"
+
+update_strategy="fast-forward"
+
+name=""
+
+usage() {
+    echo "  Usage:"
+    echo "  $0 [-t|-b] [-n] [-u <strategy>] <tag-or-branch>"
+    echo "       Check our a tag or branch and synchronise git submodules"
+    echo "          coreos-overlay and portage-stable."
+    echo "       By default, $0 tries to auto-detect whether to handle a tag or a branch,"
+    echo "         with tags being prioritised over branches of the same name."
+    echo
+    echo "       <tag-or-branch> Name of the tag or branch to check out."
+    echo "                       BRANCH: Check out the branch in scripts, coreos-overlay"
+    echo "                                and portage-stable, and fast-forward to the"
+    echo "                                latest changes by default (but see '-u' below)."
+    echo "                       TAG:    Check out the tag in scripts, and update submodules."
+    echo
+    echo "       -t            force TAG mode. Branches of the given name are ignored."
+    echo "                      Mutually exclusive with '-b'."
+    echo "       -b            force BRANCH mode. Tags of the given name are ignored."
+    echo "                      Mutually exclusive with '-t'."
+    echo
+    echo "       -u <strategy> Only applies to BRANCH checkouts."
+    echo "                      When checking out submodule branches, use <strategy>."
+    echo "                      instead of fast-forward. Strategy is one of:"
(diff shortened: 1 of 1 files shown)


checkout
```diff
@@ -0,0 +1,155 @@
...
+# --
+
+#
+# TAG
+#
+if ! $force_branch; then 
+    for dir in . \
+        sdk_container/src/third_party/coreos-overlay \
+        sdk_container/src/third_party/portage-stable ; do
+        git -C "$dir" fetch --tags
```
pothos, line 111: This fails as soon as a retag was done by someone, I think it's better to force the overwriting of local tags instead of bailing out

checkout
```diff
@@ -0,0 +1,155 @@
...
+echo "Checking out BRANCH '$branch'"
+echo "----------------------------------"
+for dir in . \
+    sdk_container/src/third_party/coreos-overlay \
+    sdk_container/src/third_party/portage-stable ; do
+
+    git -C "$dir" checkout "$branch"
+    if $update_branch; then
+        echo "Pulling / rebasing '$dir'"
+        git -C "$dir" pull --rebase
```
pothos, line 152: Why would you rebase? I think `--ff-only` is the better behavior in general.

pothos: In our case I would even say that `reset --hard` is the an appropriate mode for the manged repos, e.g., when I want to build a PR branch that gets force pushed often. Would be nice to offer this through a flag.

t-lo: I'm trying to be least intrusive to local work while offering the best possible (automated) alignment with upstream.
With `pull --rebase`, the branches get updated but local changes are preserved. This way, contributors can work on multiple changes to multiple branches locally, and can rebase their trees on upstream changes when necessary.

checkout
t-lo: That's a good catch, will update.

checkout
pothos: The rebase can lead to surprising results (e.g., a feature branch was force-pushed to drop a commit this would be sneaked in again, or?), I think ff-only is safer and when it errors the user should decide what to do when the local state differs - and probably choose between rebase or reset. Automatic rebase is normally not needed as long as the user wouldn't commit directly on a `flatcar-xy` branch but a feature branch.

pothos: Maybe we can default to `pull --ff-only` and offer `pull --rebase` behind a `-r` and `reset --hard` behind a `-f` flag or so.

t-lo: Makes sense!

t-lo:
Updated based on PR feedback. I ended up not implementing `reset` for 2 reasons:
- I would need to guess the correct remote to reset to in case there are multiple remotes
- It's very easy to lose local changes. I'd rather users reset manually.

checkout
```diff
@@ -0,0 +1,171 @@
...
+
+echo
+echo "Checking out BRANCH '$branch'"
+echo "----------------------------------"
+for dir in . \
+    sdk_container/src/third_party/coreos-overlay \
+    sdk_container/src/third_party/portage-stable ; do
+
+    git -C "$dir" checkout "$branch"
+    if [-n "$update_strategy" ] ; then
```
pothos, line 166:
```suggestion
    if [ -n "$update_strategy" ] ; then
```
(`[-n: command not found`)

pothos (approved):
Thanks

t-lo:
Merged + cherry-picked to flatcar-3033 and flatcar-3066.

PR #203, opened 2022-01-06
sdk_lib/sdk_entry: handle permission error for target version file

pothos:
The creation of the target version file failed:
/home/sdk/sdk_entry.sh: line 32: /build/amd64-usr/etc/target-version.txt: Permission denied

Use root permissions to create the file.

## How to use

Port to 3033 and 3066 branches

## Testing done

None

### DIFF
--- a/sdk_lib/sdk_entry.sh
+++ b/sdk_lib/sdk_entry.sh
@@ -29,7 +29,7 @@ chown -R sdk:sdk /home/sdk
             echo "Updating board support in '/build/${target}' to use package cache for version '${FLATCAR_VERSION_ID}'"
             echo "---"
             sudo su sdk -l -c "/home/sdk/trunk/src/scripts/setup_board --board='$target' --regen_configs_only"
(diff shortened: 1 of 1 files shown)


sdk_lib/sdk_entry.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -eu
```
t-lo, line 3: I don't think this is a good idea - if there is any issue with the entry script then this would prevent you from entering the container ever again, potentially losing a lot of (unrelated) work.

A bug in this script could be fixed from within the container; erroring out before establishing a shell would prevent you from entering the container though.

t-lo (changes requested):
Good catch! Doesn't seem to happen with docker (the 'echo' line runs as root), but the additional `sudo` doesn't hurt docker, either.

I think we shouldn't error out in the entry script though. I had something similar in that script at during development but removed it since it prevented me from fixing trivial errors that required me to enter the container.

sdk_lib/sdk_entry.sh
pothos: Ok, it means that we miss these error in CI, though

t-lo (approved):
LGTM, great work!

PR #204, opened 2022-01-06
Dockerfile.sdk-import: run update_chroot before setup_board

t-lo:
In PR https://github.com/flatcar-linux/scripts/pull/201 we added a call to setup_board to the SDK entry point `sdk_entry.sh`. Since we make use of `sdk_entry.sh` during the SDK container build, this inadvertently led to setup_board being called before update_chroot ran.

This fix does not call sdk_entry.sh for update_chroot, preventing the issue.

The fix should be cherry-picked to flatcar-3033 and flatcar-3066, too.

### DIFF
--- a/sdk_lib/Dockerfile.sdk-import
+++ b/sdk_lib/Dockerfile.sdk-import
@@ -6,6 +6,10 @@ RUN touch /etc/debian_chroot
 RUN chmod 644 /etc/passwd
 RUN chmod 644 /etc/group
 
+# User "root" is not in /etc/passwd / group in the SDK tarball
+RUN echo 'root:x:0:0:root:/root:/bin/bash' >>/etc/passwd
+RUN echo 'root:x:0:' >>/etc/group
+
 RUN if ! grep -q portage /etc/group ; then \
         echo "portage::250:portage" >>/etc/group; \
     fi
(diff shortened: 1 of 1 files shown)


pothos:
It takes a while to test this, not only due to the normal work done, also due to the many Dockerfile steps which at this image size cause a lot of intermediate images being stored to disk…

sdk_lib/Dockerfile.sdk-import
```diff
@@ -43,11 +43,11 @@ RUN echo 'export PATH="$PATH:/usr/local/bin:/usr/local/sbin"' >> /home/sdk/.bash
...
-COPY --chown=sdk:sdk sdk_lib/sdk_entry.sh /home/sdk
-RUN chmod 755 /home/sdk/sdk_entry.sh
-
 # This should be a NOP; if you see packages being rebuilt
 #  it's likely that submodules and SDK tarball are out of sync
-RUN /home/sdk/sdk_entry.sh ./update_chroot --toolchain_boards="amd64-usr arm64-usr"
+RUN sudo su sdk -l -c "FEATURES='-ipc-sandbox -network-sandbox -pid-sandbox' /home/sdk/trunk/src/scripts/update_chroot --toolchain_boards='amd64-usr arm64-usr'"
+
+COPY --chown=sdk:sdk sdk_lib/sdk_entry.sh /home/sdk
+RUN chmod 755 /home/sdk/sdk_entry.sh
```
pothos, line 51: This script has some commands that assume being root, I think we should either change `USER sdk:sdk` to `USER root:root` before this RUN step or change the script to use `sudo`

t-lo: Turns out user `root` is not in `/ect/passwd` or in `/etc/group` so `USER root:root` in `Dockerfile` fails. I'll update the `Dockerfile` to add `root`.

t-lo: (We still can't use `sdk_entry.sh` to call `update_chroot` because `update_chroot` needs to run before `setup_board` is called; however, `sdk_entry.sh` runs `setup_board` implicitly. The above is as good as it gets)

t-lo:
Closed in favour of https://github.com/flatcar-linux/scripts/pull/205.

PR #205, opened 2022-01-06
fix sdk_entry.sh setup_board fallout

t-lo:
In PR #201 we added a call to setup_board to the SDK entry point sdk_entry.sh. Since we make use of sdk_entry.sh during the SDK container build, this inadvertently led to setup_board being called before update_chroot ran.

This fix adds a check to sdk_entry.sh and only calls setup_board when `/build/...` exists. Additionally, it updates `sdk_lib/Dockerfile.sdk-import` to switch to the root user by default since `sdk_entry.sh` expects to be run as root.

Supersedes https://github.com/flatcar-linux/scripts/pull/204.

The fix should be cherry-picked to flatcar-3033 and flatcar-3066, too.

### DIFF
--- a/sdk_lib/Dockerfile.sdk-import
+++ b/sdk_lib/Dockerfile.sdk-import
@@ -6,6 +6,10 @@ RUN touch /etc/debian_chroot
 RUN chmod 644 /etc/passwd
 RUN chmod 644 /etc/group
 
+# User "root" is not in /etc/passwd / group in the SDK tarball
+RUN echo 'root:x:0:0:root:/root:/bin/bash' >>/etc/passwd
+RUN echo 'root:x:0:' >>/etc/group
+
 RUN if ! grep -q portage /etc/group ; then \
         echo "portage::250:portage" >>/etc/group; \
     fi
@@ -46,6 +50,7 @@ RUN echo "if [ -f /mnt/host/source/.sdkenv ]; then source /mnt/host/source/.sdke
 COPY --chown=sdk:sdk sdk_lib/sdk_entry.sh /home/sdk
 RUN chmod 755 /home/sdk/sdk_entry.sh
 
+USER root:root
 # This should be a NOP; if you see packages being rebuilt
 #  it's likely that submodules and SDK tarball are out of sync
 RUN /home/sdk/sdk_entry.sh ./update_chroot --toolchain_boards="amd64-usr arm64-usr"
--- a/sdk_lib/sdk_entry.sh
+++ b/sdk_lib/sdk_entry.sh
@@ -18,6 +18,9 @@ chown -R sdk:sdk /home/sdk
 
     if [ "${FLATCAR_VERSION_ID}" != "${DISTRIB_RELEASE}" ] ; then
         for target in amd64-usr arm64-usr; do
+            if [ ! -d "/build/$target" ] ; then
+                continue
+            fi
             if [ -f "/build/$target/etc/target-version.txt" ] ; then
                 source "/build/$target/etc/target-version.txt"
                 if [ "${TARGET_FLATCAR_VERSION_ID}" = "${FLATCAR_VERSION_ID}" ] ; then


pothos:
I still see this here, is this expected?
```
groupmod: Permission denied.
groupmod: cannot lock /etc/group; try again later.
```

pothos:
One more thing: In `sdk_lib/Dockerfile.sdk-build` `/home/sdk/sdk_entry.sh ./setup_board --board="amd64-usr" --binhost="${BINHOST}/amd64-usr"` gets skipped:
```
STEP 7/10: RUN /home/sdk/sdk_entry.sh ./setup_board --board="amd64-usr" --binhost="${BINHOST}/amd64-usr"
bash: cannot set terminal process group (1): Inappropriate ioctl for device
bash: no job control in this shell
WARNING setup_board: Board output directory '/build/amd64-usr' already exists.
WARNING setup_board: Not setting up board root. 
WARNING setup_board: Use --force to clobber the board root and start again.
--> bab3a4f31f2
STEP 8/10: RUN /home/sdk/sdk_entry.sh ./setup_board --board="amd64-usr" --regen_configs
```
Can this be one single call which should work?

pothos:
When using the built SDK container I directly get this error message, only for the amd64 board, the arm64 board gets configured without the error message:
```
Updating board support in '/build/amd64-usr' to use package cache for version '3066.0.0'
---
INFO    setup_board: Configuring portage in /build/amd64-usr
rm: cannot remove '/build/amd64-usr/etc/portage/make.profile': No such file or directory
!!! Error: Couldn't remove current /build/amd64-usr/etc/portage/make.profile symlink
exiting
ERROR   setup_board: script called: setup_board '--board=amd64-usr' '--regen_configs_only'
ERROR   setup_board: Backtrace:  (most recent call is last)
ERROR   setup_board:   file setup_board, line 237, called: die_err_trap 'sudo ROOT="${BOARD_ROOT}" eselect profile set --force "${PORTAGE_PROFILE}"' '250'
ERROR   setup_board: 
ERROR   setup_board: Command failed:
ERROR   setup_board:   Command 'sudo ROOT="${BOARD_ROOT}" eselect profile set --force "${PORTAGE_PROFILE}"' exited with nonzero code: 250
```

t-lo:
> I still see this here, is this expected?
> 
> ```
> ###### Building toolchains in a temporary container. ######
> […]
> groupmod: Permission denied.
> groupmod: cannot lock /etc/group; try again later.
> ```

It shouldn't since we explicitly change back to the `root` user now in [`Dockerfile.sdk-import`](https://github.com/flatcar-linux/scripts/blob/t-lo/fix-sdk_entry-setup_board-fallout/sdk_lib/Dockerfile.sdk-import#L53).
I don't see this in my output:
```
...
###### Building toolchains in a temporary container. ######
###### Writing versionfile 'sdk_container/.repo/manifests/version.txt' to SDK '3088.0.0', OS '3046.0.0+60-g0498fa84'. ######
###### Creating a new container 'flatcar-sdk-toolchains-build-3088.0.0' ######
c24e5cba7ea1791f397ac80dc434bdb6080aaa23de5e8befc57eb7badfffb72e
###### Starting stopped container 'flatcar-sdk-toolchains-build-3088.0.0' ######
flatcar-sdk-toolchains-build-3088.0.0
usermod: no changes
bash: cannot set terminal process group (15): Inappropriate ioctl for device
bash: no job control in this shell
INFO    build_toolchains: Building stages: stage4   
INFO    build_toolchains: Creating output directories...
...
```

However, `./build_sdk_container_image` tries to be *smart* and will re-use intermediate build artifacts - for instance, if there's already a container image `flatcar-sdk-import:<version-you-are-building>` present (from a previous, aborted run) then the script will [skip the step where it uses `Dockerfile.sdk-import`](https://github.com/flatcar-linux/scripts/blob/t-lo/fix-sdk_entry-setup_board-fallout/build_sdk_container_image#L133). Make sure to `docker image rm flatcar-sdk-import:<version-you-are-building>` to be safe. These intermediate images ore only removed after a build succeeded.


> One more thing: In sdk_lib/Dockerfile.sdk-build /home/sdk/sdk_entry.sh ./setup_board --board="amd64-usr" --binhost="${BINHOST}/amd64-usr" gets skipped:
[...]
> When using the built SDK container I directly get this error message, only for the amd64 board, the arm64 board gets configured without the error message:


Could this result from the same root cause as above, i.e. `Dockerfile.sdk-import` never ran and an old import image was used instead? `Dockerfile.sdk-import` copies `sdk_entry.sh` into the image, so you would end up with an outdated `sdk_entry.sh` which does not check if `/build/...` exists before calling `setup_board`.
I don't see this issue on my end, but then I remove all intermediate images when I abort a build.

t-lo:
I've updated the PR to move the check for `-d /build/$target` [into the loop](https://github.com/flatcar-linux/scripts/pull/205/files#diff-434a3abf1a3abb4f28eadf17d5e87b65458fbd77d8aa93d0aedc4b748f5645d6R21-R23), so we don't try to `setup_board` for all architectures. E.g. an arm64 SDK container should not try to `setup_board` for amd64 (this even works, astonishingly, but leaves a stub `/build/amd64-usr` in an arm64 container, which is ugly.).

pothos:
Thanks, I will try again after removing the local images

pothos:
Now it works except for this error when entering the built SDK:
```
Updating board support in '/build/amd64-usr' to use package cache for version '3046.0.0'
---
INFO    setup_board: Configuring portage in /build/amd64-usr
rm: cannot remove '/build/amd64-usr/etc/portage/make.profile': No such file or directory
!!! Error: Couldn't remove current /build/amd64-usr/etc/portage/make.profile symlink
exiting
ERROR   setup_board: script called: setup_board '--board=amd64-usr' '--regen_configs_only'
ERROR   setup_board: Backtrace:  (most recent call is last)
ERROR   setup_board:   file setup_board, line 237, called: die_err_trap 'sudo ROOT="${BOARD_ROOT}" eselect profile set --force "${PORTAGE_PROFILE}"' '250'
ERROR   setup_board: 
ERROR   setup_board: Command failed:
ERROR   setup_board:   Command 'sudo ROOT="${BOARD_ROOT}" eselect profile set --force "${PORTAGE_PROFILE}"' exited with nonzero code: 250

Updating board support in '/build/arm64-usr' to use package cache for version '3046.0.0'
```
It seems to be harmless, though and when starting another one with a different OS version it works.

pothos (approved):

t-lo:
Merged and cherry-picked to flatcar-3033 and flatcar-3066.

PR #206, opened 2022-01-06
build_sdk_container_image: force removal of running container

pothos:
At least with Podman it's not possible to call "container rm" on a
running container without the force flag.
Add the force flag which is also used elsewhere already.

## How to use

port to flatcar-3033 and 66

## Testing done

Manual removal worked with the flag


t-lo (approved):
Good catch - the cleanup script actually uses `-f` already (line 171).

PR #207, opened 2022-01-07
sdk_container: better docker / podman detection

t-lo:
In [bce3bd9031fa78c79173f29bcac5bc49c81a03b4](https://github.com/flatcar-linux/scripts/pull/199), we added support for podman for building and running the SDK container. The presence of podman is auto-detected in sdk_container_common.sh. However, podman is preverred over docker, requiring users to use *sudo* (which podman requires and
docker does not).

This change uses docker when present, podman otherwise. It also improves podman detection - 'podman' uses argv[0] in its version string, so if 'docker' is a symlink to 'podman', 'podman --version' output uses 'docker'. This broke the SDK container on hosts which have a 'docker' symlink to 'podman' since 'podman' is then run w/o 'sudo'.

This should be cherry-picked to flatcar-3033 and flatcar-3066, too.

## Testing done
Host has docker and podman installed.
```shell
$ ./run_sdk_container -t
...
# uses docker (ps ax)
...
$ sudo mv /usr/bin/docker /usr/bin/docker-real
$ ./run_sdk_container -t
...
# uses "sudo podman" (ps ax)
...
$ sudo ln -s /usr/bin/podman /usr/bin/docker
$ ./run_sdk_container -t
...
# uses "sudo podman" (ps ax)
...
$ alias docker=podman
$ ./run_sdk_container -t
...
# uses "sudo podman" (ps ax)
...
```

### DIFF
--- a/sdk_lib/sdk_container_common.sh
+++ b/sdk_lib/sdk_container_common.sh
@@ -11,10 +11,27 @@ sdk_container_common_versionfile="sdk_container/.repo/manifests/version.txt"
 sdk_container_common_registry="ghcr.io/flatcar-linux"
 sdk_container_common_env_file="sdk_container/.sdkenv"
 
+# Check for podman and docker; use docker if present, podman alternatively.
+# Podman needs 'sudo' since we need privileged containers for the SDK.
+
 is_podman=false
 if command -v podman >/dev/null; then
-  is_podman=true
+    # podman is present
+    if command -v docker >/dev/null ; then
+        # docker is present, too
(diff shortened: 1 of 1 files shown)


pothos (approved):
Thanks, had the `--version` check but removed it, sorry. On Fedora there is a podman-docker package that has a wrapper script, not a symlink, thus missed this case, too.

t-lo:
Merged and cherry-picked for flatcar-3033 and flatcar-3066.

PR #210, opened 2022-01-10
ci-automation/garbage_collect: fix binpkg url

t-lo:
`garbage_collect.sh` [was using 'docker_vernum'](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/garbage_collect.sh#L92) where it should have been using 'vernum' (as [`push_pkgs.sh`](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/push_pkgs.sh#L64) does).

Also, make sure release directories are removed, not just packages.

This fix should be cherry-picked to 3066 and 3033.


krnowak (approved):

t-lo:
Merged and cherry-picked to 3066 and 3033.

PR #212, opened 2022-01-10
set_lsb_release: only set update-engine GROUP in /usr, not /etc

pothos:
The default image group is already encoded in
/usr/share/flatcar/update.conf but it was written to
/etc/flatcar/update.conf as well. This can cause problems when the user
switches channels by forcing an update to a specific release from the
different channel (e.g., through the flatcar-update tool) as it leaves
the file under /etc/flatcar/update.conf out of sync with the new
channel version in /usr/share/flatcar/update.conf.

Since we don't really need to write a specific channel to /etc on new
images as we can rely on the value from /usr, we now leave any possible
overwriting of the value in /etc entirely to the user.

## Testing done

Scheduled test build and checked that the resulting image has the file missing as expected

### DIFF
--- a/changelog/changes/2022-01-11-etc-flatcar-update-conf.md
+++ b/changelog/changes/2022-01-11-etc-flatcar-update-conf.md
@@ -1,0 +1,1 @@
+- Removed the pre-shipped `/etc/flatcar/update.conf` file, leaving it totally to the user to define the contents as it was unnecessarily overwriting the `/use/share/flatcar/update.conf` ([PR#212](https://github.com/flatcar-linux/scripts/pull/212))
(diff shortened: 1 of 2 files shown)


changelog/changes/2022-01-11-etc-flatcar-update-conf.md
```diff
@@ -0,0 +1 @@
+- Removed the pre-shipped `/etc/flatcar/update.conf` file, leaving it totally to the user to define the contents as it was overwriting the `/use/share/flatcar/update.conf` without need ([PR#212](https://github.com/flatcar-linux/scripts/pull/212))
```
krnowak, line 1:
```suggestion
- Removed the pre-shipped `/etc/flatcar/update.conf` file, leaving it totally to the user to define the contents as it was unnecessarily overwriting the `/use/share/flatcar/update.conf` ([PR#212](https://github.com/flatcar-linux/scripts/pull/212))
```

krnowak (approved):
Looks good, had one suggestion.

changelog/changes/2022-01-11-etc-flatcar-update-conf.md
pothos: Thanks!

PR #213, opened 2022-01-11
settings / ci-automation: remove "binpkg" prefix

t-lo:
The original intention of the "binpkg" prefix in the CI binary package cache URL was to separate packages from other build artifacts like containers, images, and SDK tarballs. Motivation was to separate developer content (binary packages) from CI automation artifacts (everything else); since binary packages are not used by the CI.

This broke assumptions in scripts which use the binary host URL for other things than packages - e.g. SDK tarballs or images. These scripts would get a bincache URL with "binpkg/" prepended, while CI automation would *not* use that prefix.

This change removes the use of "binpkg/" altogether since it would not work as intended without more significant changes to build scripts.

This PR should be cherry-picked to all active release branches (likely flatcar-3033, flatcar-3066, and flatcar-3115).

### DIFF
--- a/ci-automation/garbage_collect.sh
+++ b/ci-automation/garbage_collect.sh
@@ -89,7 +89,7 @@ function garbage_collect() {
         else
             echo "## ${version} is an OS image version. ##"
             rmpat="${BUILDCACHE_PATH_PREFIX}/containers/${os_docker_vernum}/flatcar-packages-*"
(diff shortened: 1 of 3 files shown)


krnowak (approved):

t-lo:
Merged & cherry-picked for flatcar-3033, flatcar-3066, and flatcar-3115.

PR #222, opened 2022-02-06
kola/em: increase timeout

tormath1:
number of test increased. While we don't have yet a way to reduce
testing time, let's increase the timeout.


jepio (approved):
LGTM. Would be cool to merge before tonights EM test run.

PR #224, opened 2022-02-09
jenkins/kola: use httpS URL for PXE boot

tormath1:
Follow-up of:
* https://github.com/flatcar-linux/mantle/pull/288
* https://github.com/flatcar-linux/Flatcar/issues/527

### DIFF
--- a/jenkins/kola/packet.sh
+++ b/jenkins/kola/packet.sh
@@ -9,10 +9,6 @@ timeout=8h
 set -o pipefail
 
 # Construct the URLs of the image to be used during tests.
-# KERNEL/CPIO_URL will be used by iPXE and so it will use http instead of https to
-# make the boot process faster (except for signed URLs).
-# IMAGE_URL is downloaded through Flatcar and can do SSL just fine, so that one
-# can use https:// without a significant delay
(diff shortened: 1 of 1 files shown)


sayanchowdhury (approved):

pothos:
Please do a CI run first, I don't know if the proxy uses the shorter chain

pothos:
Also worth looking at the provisioning times - at least in the past I remember that ipxe seemed to download slower with https, hopefully that's not the case.

jepio:
If the proxy fetches its own cert, then we'll need to look at making it use the shorter chain as well.

jepio:
I switched the bucket cache to also use the shorter chain.

pothos (approved):

tormath1:
I'll trigger a CI run once https://github.com/flatcar-linux/mantle/pull/288 will be merged - just to be sure and avoid any surprise for the next round of releases.

pothos:
The issue is that `packet` does not build the pxe images, so `packet pxe` is what should be in `FORMAT_LIST_AMD64`

tormath1:
@pothos thanks for the information - Equinix Metal are running as expected now with HTTPS URLs.

PR #226, opened 2022-02-15
jenkins/images: print out image diff report

pothos:
Often a change results in unexpected effects on the image, e.g., when
a wrong package version gets chosen or the package installs files under
/etc, or binaries of library dependencies get pulled in. Besides
inspecting the image manually, the package-diff tool also gives
valuable insights.
Run the package-diff tool in a comparison to the last release and print
the image URL alongside for convenience.

## How to use

Port to all channels

## Testing done

[Looked at Jenkins job output](http://jenkins.infra.kinvolk.io:8080/job/os/job/board/job/image-matrix/6931/console)

### DIFF
--- a/jenkins/images.sh
+++ b/jenkins/images.sh
@@ -127,3 +127,44 @@ script build_image \
     --torcx_root=/mnt/host/source/torcx/ \
     --upload_root="${UPLOAD_ROOT}" \
     --upload prodtar container
+
+set +x
+# Don't fail the whole job
+set +e
+echo "==================================================================="
+echo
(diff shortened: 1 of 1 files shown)


jepio (approved):
Looks great in the console. The one thing that doesnt seem too helpful is the output of ` CALCSIZE=1 ./package-diff `. I'd much rather see something equivalent to `df -h /boot/ /usr/ /`. But wouldn't block this PR on that.

pothos:
> I'd much rather see something equivalent to `df -h /boot/ /usr/ /`. But wouldn't block this PR on that.

Good idea to have it more detailed

PR #227, opened 2022-02-16
build_library: move package sysusers to /usr database

pothos:
## Testing done

TODO


build_library/build_image_util.sh
```diff
@@ -607,6 +607,25 @@ finish_image() {
        "${root_fs_dir}/boot/flatcar/vmlinuz-a"
   sudo rm "${root_fs_dir}/usr/boot/vmlinuz"*
 
+  # Forbid dynamic user ID allocation because we want stable IDs
+  local found=""
+  # We want to forbid "-", "X:-", "-:X", "-:-"
+  found=$({ grep '^[^#]' "${root_fs_dir}"/usr/lib/sysusers.d/*.conf || true ; } | awk '{print $3}' | { grep -x -- "-"$'\n'".*:-"$'\n'"-:.*" || true ; })
+  if [ "${found}" != "" ]; then
+    die "Found dynamic ID allocation instead of hardcoded ID in /usr/lib/sysusers.d/*.conf (third column must not be '-')"
```
krnowak, line 615:
```suggestion
    die "Found dynamic ID allocation instead of hardcoded ID in /usr/lib/sysusers.d/*.conf (third column must not use '-')"
```

build_library/build_image_util.sh
```diff
@@ -607,6 +607,25 @@ finish_image() {
        "${root_fs_dir}/boot/flatcar/vmlinuz-a"
   sudo rm "${root_fs_dir}/usr/boot/vmlinuz"*
 
+  # Forbid dynamic user ID allocation because we want stable IDs
+  local found=""
+  # We want to forbid "-", "X:-", "-:X", "-:-"
+  found=$({ grep '^[^#]' "${root_fs_dir}"/usr/lib/sysusers.d/*.conf || true ; } | awk '{print $3}' | { grep -x -- "-"$'\n'".*:-"$'\n'"-:.*" || true ; })
```
krnowak, line 613: I'd limit that to `u` and `g` entries (for users and groups) so:
```suggestion
  found=$({ grep '^[ug]' "${root_fs_dir}"/usr/lib/sysusers.d/*.conf || true ; } | awk '{print $3}' | { grep -x -- "-"$'\n'".*:-"$'\n'"-:.*" || true ; })
```

krnowak: About `"-"$'\n'".*:-"$'\n'"-:.*"`. I suppose that newlines are separating three patterns: `-`, `.*:-` and `-:.*`. So the second pattern will match the unlikely line like `/file/owned/by/user/:-`.

I'd change the pattern to `'-\|-:-\|[0-9]\+:-\|-:[0-9]\+'` (basically use `\|` instead of `$'\n'`, and `[0-9]\+` instead of `.*`, and add `-:-` as another subpattern).

build_library/build_image_util.sh
```diff
@@ -607,6 +607,25 @@ finish_image() {
...
+    die "Found dynamic ID allocation instead of hardcoded ID in /usr/lib/sysusers.d/*.conf (third column must not be '-')"
+  fi
+  # Run systemd-sysusers once to create users in /etc/passwd so that
+  # we can move them to /usr (relying on nss-altfiles to provide them
+  # at runtime, but we could use systemd's userdb, too).
+  sudo systemd-sysusers --root="${root_fs_dir}"
+  for databasefile in passwd group shadow gshadow; do
+    newentries=$(comm -23 <(sudo cut -d ":" -f 1 "${root_fs_dir}/etc/${databasefile}" | sort) <(sudo cut -d ":" -f 1 "${root_fs_dir}/usr/share/baselayout/${databasefile}" | sort))
+    for newentry in ${newentries}; do
+      sudo grep "^${newentry}" "${root_fs_dir}/etc/${databasefile}" | sudo tee -a "${root_fs_dir}/usr/share/baselayout/${databasefile}"
```
krnowak, line 624: Would it be safe to use colon after `${newentry}` to match the user/group name completely?

```suggestion
      sudo grep "^${newentry}:" "${root_fs_dir}/etc/${databasefile}" | sudo tee -a "${root_fs_dir}/usr/share/baselayout/${databasefile}"
```

build_library/build_image_util.sh
```diff
@@ -607,6 +607,25 @@ finish_image() {
...
+    done
+    sudo rm "${root_fs_dir}/etc/${databasefile}"
+    sudo rm -f "${root_fs_dir}/etc/${databasefile}-"
```
krnowak, line 627:
```suggestion
    sudo rm -f "${root_fs_dir}/etc/${databasefile}" "${root_fs_dir}/etc/${databasefile}-"
```

build_library/build_image_util.sh
pothos: `/file/owned/by/user/:-` should be matched, too, because it means to pick a random group ID, or?

krnowak: > `/file/owned/by/user/:-` should be matched, too, because it means to pick a random group ID, or?

No, if it's an absolute path then it's treated as a whole as an absolute path, and from that path UID and GID are taken. But there is another corner case I haven't covered yet, probably. You could also use `uid:groupname` syntax, so I'd say that pattern `-:.*` was correct, whereas `-:[0-9]\+` is too narrow. For more info, there's `man sysusers.d` for describing the file format.

krnowak: You could also skip the grep and use `awk '{if ($3 !~ /^\// && ($3 ~ /^-:|:-$|^-$/)) {print $3}}'`, probably.

krnowak: Question: should we also be bailing out on paths, actually?

pothos: > Question: should we also be bailing out on paths, actually?

Ok, let's do it for clarity

krnowak: > (I had thought `/path:-` would be valid, that's why I had asked)

It is valid, but it means "take owner UID and GID from file `path:-` located in `/`.

krnowak (approved):
Looks good.

PR #228, opened 2022-02-16
ci-automation: Add foundational testing primitives

t-lo:
This PR adds testing primitives and tapfile handling to the "containerised builds" `ci-automation`.

Design:
- A generic "test.sh" stub takes care of starting test wrappers within the SDK container. The script will also re-run failed tests to ensure failures are consistent, avoiding breakage from transient failures. The retry count is very high by default (999).
- Image / vendor specific scripts in `ci-automation/vendor-testing/` take care of running the actual tests. These scripts are being run in the SDK container.
- A helper library for querying and for merging tap files is also supplied. It's currently used to figure out if tests failed and should be re-run. It is envisioned this script will later be used by another job to merge all tap files of a release.


ci-automation/test.sh
```diff
@@ -0,0 +1,122 @@
...
+    local container_name="flatcar-tests-${arch}-${docker_vernum}-${image}"
+
+    local retry=""
+    local success=false
+    for retry in $(seq "${retries}"); do
+        local tapfile="results-run-${retry}.tap"
+        local failfile="failed-run-${retry}."
+
+        set -o noglob
+        ./run_sdk_container -n "${container_name}" -C "${packages_image}" -v "${vernum}" \
```
pothos, line 86: While the QEMU tests run inside the SDK container this was not the case for the normal tests and I think it's good to have it that way because the kola binary that should be used is the one supplied by Jenkins and the SDK container would just add overhead to the test job.

ci-automation/test.sh
```diff
@@ -0,0 +1,122 @@
...
+            ci-automation/vendor/testing/"${image}".sh \
+                "${tests_dir}" \
+                "${arch}" \
+                "${vernum}" \
+                "${tapfile}" \
+                $@
+        set +o noglob
+
+        ./run_sdk_container -n "${container_name}" -C "${packages_image}" -v "${vernum}" \
```
pothos, line 95: There was this `-x ci-cleanup.sh` logic which we should use here as well

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,281 @@
+#!/bin/bash
+#
```
pothos, line 2: There are some pipes used and I think we should do `set -euo pipefail` here instead of silencing the errors

ci-automation/test.sh
```diff
@@ -0,0 +1,122 @@
...
+#
+#   1. 2 merged TAP reports with all test runs / vendors.
+#        - a "summary" report which contains error messages only for tests which never succeeded (per vendor).
+#        - a "detailed" report which also contains error messages of transient failures which succeeded after re-runs.
+#        These reports will be updated after each (re-)run of each vendor, making the test job safe
+#          to abort at any point - the previous runs' results won't be lost.
+#   2. "./ci-cleanup.sh" with commands to clean up temporary build resources,
+#        to be run after this step finishes / when this step is aborted.
+
+set -eu
```
pothos, line 50:
```suggestion
set -euo pipefail
```
… or here if it shouldn't be done in the sourced file

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,281 @@
...
+
+# wrapper around sqlite3 w/ retries if DB is locked
+function __sqlite3_wrapper() {
+    local dbfile="$1"
+    shift
+
+    while true; do
+        sqlite3 "${dbfile}" "$@"
+        if [ $? -ne 5 ] ; then
+            return $?
```
pothos, line 28:
```suggestion
            return $?
```
$? is now always 0

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,281 @@
...
+                in_error_message=false
+            else
+                error_message="$(echo -e "$line" \
+                                    | sed -e 's/^Error: "--- FAIL: /"/' -e 's/^[[:space:]]*//' \
+                                    | sed -e "s/[>\"']/_/g" -e 's/[[:space:]]/ /g')"
+                continue
+            fi
+        else
+            test_name="$(echo "${line}" | sed 's/^[^-]* - //')"
+            local result_string="$(echo "${line}" | sed 's/ - .*//')"
```
pothos, line 120:
```suggestion
            local result_string
            result_string="$(echo "${line}" | sed 's/ - .*//')"
```
otherwise the error code of the $() evaluation is masked

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,281 @@
...
+# 3: <include_transient_errors> - If set to "true" then debug output of transient test failures
+#                   is included in the result report.
+function tap_generate_report() {
+    local arch="$1"
+    local version="$2"
+    local full_error_report="${3:-false}"
+
+    local dbname="${TAPFILE_HELPER_DBNAME}"
+
+    local count="$(__sqlite3_wrapper "${dbname}" 'SELECT count(name) FROM test_case;')"
```
pothos, line 189:
```suggestion
    local count
    count="$(__sqlite3_wrapper "${dbname}" 'SELECT count(name) FROM test_case;')"
```

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,281 @@
...
+    local vendors="$(__sqlite3_wrapper "${dbname}" 'SELECT name FROM vendor;' | tr '\n' ' ')"
```
pothos, line 190:
```suggestion
    local vendors
    vendors="$(__sqlite3_wrapper "${dbname}" 'SELECT name FROM vendor;' | tr '\n' ' ')"
```

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,281 @@
...
+    echo "   ..."
+
+    # Print result line for every test, including platforms it succeeded on
+    #  and transient failed runs.
+    __sqlite3_wrapper "${dbname}" 'SELECT DISTINCT name from test_case;' | \
+    while read -r test_name; do
+
+        # "ok" if the test succeeded at least once for all vendors that run the test,
+        #   "not ok" otherwise.
+        local verdict="$(__sqlite3_wrapper "${dbname}" "
```
pothos, line 205:
```suggestion
        local verdict
        verdict="$(__sqlite3_wrapper "${dbname}" "
```

ci-automation/test.sh
```diff
@@ -0,0 +1,122 @@
...
+        set -- *
+    fi
+
+    source ci-automation/tapfile_helper_lib.sh
+    source ci-automation/ci_automation_common.sh
+    init_submodules
+
+    source sdk_container/.repo/manifests/version.txt
+    local vernum="${FLATCAR_VERSION}"
+    local docker_vernum="$(vernum_to_docker_image_version "${vernum}")"
```
pothos, line 67:
```suggestion
    local docker_vernum
   docker_vernum ="$(vernum_to_docker_image_version "${vernum}")"
```

ci-automation/test.sh
```diff
@@ -0,0 +1,122 @@
...
+                "${tapfile}" \
+                $@
+        set +o noglob
+
+        ./run_sdk_container -n "${container_name}" -C "${packages_image}" -v "${vernum}" \
+            ci-automation/test_update_reruns.sh \
+                "${tests_dir}/${tapfile}" "${image}" "${retry}" \
+                "${tests_dir}/failed-run-${retry}.txt"
+
+        local failed_tests="$(cat "${tests_dir}/failed-run-${retry}.txt")"
```
pothos, line 100:
```suggestion
        local failed_tests
        failed_tests="$(cat "${tests_dir}/failed-run-${retry}.txt")"
```

ci-automation/test_update_reruns.sh
```diff
@@ -0,0 +1,20 @@
...
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Helper script for test.sh to update the test failures text file.
+# test.sh uses this to determine which tests need to re-run.
+# This script is run within the SDK container.
+
+set -eu
```
pothos, line 11:
```suggestion
set -euo pipefail
```

ci-automation/vendor-testing/qemu.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# Test execution script for the qemu vendor image.
+# This script is supposed to run in the SDK container.
+
+function run_testsuite() {
```
pothos, line 10: Where does this function get called? I think the call is missing at the end of this file, or?

ci-automation/vendor-testing/qemu.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
```
pothos, line 2:
```suggestion
set -euo pipefail
```

pothos (commented):
In general I would have liked to start from the current working setup instead and adapting it to use the SDK container where needed instead of rearchitecting it which means that we have to rewrite most test platform integrations and get them right which is very time consuming.

pothos:
Would it help to do some of the restart logic directly in kola? We have reprovisioning retries there already and could also do this for the tests itself (with one TAP report entry per test execution).

t-lo:
Thank you very much for the review! I'll implement the feedback today and update the PR.

> In general I would have liked to start from the current working setup instead and adapting it to use the SDK container where needed instead of rearchitecting it which means that we have to rewrite most test platform integrations and get them right which is very time consuming.

I would like to try and prevent this. The current test implementation is tightly coupled to our internal Jenkins logic and will be very hard to generalise. We risk ending up polluting the clean interfaces the containerised CI automation currently employs - just consider the dozens of env vars woven into the current set-up, and our failure to adapt these when we were struggling with the Google bucket issue. I think there is a lot of value in re-thinking our approach to CI and testing.

The main criteria I'm pursuing is: What if you would need to set up the whole infra from scratch - how much effort would that take you? What components / services are you required to use because of assumptions baked into the automation? To give you an example: Can you set up the current infrastructure - or the re-designed containerised infra retrofitted to run the "old" test automation - so it builds for and tests a single platform we don't support (say, openstack)? Do you need AWS and GCP accounts (my assumption would be "yes")? Do you need to use Jenkins? I'd like to keep the effort at a minimum, and make as few assumptions as possible. Doing a re-design seems the faster / easier way to get there.

> Would it help to do some of the restart logic directly in kola? We have reprovisioning retries there already and could also do this for the tests itself (with one TAP report entry per test execution).

I don't think that's the right place; it would make it harder to adapt to custom environments (changing go code and automation instead of just adapting automation glue).

ci-automation/test.sh
t-lo: Makes sense, I'll do some benchmarking on test time. The benefits of this approach is that tests are reproducible - any given release can be re-tested with the tools in its respective SDK.

ci-automation/tapfile_helper_lib.sh
t-lo: Since these are pure library functions I'd leave this to the script sourcing the lib, but it's a good point. Will add it to `test.sh` and to `test_update_reruns.sh`.

ci-automation/vendor-testing/qemu.sh
t-lo: Yes, this script is still work in progress. I did the generic / specific split very late when writing the initial draft, I don't think there needs to be a wrapper function at all.

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,288 @@
...
+# The script uses a temporary SQLite DB for querzing and for result generation.
+#
+# Brief usage overview (scroll down for parameters etc.):
+#   tap_ingest_tapfile - add test results from tap file to the DB 
+#   tap_list_vendors   - list all vendors TAP files have been ingested for
+#   tap_failed_tests_for_vendor - list all tests that never succeded even once, per vendor
+#   tap_generate_report - generate a merged test report
+
+
+TAPFILE_HELPER_DBNAME="results.sqlite3"
```
jepio, line 18: Is this expected to be shared across runs - e.g. if you can't run all vendor tests at the same time? If so then it should be a parameter passed by the caller.

ci-automation/test.sh
```diff
@@ -0,0 +1,129 @@
...
+    source ci-automation/tapfile_helper_lib.sh
+    source ci-automation/ci_automation_common.sh
+    init_submodules
+
+    source sdk_container/.repo/manifests/version.txt
+    local vernum="${FLATCAR_VERSION}"
+    local docker_vernum
+    docker_vernum="$(vernum_to_docker_image_version "${vernum}")"
+
+    local packages="flatcar-packages-${arch}"
```
jepio, line 72: This doesn't have to be the packages container since we shouldn't be depending on any of the built packages. Images are stored externally.

ci-automation/vendor-testing/qemu.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# $@ now contains tests / test patterns to run
+
+source ci-automation/ci_automation_common.sh
+
+mkdir -p "${work_dir}"
+cd "${work_dir}"
+
+echo "++++ QEMU test: downloading ${QEMU_IMAGE_NAME} for ${vernum} (${arch}) ++++"
+copy_from_buildcache "images/${arch}/${vernum}/${QEMU_IMAGE_NAME}" .
```
jepio, line 24: I'm not super fond of the test script reaching out to the buildcache directly, the path to the image should be passed from the outside.

ci-automation/vendor-testing/qemu.sh
```diff
@@ -0,0 +1,38 @@
...
+set -o noglob
+
+sudo kola run \
+    --board="${arch}-usr" \
+    --parallel="${QEMU_PARALLEL}" \
+    --platform=qemu \
+    --qemu-bios=/usr/share/qemu/bios-256k.bin \
+    --qemu-image="${QEMU_IMAGE_NAME}" \
+    --tapfile="${tapfile}" \
+    --torcx-manifest="${CONTAINER_TORCX_ROOT}/${arch}-usr/latest/torcx_manifest.json" \
```
jepio, line 35: This is a similar artifact to the image, so should come from the same source.

ci-automation/vendor-testing/qemu.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -euo pipefail
+
+# Test execution script for the qemu vendor image.
+# This script is supposed to run in the SDK container.
```
jepio, line 9: Do we need to couple this to the SDK container? Look at what we do for aarch64 - kola depends on a working qemu and dnsmasq, nothing more.

ci-automation/test.sh
```diff
@@ -0,0 +1,129 @@
...
+function test_run() {
+    local arch="$1" ; shift
+    local image="$1"; shift
+
+    # default to all tests
+    if [ $# -le 0 ] ; then
+        set -- '*'
+    fi
+
+    local retries="${MAX_RETRIES:-999}"
```
jepio, line 61: Excessive default, this will cause actual failing tests to loop for hours

ci-automation/test.sh
```diff
@@ -0,0 +1,129 @@
...
+#        These reports will be updated after each (re-)run of each vendor, making the test job safe
+#          to abort at any point - the previous runs' results won't be lost.
+#   2. "./ci-cleanup.sh" with commands to clean up temporary build resources,
+#        to be run after this step finishes / when this step is aborted.
+
+set -euo pipefail
+
+function test_run() {
+    local arch="$1" ; shift
+    local image="$1"; shift
```
jepio, line 54: I can't help but think that we should be passing to this function from the outside. Here you're implicitly relying on the container to provide the version but there is no need to couple testing to a specific container version. Kola tests are developed over time and will work with older releases.

jepio (commented):
Just a quick scan of the concepts without reviewing the implementation thoroughly.

ci-automation/tapfile_helper_lib.sh
t-lo: No, it's a transient / temporary throw-away database I use for structuring information from a bunch of TAP files into a query-able format.

ci-automation/test.sh
t-lo: Tests need the torcx manifest which is only available in the packages container. Technically we could extract it from the OS image but that's pretty hacky. Another alternative would be to publish the manifest to bincache, alongside the images.

ci-automation/vendor-testing/qemu.sh
t-lo: It's configurable: https://github.com/flatcar-linux/scripts/blob/main/ci-automation/ci-config.env
This can be changed e.g. by using a separate branch, or overridden by CI (change the default in the env file before calling the script).
In a custom set-up the build cache could e.g. be on the local machine.
However I don't expect this parameter to change often, so I'd like to persist a sane default somewhere so CI is not required to feed this information.

ci-automation/vendor-testing/qemu.sh
t-lo: This is for portability / reproducibility: any given version / release has its test environment pinned to a version that is derived from the git commit in `scripts` which corresponds to that version. I do understand aarch64 will need separate handling since there is no SDK for ARM currently.

ci-automation/test.sh
t-lo: I'd argue keeping hardware busy is the lesser evil.
The motivation is that test failures on actual issues *do* make the test run for hours, until a human is around to abort the test. Think about kicking off tests on Friday and coming back on Monday to look at the actual failures, instead of spending Monday restarting fragile tests.
I realise it's a trade-off, but as long as our tests aren't robust it's a good trade-off.

ci-automation/test.sh
t-lo: In general I agree - but we are specifically focusing on considering CI / release automation here. For reproducible releases I very much want to pin the version of the test suite to the release I've tested it with.
This doesn't mean that old releases cannot be tested with newer kola versions (e.g. by using a newer SDK container) but that shouldn't be the default for releases.

ci-automation/tapfile_helper_lib.sh
t-lo: The TAP files of individual tests should be archived for a build, but not the sqlite file. The same tapfile helper can be used later (with a separate throw-away DB file) to merge all TAP files of all tests.

t-lo:
- fixed some issues in `test.sh` which effectively broke re-runs
- reduced retries to 20 (which feels sufficient for transient issues) in an attempt to address @jepio 's [concern](https://github.com/flatcar-linux/scripts/pull/228#discussion_r809044029).

t-lo:
Updated `packages.sh` and the test script in accordance with this proposal: https://github.com/flatcar-linux/scripts/pull/228#issuecomment-1044571244

ci-automation/tapfile_helper_lib.sh
```diff
@@ -0,0 +1,288 @@
...
+    );
+    CREATE TABLE IF NOT EXISTS "test_run" (
+        "id"        INTEGER NOT NULL,
+        "result"    INTEGER NOT NULL,
+        "output"    TEXT,
+        "case_id"   INTEGER NOT NULL,
+        "run"       INTEGER NOT NULL,
+        "vendor_id" INTEGER,
+        PRIMARY KEY("id"),
+        FOREIGN KEY("case_id") REFERENCES "test_case"("id"),
```
jepio, line 61: foregin keys are not enforced by default, you would need an explicit `PRAGMA foreign_keys = ON;` to enable.

ci-automation/vendor-testing/qemu.sh
jepio: Instead of treating arm64 differently, it would be best to find an approach that works the same way for both. See also the comment from Kai:
> While the QEMU tests run inside the SDK container this was not the case for the other tests and I think it's good to have it that way because the kola binary that should be used is the one supplied by Jenkins and the SDK container would just add overhead to the test job.

ci-automation/vendor-testing/qemu.sh
jepio: I'm fine with persisting the default, but I'm saying I would suggest to separate out "fetching images" from "executing tests". Since test_run() fetches "artifacts" the image fetching can also happen in test.sh. That way "qemu.sh" only has to test a provided image. Think of it this way: instead of passing credentials to buildcache all the way down, we do privileged stuff early, and then we don't need as many privileges deeper in the call stack.

ci-automation/test.sh
jepio: Ourt tests are not *that* flaky, and the super large retry would really only work on friday's test runs, except that test runs from friday/saturday/sunday would stack up and eventually clog the whole CI. So thanks for reducing the retry counter.

ci-automation/test.sh
jepio: I like the "publish to bincache" idea

ci-automation/vendor-testing/qemu.sh
pothos: @jepio Your idea would be to use the `ghcr.io/kinvolk/kola-test-runner` image for all cases? That makes sense and can provide a similar reproducibility that @t-lo aims for here but with the two benefits that we support arm64 from day one and that we don't try to use kola from the SDK which will be outdated but expect it to exist in a folder (supplied by Jenkins or the user).

ci-automation/vendor-testing/qemu.sh
pothos: I think it's also worth considering to pass the URL and the test run can decide to download locally or pass it to kola as done with Digital Ocean where the API takes a public URL

ci-automation/test.sh
```diff
@@ -0,0 +1,184 @@
...
+    local retry=""
+    local success=false
+    for retry in $(seq "${retries}"); do
+        local tapfile="results-run-${retry}.tap"
+        local failfile="failed-run-${retry}.txt"
+
+        # Ignore retcode since tests are flaky. We'll re-run failed tests and
+        #  determine success based on test results (tapfile).
+        set +e -o noglob
+        ./run_sdk_container -x ./ci-cleanup.sh \
```
pothos, line 145: It would be good to make Jenkins' `bin/` folder of the workspace available to use the provided kola binaries. The SDK kola binary is likely outdated and is hard to update for bug fixes. I think a well-known path for outside-provided kola binaries should always take precedence over the internal kola binary of the SDK.

t-lo: I'd argue we should rather update the SDK container more often then - we can make this straightforward and even automate the process for new kola releases. Motivation of using the test suite from the SDK container is that the test suite is versioned alongside each release version, and release tests are reproducible.

t-lo: Note that updating kola in the SDK container does not necessarily mandate a full SDK rebuild, as it would for the cork SDK - we could base the updated SDK on an existing SDK container, and supply a respective `Dockerfile` that only updates `kola`.

For example, consider SDK 3139.0.0 needs a `kola` update. We could provide a Dockerfile that builds a new SDK version 3139.1.0 based on the 3139.0.0 image, with an updated `kola` binary.

pothos: With the current bind-mounts we could do something like `mv WORKSPACE/bin __build__/images/` to have the kola binaries available under `/mnt/host/source/src/build/bin` from inside the SDK

pothos: > I'd argue we should rather update the SDK container more often then

The main problem is that we don't want to touch released SDKs which we would have to do in the two cases: 1. during a release we finished building everything and just have to fix a test case and your proposed way requires us to update the SDK image manually to do so. 2. the test's URLs for e.g., the kubeadm test change and the test fail due to this and we have to update the released SDKs now for users to be able to run the test suite successfully

t-lo: Really like the idea of injecting newer kola versions but I'd like to make it visible for users. Currently we don't. We also don't test new kola releases with old OS images, so there's likely bit-rot in that direction, too.

I wouldn't touch the SDKs once released, but we should make it easy for ourselves to supply an updated SDK for new (minor version and/or patch level) releases. I don't think this requires manually updating the SDK. Consider this:

```dockerfile
FROM ghcr.io/flatcar-linux/flatcar-sdk-all:3139.0.0
RUN ./update_chroot
```

being run with `docker build -t ghcr.io/flatcar-linux/flatcar-sdk-all:3139.1.0 .` would provide a new SDK with an updated kola (assuming the kola ebuild's `CROS_WORKON_COMMIT` points to the new version).

pothos: Yes, it can be that some tests don't work on year old releases but for these cases one can use an older kola release (e.g., the one from the SDK at that time). In general we have the version filters to ensure that kola works with multiple release at the same time.

I don't follow the update_chroot example because the assumption that the ebuild file gets updated only works when the tag is forced pushed, or?

t-lo: > I don't follow the update_chroot example because the assumption that the ebuild file gets updated only works when the tag is forced pushed, or?

Force-push would not be necessary since we'd be releasing a new SDK (with a minor version / patch level version bump).

Trying to address the 2 cases you've stated:
> 1. during a release we finished building everything and just have to fix a test case and your proposed way requires us to update the SDK image manually to do so.
> 2. the test's URLs for e.g., the kubeadm test, change (or we have other fixes for the test because it is flaky etc) and the test fail due to this and we have to update the released SDKs now for users to be able to run the test suite successfully

In both cases I would release a new SDK based on an existing container, with minor / patch version bump. My argument is that this could be done automatically, and it should be done to keep the release version in sync with the environment used to build and to test it.

I've worked on a script which can be used to automate this: https://github.com/flatcar-linux/scripts/pull/230

t-lo: I've elaborated on the scenarios here: https://github.com/flatcar-linux/scripts/pull/230#issuecomment-1047612468

jepio: > We also don't test new kola releases with old OS images, so there's likely bit-rot in that direction, too.

Actually, this is exactly what we do. All our release channels (including lts) use the current kola build.

jepio:
I would be happy to have this merged and integrated into CI, but with standalone builds of kola outside of the SDK container.

t-lo:
> I would be happy to have this merged and integrated into CI, but with standalone builds of kola outside of the SDK container.

I agree this is a different topic to tackle. With standalone builds I would like to see a proposal on how I can reproduce the tests locally (or on another server) without having access to the parameters of the respective Jenkins build.

ci-automation/vendor-testing/qemu.sh
t-lo: I like this idea! Makes a lot of sense.

t-lo: We could wire up kola test runner container in `version.txt` (or another versionfile) the same way we already wire up the SDK version. The test scripts could be amended to use the test runner instead of the SDK container.

ci-automation/test.sh
t-lo: The response above was to an argument where users could reproduce release tests of old, released, OS images with new, updated kola versions. This is not guaranteed to work since we do not test this.

pothos (approved):

t-lo:
PR cherry-picked to flatcar-3033, flatcar-3066, and flatcar-3139.

PR #230, opened 2022-02-21
update_sdk_container_image: New SDK container from existing image (minor / patch updates)

t-lo:
This change introduces update_sdk_container_image, a script to generate a new SDK container image based on an existing SDK container. The script is meant to be used for minor / patch level SDK changes (like test suite updates).

### DIFF
--- a/sdk_lib/Dockerfile.sdk-update
+++ b/sdk_lib/Dockerfile.sdk-update
@@ -1,0 +1,11 @@
+ARG BASE
+
+FROM ${BASE}
+COPY --chown=sdk:sdk sdk_container/ /mnt/host/source
+COPY --chown=sdk:sdk . /mnt/host/source/src/scripts
+
+RUN chown sdk:sdk /mnt/host/source
+RUN /home/sdk/sdk_entry.sh ./update_chroot --toolchain_boards="amd64-usr arm64-usr"
+
+RUN /home/sdk/sdk_entry.sh ./setup_board --board="arm64-usr" --regen_configs
+RUN /home/sdk/sdk_entry.sh ./setup_board --board="amd64-usr" --regen_configs
--- a/sdk_lib/sdk_container_common.sh
+++ b/sdk_lib/sdk_container_common.sh
@@ -61,6 +61,11 @@ function get_sdk_version_from_versionfile() {
 }
 # --
 
+function get_version_from_versionfile() {
+    ( source "$sdk_container_common_versionfile"; echo "$FLATCAR_VERSION"; )
+}
+# --
+
 # return true if a given version number is an official build
 #
 function is_official() {
--- a/update_sdk_container_image
+++ b/update_sdk_container_image
@@ -1,0 +1,101 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# This script will update an SDK container image and create a new minor version.
+
+
+set -eu
+set -x
+
+cd $(dirname "$0")
+source sdk_lib/sdk_container_common.sh
+
+os_version="$(get_version_from_versionfile)"
+base_sdk_version="$(get_sdk_version_from_versionfile)"
+new_sdk_version=""
+
+keep="false"
+cleanup=""
+
+usage() {
+    echo "  $0 - Update SDK container image."
+    echo "       Create a new container image based on the current SDK ($base_sdk_version)"
+    echo "       with current changes from coreos-overlay and portage-stable."
+    echo
+    echo "       Just like build_sdk_container_image the resulting container comes in 3 flavours:"
+    echo "        1.    flatcar-sdk-all - includes both ARM64 and AMD64 support"
+    echo "        2.+3. flatcar-sdk-(amd64|arm64) - only includes support for one target."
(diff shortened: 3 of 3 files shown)


t-lo:
@jeremi @pothos this is my proposed solution for injecting new Mantle version into the testing process. Two [scenarios discussed in the container testing PR](https://github.com/flatcar-linux/scripts/pull/228#discussion_r811322385) could be addressed by using this primitive:

> 1. during a release we finished building everything and just have to fix a test case and your proposed way requires us to update the SDK image manually to do so.
> 2. the test's URLs for e.g., the kubeadm test, change (or we have other fixes for the test because it is flaky etc) and the test fail due to this and we have to update the released SDKs now for users to be able to run the test suite successfully.

3. Furthermore, it can be used to build a new Mantle version from a dev branch into a new SDK version, then use the new SDK to run tests for an existing build - thus testing the new Mantle build.

This would work as follows:
1. Pick a nightly or a release version to base the changes on - for example, `alpha-3139.0.0`.
2. Create a dev branch in coreos-overlay, update the [Mantle ebuild to point to the new commit](https://github.com/flatcar-linux/coreos-overlay/tree/main/coreos-devel/mantle), and bump the ebuild `-rc` version.
3. Use `update_sdk_container_image` to create a new SDK version with the Mantle changes based on `alpha-3139.0.0` and the coreos-overlay dev branch. Since the update script bases its work on an existing SDK container it's rather fast. We now have created sdk 3139.0.1. Note we did not change the Flatcar OS image version. Create a dev tag on the scripts repo, say `sdk-3139.0.1-mantletest1` to save state.
4. Run the `tests.sh` script with the `sdk-3139.0.1-mantletest1` scripts branch. It will use `3139.0.0` OS images since the OS version did not change.

Step 1. - 3. above could be a mantle build job in `ci-automation`. 3. is just re-using `tests.sh` from https://github.com/flatcar-linux/scripts/pull/228 .

Benefit over mantle injection in the traditional build is that we do not rely on state in the build server variables (i.e. we don't need to review CI settings to reproduce the build). We would also gain a fully usable and publishable new SDK container with the mantle fixes, for free.

pothos:
So, for this SDK to be used it means that the original scripts tag used to build the image gets a force push because otherwise it would still point to the old SDK in its `version.txt`. I see how this can work but overall my opinion is that there are many hoops to jump through for little benefit. If we want to log with which kola version we tested an image we can could also add this to https://stable.release.flatcar-linux.net/amd64-usr/current/version.txt - I don't see a strong reason to freeze the kola version as we will probably recommend to use the latest version most time.
Anyway, no problem with merging this PR.

jepio:
I don't see this approach as an improvement. We want to test images with the current test suite, because the test cases are versioned (MinVersion/MaxVersion) and we get improved test coverage for older release over time. The approach from this PR adds additional steps to the process, and requires updating multiple branches for mantle commits.

> Benefit over mantle injection in the traditional build is that we do not rely on state in the build server variables (i.e. we don't need to review CI settings to reproduce the build).
There are times when you want to version the test suite together with the code, but this is not one of those times. Here the test suite is decoupled from the code and so you explicitly don't want to do that. The build server variables serve as a way to iterate quickly, no one needs them to be reproduceable.

> We would also gain a fully usable and publishable new SDK container with the mantle fixes, for free.
We don't have any further use for this SDK container so the gain is questionable. Getting it to consumers would require tag rewriting or micro-managing versions/commits in scripts.

t-lo:
> So, for this SDK to be used it means that the original scripts tag used to build the image gets a force push because otherwise it would still point to the old SDK in its version.txt

No, because we are running the tests from the new (`sdk-3139.0.1`) branch. The SDK release version is updated, the OS release version remains the same.

> I don't see this approach as an improvement. We want to test images with the current test suite, because the test cases are versioned (MinVersion/MaxVersion) and we get improved test coverage for older release over time.

Quick iterations are even simpler: update the SDK with the new Mantle changes in coreos-overlay, then run tests, all in a single job. No tagging involved. The example I gave was to fix a test when we're about to publish a release.

> There are times when you want to version the test suite together with the code, but this is not one of those times. Here the test suite is decoupled from the code and so you explicitly don't want to do that. The build server variables serve as a way to iterate quickly, no one needs them to be reproduceable.

We currently inject custom test suite builds into the release process. To reproduce, I would need the combined state of 3 code repositories, one of 2 manifest repositories, a google bucket, and all of the 20-ish variables that went into the respective Jenkins build pipeline. Even then, as soon as the Jenkins build parameters are garbage collected, I will not be able to reproduce either build or tests. I realise we have gotten used to this, but having state restored by a simple "git checkout" is a massive improvement.

pothos:
I'm still not understanding who would know about and use this new `sdk-3139.0.1` branch to run kola. With retagging it could at least find its way to the user (assuming that they can fetch changed tags).

t-lo:
> I'm still not understanding who would know about and use this new sdk-3139.0.1 branch to run kola. With retagging it could at least find its way to the user (assuming that they can fetch changed tags).

My thinking is that at the point when we update kola, we did not actually publish a release yet - because we're fixing a test to get the release out. When the OS release is published it would then reference the updated SDK version with the "new" kola.

pothos (approved):

t-lo:
PR cherry-picked to flatcar-3033, flatcar-3066, and flatcar-3139.

PR #231, opened 2022-02-22
build_library/build_image_util: disable ebuild-locks when merging bin…

jepio:
# build_library/build_image_util: disable ebuild-locks when merging binary packages

Set `FEATURES=-ebuild-locks` for the duration of the binary package merge to the image. This makes the operation much faster (20min -> 10min) and I haven't seen any issues with it so far.

## How to use

```
time ./build_image --board=amd64-usr
```

## Testing done

Have been using this for months, and CI running.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -155,6 +155,7 @@ emerge_to_image() {
   fi
 
   sudo -E ROOT="${root_fs_dir}" \
+      FEATURES="-ebuild-locks" \
       PORTAGE_CONFIGROOT="${BUILD_DIR}"/configroot \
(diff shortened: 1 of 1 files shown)


pothos:
Thanks, sounds very useful.
Is `parallel-install` also active from `coreos-overlay/profiles/coreos/base/make.defaults` and if yes, should both go there or if not, should `parallel-install` go here?

jepio:
`parallel-install` is active from the profile and I would leave it there. We could put `-ebuild-locks` in the profile as well, but I felt more comfortable adding it to the build_image phase only. It also makes the largest difference there.

jepio:
In CI the effect is 55 minutes -> 31 minutes.

pothos (approved):

PR #232, opened 2022-02-22
ci-automation: publish test results, add to GC

t-lo:
This change adds copying test results to the build cache server, and adds respective deletion to the garbage collector.
Also, the patch fixes an issue with torcx publishing (manifest publishing had arch hard-coded).

### DIFF
--- a/ci-automation/garbage_collect.sh
+++ b/ci-automation/garbage_collect.sh
@@ -92,6 +92,7 @@ function garbage_collect() {
             rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/boards/*/${os_vernum}/"
             rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/containers/${os_docker_vernum}/flatcar-images-*"
             rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/images/*/${os_vernum}/"
+            rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/testing/${os_vernum}/"
         fi
 
         echo "## The following files will be removed ##"
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -179,7 +179,7 @@ function test_run() {
     # publish TAP files to build cache
     copy_to_buildcache "testing/${vernum}/${arch}/${image}" \
         "${tests_dir}/"*.tap
-    copy_to_buildcache "testing/${vernum}/${arch}/${image}/debug" \
-        "${tests_dir}/_kola_temp/"*
+    copy_to_buildcache "testing/${vernum}/${arch}/${image}" \
+        "${tests_dir}/_kola_temp"
 }
 # --
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -176,7 +176,7 @@ function test_run() {
         echo "########### All re-runs exhausted ($retries). Giving up. ###########"
     fi
 
-    # publish TAP files to buold cache
+    # publish TAP files to build cache
     copy_to_buildcache "testing/${vernum}/${arch}/${image}" \
         "${tests_dir}/"*.tap
     copy_to_buildcache "testing/${vernum}/${arch}/${image}/debug" \
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -176,9 +176,10 @@ function test_run() {
         echo "########### All re-runs exhausted ($retries). Giving up. ###########"
     fi
 
-    # TODO: publish to bincache?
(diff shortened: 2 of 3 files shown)


ci-automation/test.sh
```diff
@@ -176,9 +176,10 @@ function test_run() {
...
     fi
 
-    # TODO: publish to bincache?
-    # "${tests_dir}/"*.tap
-    # "${tests_dir}/_kola_temp.tar.xz"
-
+    # publish TAP files to buold cache
+    copy_to_buildcache "testing/${vernum}/${arch}/${image}" \
+        "${tests_dir}/"*.tap
+    copy_to_buildcache "testing/${vernum}/${arch}/${image}/debug" \
```
jepio, line 182: nit: why not just put it next to the tap file, skipping the debug subdir?.

ci-automation/test.sh
```diff
@@ -176,9 +176,10 @@ function test_run() {
         echo "########### All re-runs exhausted ($retries). Giving up. ###########"
     fi
 
-    # TODO: publish to bincache?
-    # "${tests_dir}/"*.tap
-    # "${tests_dir}/_kola_temp.tar.xz"
-
+    # publish TAP files to buold cache
```
jepio, line 179:
```suggestion
    # publish TAP files to build cache
```

ci-automation/garbage_collect.sh
```diff
@@ -92,6 +92,7 @@ function garbage_collect() {
             rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/boards/*/${os_vernum}/"
             rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/containers/${os_docker_vernum}/flatcar-images-*"
             rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/images/*/${os_vernum}/"
+            rmpat="${rmpat} ${BUILDCACHE_PATH_PREFIX}/testing/${os_vernum}/"
```
jepio, line 95: Should this also be `testing/*/${os_vernum}`?

ci-automation/test.sh
pothos: I would test for existence first before doing the call because it would fail.

ci-automation/garbage_collect.sh
t-lo: No - for compatibility reasons we store packages and images at <arch>/<version> (like in the google bucket) - but this makes it hard to browse as Packages and images of the same version are scattered across 4 directories.
For the testing artifacts I've changed that, so everything belonging to one version is below the same version directory.
This corresponds to https://github.com/flatcar-linux/scripts/blob/3397167b5e90d155cf3f64d00b4d4e40efbf2121/ci-automation/test.sh#L180-L183 (`... "testing/${vernum}/${arch}/${image}" ...`).

ci-automation/test.sh
t-lo: `copy_to_buildcache` uses `rsync` [under the hood](https://github.com/flatcar-linux/scripts/blob/main/ci-automation/ci_automation_common.sh#L127), so existing directories would be extended, and existing files would be replaced with newer versions where applicable.

jepio: The *.tap and _kola_temp/* may not match any files and will cause rsync to fail

t-lo: Ah, gotcha. In that case, `kola` did not run at all, and the script would have aborted earlier (since we use `set -euo pipefail`).

pothos: Yes, was talking about that. In this case we would still fail here because rsync does not find a file under the unexpanded name, so no problem with having it missing.

t-lo:
Re-capturing conversation from https://github.com/flatcar-linux/scripts/pull/232#discussion_r812171729 ff. since the underlying code changed and the conversation is not visible anymore:

Kai
> I would test for existence first before doing the call because it would fail.

Jeremi
> The *.tap and _kola_temp/* may not match any files and will cause rsync to fail

Thilo
> Ah, gotcha. In that case, kola did not run at all, and the script would have aborted earlier (since we use set -euo pipefail).

Kai
> Yes, was talking about that. In this case we would still fail here because rsync does not find a file under the unexpanded name, so no problem with having it missing.

I believe we would fail earlier - no .tap file means no test started successfully, which would make the script fail when trying to collect tap files for analysis (line 158ff.). And it's OK to fail then, because test runs failed.

jepio (approved):

t-lo:
Cherry-picked to `flatcar-3033`, `flatcar-3066`, `flatcar-3139`.

PR #234, opened 2022-02-23
ci-automation/test*: publish per-image summary report

t-lo:
Generate and publish per-image TAP reports for CI consumption.


t-lo:
Cherry-picked to flatcar-3033, flatcar-3066, and flatcar-3139 just now.

PR #237, opened 2022-02-23
ci-automation: SDK build updates version.txt in main branch

t-lo:
This change has sdk_bootstrap push the branch to origin when run from the main branch, updating the SDK and OS version in 'main' for each SDK bootstrap build.

Release / maintenance branches have the SDK version set in the versionfile at release time. But main is never updated.

Updating the versionfile in main when a new SDK is built ensures that dev branches based on main will also use the correct SDK version (e.g.in subsequent CI builds). As an example, this PR makes changes like these: https://github.com/flatcar-linux/scripts/pull/235/files#diff-48b3a88e2d986f5123acefc4dd59b5e6ae86f45975cc28e105e3ed1b3815f420L4 unnecessary.

Note this change implies we'll have one automated commit each night in our main branch, from the nightly build.

As a (positive) side effect this will also update `main`s submodule pinnings to the `main` tips in coreos-overlay and portage-stable every night.

### DIFF
--- a/ci-automation/sdk_bootstrap.sh
+++ b/ci-automation/sdk_bootstrap.sh
@@ -86,10 +86,13 @@ function sdk_bootstrap() {
 
     # Create new tag in scripts repo w/ updated versionfile + submodules.
     # Also push the changes to the branch ONLY IF we're doing a nightly
-    #   build of the 'main' branch AND we're definitely ON the main branch.
+    #   build of the 'main' branch AND we're definitely ON the main branch
+    #   (`scripts` and submodules).
     local push_branch="false"
     if   [[ "${version}" =~ ^main-[0-9.]+-nightly-[-0-9]+$ ]] \
-       && [ "$(git rev-parse --abbrev-ref HEAD)" = "main"  ] ; then
+       && [ "$(git rev-parse --abbrev-ref HEAD)" = "main"  ] \
+       && [ "$(git -C sdk_container/src/third_party/coreos-overlay/ rev-parse --abbrev-ref HEAD)" = "main"  ] \
+       && [ "$(git -C sdk_container/src/third_party/portage-stable/ rev-parse --abbrev-ref HEAD)" = "main"  ] ; then
         push_branch="true"
     fi
     update_and_push_version "sdk-${git_vernum}" "${push_branch}"
--- a/ci-automation/sdk_bootstrap.sh
+++ b/ci-automation/sdk_bootstrap.sh
@@ -85,10 +85,11 @@ function sdk_bootstrap() {
     cd -
 
     # Create new tag in scripts repo w/ updated versionfile + submodules.
-    # When on the 'main' branch then also push to the branch so the versionfile
-    #  and git submodules stay up to date.
+    # Also push the changes to the branch ONLY IF we're doing a nightly
+    #   build of the 'main' branch AND we're definitely ON the main branch.
     local push_branch="false"
-    if [ "$(git rev-parse --abbrev-ref HEAD)" = "main" ] ; then
+    if   [[ "${version}" =~ ^main-[0-9.]+-nightly-[-0-9]+$ ]] \
+       && [ "$(git rev-parse --abbrev-ref HEAD)" = "main"  ] ; then
         push_branch="true"
     fi
     update_and_push_version "sdk-${git_vernum}" "${push_branch}"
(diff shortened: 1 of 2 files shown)


ci-automation/sdk_bootstrap.sh
```diff
@@ -84,7 +84,13 @@ function sdk_bootstrap() {
     copy_to_buildcache "sdk/${ARCH}/${vernum}" "${dest_tarball}"*
     cd -
 
-    # Create new tag in scripts repo w/ updated versionfile + submodules
-    update_and_push_version "sdk-${git_vernum}"
+    # Create new tag in scripts repo w/ updated versionfile + submodules.
+    # When on the 'main' branch then also push to the branch so the versionfile
+    #  and git submodules stay up to date.
+    local push_branch="false"
+    if [ "$(git rev-parse --abbrev-ref HEAD)" = "main" ] ; then
```
pothos, line 91: This should only be done in nightly builds, not in normal dev builds that happen to use `main` here

t-lo: Updated.

ci-automation/sdk_bootstrap.sh
```diff
@@ -84,7 +84,14 @@ function sdk_bootstrap() {
     copy_to_buildcache "sdk/${ARCH}/${vernum}" "${dest_tarball}"*
     cd -
 
-    # Create new tag in scripts repo w/ updated versionfile + submodules
-    update_and_push_version "sdk-${git_vernum}"
+    # Create new tag in scripts repo w/ updated versionfile + submodules.
+    # Also push the changes to the branch ONLY IF we're doing a nightly
+    #   build of the 'main' branch AND we're definitely ON the main branch.
+    local push_branch="false"
+    if   [[ "${version}" =~ ^main-[0-9.]+-nightly-[-0-9]+$ ]] \
```
pothos, line 91: Two more things to check are the `coreos-overlay` and `portage-stable` references.
Currently the doc string in Jenkins tells this: `use a suffix for dev builds (e.g. alpha-3234.0.0-nightly-1234).` and by checking for the two references we can at least prevent to have something else than main pop up here. People may also retrigger the nightly build to test a `jenkins-os` change but that's hard to catch here.

jepio: I'd like to suggest a different approach. Placeholder value for SDK_VERSION (sdk-main-nightly) that does the lookup and finds the last built nightly SDK (from version.txt somewhere).

But I also don't understand what is failing here: the os->manifest->sdk chain uses the last released SDK as a base and works correctly for nightlies, so why does this chain need special handling?

t-lo: > Two more things to check are the coreos-overlay and portage-stable references.
> Currently the doc string in Jenkins tells this: use a suffix for dev builds (e.g. alpha-3234.0.0-nightly-1234).

That's bad documentation (and a good catch), filed a PR to the CI to update the docstring. Will add a check for the submodules branches.

> People may also retrigger the nightly build to test a jenkins-os change but that's hard to catch here.

I don't think that's a pattern we should encourage - if for some reason someone needs to test the CI then starting the sdk job is way leaner.

t-lo: Updated.

pothos: A placeholder plus a way of automatically updating the submodules would work, too - both seem to have pros and cons…

t-lo: > I'd like to suggest a different approach. Placeholder value for SDK_VERSION (sdk-main-nightly) that does the lookup and finds the last built nightly SDK (from version.txt somewhere).

This is the SDK bootstrap job, it generates a new SDK version (SDK `version` number is an input to that job). It then updates `version.txt` and generates a commit, so follow-up jobs use the correct (new build) SDK version.  

> But I also don't understand what is failing here: the os->manifest->sdk chain uses the last released SDK as a base and works correctly for nightlies, so why does this chain need special handling?

Normally, SDK builds push build changes to a tag but do not update the branch. That tag is the used downstream to continue the build pipeline with the right version. This section introduces special handling for nightly builds of the main branch, which will update the main branch of the `scripts` repo with the new versionfile. Motivation is to make development branches based on `main` use the correct SDK version and remove the necessity for hacks like e.g. this one: https://github.com/flatcar-linux/scripts/pull/235/files#diff-48b3a88e2d986f5123acefc4dd59b5e6ae86f45975cc28e105e3ed1b3815f420L4

pothos (approved):

PR #240, opened 2022-02-25
ci-automation: foundation for package + sdk update

t-lo:
This PR adds foundational functions to ci-automation for CI to update ebuilds which use `CROS_WORKON_COMMIT` and generate a new SDK container image with the changes (based on an existing SDK container).

Updating the SDK container (instead of building a new SDK from scratch) is meant to be used for patch-level changes which do not affect OS images built with the SDK. For example, an SDK's mantle suite (kola etc.) could be updated this way.

To update mantle to the latest commit in the default branch, higher-level CI would run:
1. `ci-automation/util/update_cros_workon.sh sdk_container/src/third_party/coreos-overlay/coreos-devel/mantle/ HEAD push`
2. `source ci-automation/sdk_container_update.sh`
3. `sdk_container_update`

`sdk_container_update` would create and push a tag to `scripts` when can then be used bu successive jobs, e.g. `test.sh`.

### DIFF
--- a/ci-automation/sdk_container_update.sh
+++ b/ci-automation/sdk_container_update.sh
@@ -64,7 +64,7 @@ function sdk_container_update_generate_new_version() {
 
     local current_major_minor="$(echo "${current_vernum}" | sed 's/^\([0-9]\+\.[0-9]\+\).*/\1/')"
     local current_patchlevel="$(echo "${current_vernum}" | sed 's/^[0-9]\+\.[0-9]\+\.\([0-9]\+\).*/\1/')"
-    local current_suffix="$(echo "${current_vernum}" | sed 's/^\([0-9]\+\.[0-9]\+\.[0-9]\+\)//')"
+    local current_suffix="$(echo "${current_vernum}" | sed 's/^[0-9]\+\.[0-9]\+\.[0-9]\+//')"
 
     current_patchlevel="$((current_patchlevel + 1))"
     echo "${current_major_minor}.${current_patchlevel}${current_suffix}"
--- a/ci-automation/sdk_container_update.sh
+++ b/ci-automation/sdk_container_update.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# >>> This file is supposed to be SOURCED from the repository ROOT. <<<
+#
+# sdk_container_update() should be called w/ the positional INPUT parameters below.
+
+# SDK container update automation stub.
+#  This script will create a new SDK container image based on an existing container image.
+#  It is meant to be used for updating tools in the SDK that do not directly affect OS image
+#  generation - like e.g. the mantle suite tools (kola, ore, etc.).
+#
+# PREREQUISITES:
+#
+#   1. SDK version is recorded in sdk_container/.repo/manifests/version.txt and a matching source
+#       SDK container is available on build cache at "/containers/[VERSION]/flatcar-sdk-all-[VERSION].tar.gz"
+#         (nightly / dev builds)
+#      OR
+#       SDK image is available on ghcr.io/flatcar-linux/flatcar-sdk-all:[VERSION]
+#         (official SDK releases)
+#      Specifically, the "-all" version of the source SDK image is required.
+#
+# OPTIONAL INPUT
+#
+#   1. Target SDK version number (without alpha-/beta-/etc. prefix).
+#      If not provided it will be computed from the source SDK, and the patch level is bumped
+#      (e.g. 3139.0.0 => 3139.0.1).
+#
+#   2. coreos-overlay repository tag to use (commit-ish).
+#       Optional - use scripts repo sub-modules as-is if not set.
+#       This version will be checked out / pulled from remote in the coreos-overlay git submodule.
+#       The submodule config will be updated to point to this version before the TARGET SDK tag is created and pushed.
+#
+#   3. portage-stable repository tag to use (commit-ish).
+#       Optional - use scripts repo sub-modules as-is if not set.
+#       This version will be checked out / pulled from remote in the portage-stable git submodule.
+#       The submodule config will be updated to point to this version before the TARGET SDK tag is created and pushed.
+#
+# OUTPUT:
+#
+#   1. SDK container image of the new SDK, published to buildcache
+#       at "/containers/[NEW_VERSION]/flatcar-sdk-all-[NEW_VERSION].tar.gz"
+#
+#   2. New SDK version tagged in scripts and tag pushed to origin.
+#
+#   3. "./ci-cleanup.sh" with commands to clean up temporary build resources,
+#        to be run after this step finishes / when this step is aborted.
+
+set -euo pipefail
+
+# Helper function to generate a new version from the current SDK version
+#  in version.txt.
+# Can also be used by CI automation to generate "$1" for sdk_container_update
+#  coreos_git and portage_git are to be provided.
+#
+function sdk_container_update_generate_new_version() {
+    source ci-automation/ci_automation_common.sh
+    source sdk_container/.repo/manifests/version.txt
+
+    local current_vernum="${FLATCAR_SDK_VERSION}"
+
+    local current_major_minor="$(echo "${current_vernum}" | sed 's/^\([0-9]\+\.[0-9]\+\).*/\1/')"
+    local current_patchlevel="$(echo "${current_vernum}" | sed 's/^[0-9]\+\.[0-9]\+\.\([0-9]\+\).*/\1/')"
+    local current_suffix="$(echo "${current_vernum}" | sed 's/^\([0-9]\+\.[0-9]\+\.[0-9]\+\)//')"
+
+    current_patchlevel="$((current_patchlevel + 1))"
+    echo "${current_major_minor}.${current_patchlevel}${current_suffix}"
+}
+# --
+
+function sdk_container_update() {
+    local vernum="${1:-}"
+    local coreos_git="${2:-}"
+    local portage_git="${3:-}"
+
+    source ci-automation/ci_automation_common.sh
+    init_submodules
+
+    # Make sure source SDK is present for `update_sdk_container_image`
+    source sdk_container/.repo/manifests/version.txt
+    local source_docker_vernum="$(vernum_to_docker_image_version "${FLATCAR_SDK_VERSION}")"
+    docker_image_from_registry_or_buildcache "flatcar-sdk-all" "${source_docker_vernum}"
+    local source_sdk_image="$(docker_image_fullname "flatcar-sdk-all" "${source_docker_vernum}")"
+    echo "docker image rm -f '${source_sdk_image}'" >> ./ci-cleanup.sh
+
+    if [ -n "${coreos_git}" ] ; then
+        update_submodule "coreos-overlay" "${coreos_git}"
+    fi
+    if [ -n "${portage_git}" ] ; then
+        update_submodule "portage-stable" "${portage_git}"
+    fi
+
+    if [ -z "${vernum}" ] ; then
+        vernum="$(sdk_container_update_generate_new_version)"
+    fi
+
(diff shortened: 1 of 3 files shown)


ci-automation/util/update_cros_workon.sh
```diff
@@ -0,0 +1,79 @@
...
+                "${ebuild}"
+
+        rm "${softlink}"
+
+        echo "Updated '${ebuild}' to commit '${commit_id}'"
+        echo " ('${softlink}' ==> '${new_softlink}'"
+
+        if [ "${commit}" = "true" ] then
+            git add .
+            git commit -m "${new_softlink}: Update ${name} to latest ${branch}"
```
jepio, line 64: This commit is not pushed anywhere

ci-automation/sdk_container_update.sh
```diff
@@ -0,0 +1,109 @@
...
+    source ci-automation/ci_automation_common.sh
+    source sdk_container/.repo/manifests/version.txt
+
+    local current_vernum="${FLATCAR_SDK_VERSION}"
+
+    local current_major_minor="$(echo "${current_vernum}" | sed 's/^\([0-9]\+\.[0-9]\+\).*/\1/')"
+    local current_patchlevel="$(echo "${current_vernum}" | sed 's/^[0-9]\+\.[0-9]\+\.\([0-9]\+\).*/\1/')"
+    local current_suffix="$(echo "${current_vernum}" | sed 's/^\([0-9]\+\.[0-9]\+\.[0-9]\+\)//')"
+
+    current_patchlevel="$((current_patchlevel + 1))"
```
jepio, line 69: how do you prevent collisions here?

ci-automation/util/update_cros_workon.sh
```diff
@@ -0,0 +1,79 @@
...
+#
+# OPTIONAL INPUT
+#
+#   2. Branch to update the ebuild to. The latest commit of the branch will be used.
+#       Defaults to HEAD (i.e. the repo's default branch).
+#   3. "true" if a git commit (both in the submodule as well as in scripts) should be created.
+#       Defaults to "false".
+#
+
+function update_cros_workon() {
```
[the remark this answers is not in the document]
t-lo: It's called in [line 78](https://github.com/flatcar-linux/scripts/pull/240/files#diff-5e9df63c2f49c6196f9687924e21f7708d8a97cfafffed60b6cd613bdb5b159aR78) of `update_cros_workon.sh`. It's a pattern to make a script runnable as well as source-able from another script.

ci-automation/sdk_container_update.sh
t-lo: I'd argue it's out of scope for this script and should be ensured by whoever called the script. When in doubt there's always the `version` parameter to `sdk_container_update` ([line 75](https://github.com/flatcar-linux/scripts/pull/240/files/fae06c6d7fe173a3eb55e2040fd3eebfc7b77c0b#diff-19c30e9a3870cff9d8ab067cdbac510a9b281976406c1f54c2ae9ba718367fbcR75)).

ci-automation/util/update_cros_workon.sh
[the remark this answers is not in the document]
t-lo: Oooh, this is nice, thank you!

t-lo:
> We already have a mechanism for updating packages in coreos-overlay through github actions, and that's what we should be using.

@jepio Where can I find the script for that action? We might want to merge the two.

ci-automation/util/update_cros_workon.sh
jepio: Unless i'm reading it wrong, `update_and_push_version` pushes the scripts repository, nothing pushes the commit in coreos-overlay. So this is building a container from a dangling commit.

jepio:
> > We already have a mechanism for updating packages in coreos-overlay through github actions, and that's what we should be using.
> 
> @jepio Where can I find the script for that action? We might want to merge the two.

https://github.com/flatcar-linux/coreos-overlay/tree/main/.github/workflows

But what I'm trying to say is that this PR is pursuing the wrong approach. We have a good way of fetching the latest built kola binaries from CI for a job, and we have github actions to update package ebuilds. Building throwaway SDKs is weird.

ci-automation/util/update_cros_workon.sh
t-lo: Oh, that's correct! Great catch, will fix.

ci-automation/sdk_container_update.sh
t-lo: But the change is pushed at the very last line of `update_sdk_container`, which means the step succeeded. Why would we retrigger a step that succeeded?

jepio: To get another mantle update.

t-lo:
> > > We already have a mechanism for updating packages in coreos-overlay through github actions, and that's what we should be using.
> > 
> > 
> > @jepio Where can I find the script for that action? We might want to merge the two.
> 
> https://github.com/flatcar-linux/coreos-overlay/tree/main/.github/workflows

Thanks!

> 
> But what I'm trying to say is that this PR is pursuing the wrong approach. We have a good way of fetching the latest built kola binaries from CI for a job, and we have github actions to update package ebuilds. Building throwaway SDKs is weird.

The PR is trying to establish means for auditing which version of the OS image was tested with which version of Kola. Currently this information is lost in a Jenkins variable. 
I realise there are gaps left even after this PR is merged, and I understand we might want to take a different approach altogether in the end (for instance, via a separate versioned kola container). I would consider this PR a first step to iterate on - the containerised builds currently run in parallel and aren't used for production releases, giving us room to experiment.

ci-automation/util/update_cros_workon.sh
```diff
@@ -0,0 +1,79 @@
...
+        release="$(echo "${softlink}" | sed -n 's/.*-[0-9.]\+[-.]r\([0-9]\+\)\.ebuild/\1/p')"
+        new_release="$((release + 1))"
+
+        ebuild="${name}-9999.ebuild"
+        new_softlink="${name}-${version}-r${new_release}.ebuild"
+        ln -s "${ebuild}" "${new_softlink}"
+
+        local repo="$(eval "$(grep -E '(CROS_WORKON_REPO|CROS_WORKON_PROJECT)' "${ebuild}")"
+                      echo $CROS_WORKON_REPO/$CROS_WORKON_PROJECT)"
+        commit_id="$(git ls-remote "${repo}" "${branch}" | awk "/[[:space:]]${branch}/ {print \$1}")"
```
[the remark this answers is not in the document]
krnowak: Oh, right. My bad. Sorry for the noise.

ci-automation/util/update_cros_workon.sh
t-lo: Added a "push" option, updated CI usage example in PR description.

ci-automation/util/update_cros_workon.sh
```diff
@@ -0,0 +1,86 @@
...
+            category="${category##*/}" # strip everything from the beginning to the last slash
+            git commit -m "${category}/${name}: Update to latest ${branch}"
+
+            if [ "${commit}" = "push" ] ; then
+                git push
+            fi
+        fi
+    )
+
+    if [ "${commit}" = "true" ] then
```
jepio, line 75:
```suggestion
    if [ "${commit}" = "true" ] ; then
```

jepio: Also: `${commit}` can now be `true`, `commit` or `push`, and this logic doesn't work any more.

ci-automation/util/update_cros_workon.sh
jepio: The HEAD is detached so the push is a no-op.

ci-automation/sdk_container_update.sh
jepio: OK, I can now follow this and that there is a way for the user to override the version.

jepio:
> The PR is trying to establish means for auditing which version of the OS image was tested with which version of Kola. Currently this information is lost in a Jenkins variable.

I maintain that this is solving a problem we don't have. This information is not lost, it simply has never been needed. If we ever do need to know which version of kola was used, let's just print the output `kola version` in the test log in case anyone needs to look back.

t-lo:
> I maintain that this is solving a problem we don't have. This information is not lost, it simply has never been needed. If we ever do need to know which version of kola was used, let's just print the output kola version in the test log in case anyone needs to look back.

I get your point but that does not help at all with automation. Overall motivation of the CI containerisation initiative is to bind the _state_ of a build or test to the _version_ that's being built or tested. Each step in the build pipeline can be reproduced just from the repo, because the repo contains all of the _state_ required to build or test. This aims to both reduce external dependencies as well as to drastically simplify the build and test process.

As an example, any given test step of any given version can be reproduced by
```
./checkout <version>
source ci-automation/test.sh
test_run <arch> <image-type>
```
precisely because all build state is implicit and available after `./checkout`. 

I also understand the (current, intermediate) use of the SDK container to implement this for tests has drawbacks and creates friction. Please understand this is only an intermediate solution - something to experiment and to gain experience with. Ultimately the SDK container needs to be replaced with something more lightweight for testing (like the versioned kola test-runner container we've been discussing), however, the SDK is available right now and seems useful enough as a stop-gap solution.

ci-automation/util/update_cros_workon.sh
t-lo: Fixed.

ci-automation/util/update_cros_workon.sh
t-lo: Addressed. Thanks again for catching this!

t-lo:
I'll add an option to `sdk_container_update` to not create a tag and push a commit, so we can iterate on e.g. mantle/kola dev builds without pushing to github.

ci-automation/util/update_cros_workon.sh
```diff
@@ -0,0 +1,111 @@
...
+    # Use a subshell and operate directly in $ebuild_dir
+    (
+        cd "${ebuild_dir}"
+
+        # If we're in detached head mode and are supposed to push later then
+        #  check out the branch the current commit is the tip of.
+        # We do this early so we error out before touching anything.
+        if [ "${commit}" = "push" ] ; then
+
+            local ebuild_branch="$(git -C branch --show-current)"
```
jepio, line 47:
```suggestion
            local ebuild_branch="$(git branch --show-current)"
```

ci-automation/util/update_cros_workon.sh
```diff
@@ -0,0 +1,111 @@
...
+        echo " ('${softlink}' ==> '${new_softlink}'"
+
+        if [[ "${commit}" =~ (commit|push) ]] ; then
+            git add .
+            local category="$(dirname "${PWD}")"
+            category="${category##*/}" # strip everything from the beginning to the last slash
+            git commit -m "${category}/${name}: Update to latest ${branch}"
+
+            if [ "${commit}" = "push" ] ; then
+                git push
```
jepio, line 95: The submodules are configured with https url, so the push will not work because there's no way to pass credentials.

jepio:
> I'll add an option to `sdk_container_update` to not create a tag and push a commit, so we can iterate on e.g. mantle/kola dev builds without pushing to github.

But without pushing the commit, how would i know what version of kola is being used?

t-lo:
> > I'll add an option to `sdk_container_update` to not create a tag and push a commit, so we can iterate on e.g. mantle/kola dev builds without pushing to github.
> 
> But without pushing the commit, how would i know what version of kola is being used?

You wouldn't - this option would allow quick iterations on Mantle/kola + running tests on an image already built when we know in advance we don't want to use the mantle build beyond the current testing, e.g. when doing intermediate builds while iterating on a larger change. Re-reading our conversation I agree intermediate steps don't mandate individual commits to coreos-overlay, just the final feature should.

t-lo:
Really enjoying the discussion and I feel now that this needs a bit more thinking - setting PR back to draft (Active reviewers please keep reviewing new changes though!).

t-lo:
Found a much simpler approach to do the same: https://github.com/flatcar-linux/scripts/pull/258

t-lo:
Will close this PR in favour of #258 if that one should get merged.

t-lo:
Closing in favour of #258 .

PR #242, opened 2022-03-02
ci-automation/tapfile_helper_lib.sh: read test output from file

t-lo:
This change updates the tapfile helper to read test output from a file instead of passig it inline in the SQL INSERT statement.

This fixes an issue with large error output from tests, which breaks `tap_ingest_tapfile()`:
```
    ci-automation/tapfile_helper_lib.sh: line 31: /usr/bin/sqlite3: Argument list too long
```
Error observed with the `cl.toolbox.dnf-install` test, which can generate 8000 lines of output. Fix tested with the same output.

Fix should be cherry-picked to flatcar-3033, flatcar-3066, flatcar-3139, and flatcar-3165.

### DIFF
--- a/ci-automation/tapfile_helper_lib.sh
+++ b/ci-automation/tapfile_helper_lib.sh
@@ -77,13 +77,14 @@ function __db_init() {
 # 3: <run>     - re-run iteration
 
 function tap_ingest_tapfile() {
+
     local tapfile="${1}"
     local vendor="${2}"
     local run="${3}"
 
     local result=""
     local test_name=""
-    local error_message=""
+    local error_message_file="$(mktemp)"
     local in_error_message=false
 
     if ! [ -f "${TAPFILE_HELPER_DBNAME}" ] ; then
@@ -91,7 +92,10 @@ function tap_ingest_tapfile() {
     fi
 
     # Wrap all SQL commands in a transaction to speed up INSERTs
+    # We will commit intermediately if there's an error message to insert so the
+    # error message file can be reused.
     local SQL="BEGIN TRANSACTION;"
+    local has_error_message="false"
 
     # Example TAP input:
     # ok - coreos.auth.verify
@@ -112,10 +116,12 @@ function tap_ingest_tapfile() {
         if $in_error_message ; then
             if [ "${line}" = "..." ] ; then
                 in_error_message=false
+                has_error_message="true"
             else
-                error_message="$(echo -e "$line" \
-                                    | sed -e 's/^Error: "--- FAIL: /"/' -e 's/^[[:space:]]*//' \
-                                    | sed -e "s/[>\"']/_/g" -e 's/[[:space:]]/ /g')"
+                echo -e "$line" \
+                    | sed -e 's/^Error: "--- FAIL: /"/' -e 's/^[[:space:]]*//' \
+                    | sed -e "s/[>\"']/_/g" -e 's/[[:space:]]/ /g' \
+                    >> "${error_message_file}"
                 continue
             fi
         else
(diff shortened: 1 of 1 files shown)


ci-automation/tapfile_helper_lib.sh
```diff
@@ -77,21 +77,25 @@ function __db_init() {
...
-    local error_message=""
+    local error_message_file="$(mktemp)"
     local in_error_message=false
 
     if ! [ -f "${TAPFILE_HELPER_DBNAME}" ] ; then
        __db_init
     fi
 
     # Wrap all SQL commands in a transaction to speed up INSERTs
+    # We will commit intermediately if there's an error message to insert so the
```
krnowak, line 95: Immediately? Intermittently? I think you meant the former. :)

ci-automation/tapfile_helper_lib.sh
```diff
@@ -128,18 +134,32 @@ function tap_ingest_tapfile() {
...
 
         SQL="${SQL}INSERT OR REPLACE INTO test_run(run,result,output,case_id,vendor_id)
-                             VALUES ('${run}','${result}', '${error_message}',
+                             VALUES ('${run}','${result}', readfile('${test_output}'),
                                      (SELECT id FROM test_case WHERE name='${test_name}'),
                                      (SELECT id FROM vendor WHERE name='${vendor}'));"
-        error_message=""
+
+        if [ "${has_error_message}" = "true" ] ; then
+            SQL="${SQL}COMMIT;"
```
krnowak, line 150: Nitpick: lines like `STUFF="${STUFF}${MORESTUFF}"` could be written like `STUFF+="${MORESTUFF}"` too.

ci-automation/tapfile_helper_lib.sh
```diff
@@ -77,21 +77,25 @@ function __db_init() {
...
 function tap_ingest_tapfile() {
+
     local tapfile="${1}"
     local vendor="${2}"
     local run="${3}"
 
     local result=""
     local test_name=""
-    local error_message=""
+    local error_message_file="$(mktemp)"
```
krnowak, line 87: The only thing I was wondering about here is whether we should run this function in a subshell and use a trap to remove the temporary file, so we don't litter the tmp filesystem in case of some errors.

krnowak (commented):
Looks ok, some nitpicks only.

ci-automation/tapfile_helper_lib.sh
krnowak: Actually, just googled "intermediate" and this does not seem to be a wrong word actually. TIL.

ci-automation/tapfile_helper_lib.sh
t-lo: Making a mental note!

ci-automation/tapfile_helper_lib.sh
t-lo: That's not a bad idea. Will add.

t-lo:
@krnowak Subshell added, and also added a forced line-break of the error / debug output every 200 characters. I found an output line almost 7000 chars long...

ci-automation/tapfile_helper_lib.sh
```diff
@@ -77,69 +77,92 @@ function __db_init() {
...
+
+            if $in_error_message ; then
+                if [ "${line}" = "..." ] ; then
+                    in_error_message=false
+                    has_error_message="true"
+                else
+                    echo -e "$line" \
+                        | sed -e 's/^Error: "--- FAIL: /"/' -e 's/^[[:space:]]*//' \
+                              -e "s/[>\"']/_/g" -e 's/[[:space:]]/ /g' \
+                              -e 's/.\{200\}/&\n/g' \
```
krnowak, line 128: Maybe replacing literal `\n` with newlines could also help with overlong lines - those seem to exist even in the example tap output above.

```suggestion
                              -e "s/[>\"']/_/g" -e 's/[[:space:]]/ /g' \
                              -e 's/\\n/\n/g' \
                              -e 's/.\{200\}/&\n/g' \
```

t-lo:
Cherry-picked to flatcar-3033, flatcar-3066, flatcar-3139, and flatcar-3165.

PR #243, opened 2022-03-03
ci-automation: fix tap parser errors

t-lo:
This change fixes and adds more string chars escaping in the test error debug output (`\` are removed and a bug in removing `"` is fixed), addressing a parser errof the CI encountered when ingesting TAP output.

Furthermore, line numbering is shortened, and test names have a spurious `-` prefix removed.


jepio (approved):

t-lo:
Cherry-picked for 3033, 3139, 3165 (forgot to do that on Friday).

PR #244, opened 2022-03-03
Fix and improve channel handling

pothos:
- jenkins/images: find channel from base channel variable  
    Using the tags on the branch is not enough to find the channel we want
    to be the dev build be related to.
    Use the base channel variable which was introduced for this.
- ci-automation: set the channel from the git tag
    For now we had only "developer" images in the new pipeline.
    Based on the git tag like "alpha-1234.0.0" set the channel (group) for
    the image and also use this logic when finding the channel in the QEMU
    update test.

## How to use

Port to all channels

The first change fixes the package-diff comparison that sometimes failed (for dev builds and for release builds). The second is for being able to build a release image in the new pipeline (a first step towards making releases from it).

## Testing done

[ongoing](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5023/cldsv/) for jenkins/images.sh

[ongoing](http://jenkins.infra.kinvolk.io:8080/job/_testing_/job/container-test/job/packages/945/cldsv/) for ci-automation group setting and qemu update test

### DIFF
--- a/ci-automation/image.sh
+++ b/ci-automation/image.sh
@@ -37,6 +37,9 @@ set -eu
 function image_build() {
     local arch="$1"
 
+    source sdk_lib/sdk_container_common.sh
+    local channel=""
+    channel="$(get_git_channel)"
     source ci-automation/ci_automation_common.sh
     init_submodules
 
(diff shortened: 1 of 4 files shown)


t-lo (approved):
LGTM, thank you!

pothos:
I failed to rerun the test with the updated branch here and would have to rebuild everything again, will just merge

PR #246, opened 2022-03-09
automation/test.sh: pass PARALLEL_TESTS to container

t-lo:
Pass PARALLEL_TESTS to the SDK container if it is set before running tests. Since the tests are started in the container, we need to make available the variable inside the container.

Needs to be cherry-picked to flatcar-3033, flatcar-3139, and flatcar-3165.

### DIFF
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -137,7 +137,7 @@ function test_run() {
     __prepare_torcx "${arch}" "${vernum}" "${work_dir}"
 
     # Pass PARALLEL_TESTS to the container
-    if [ -n "${PARALLEL_TESTS}" ] ; then
+    if [ -n "${PARALLEL_TESTS-}" ] ; then
         echo "PARALLEL_TESTS=\"${PARALLEL_TESTS}\"" > sdk_container/.env
         echo "rm -f 'sdk_container/.env'" >> ./ci-cleanup.sh
     fi
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -38,6 +38,8 @@
 #      All positional arguments after the first 2 (see above) are tests / patterns of tests to run.
 #
 #   MAX_RETRIES. Environment variable. Number of re-runs to overcome transient failures. Defaults to 20.
+#   PARALLEL_TESTS. Environment variable. Number of test cases to run in parallel.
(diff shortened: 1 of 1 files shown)


pothos:
In https://github.com/flatcar-linux/scripts/pull/245 I set it directly as prefix to the command

ci-automation/test.sh
```diff
@@ -134,6 +136,12 @@ function test_run() {
     # Make the torcx artifacts available to test implementation
     __prepare_torcx "${arch}" "${vernum}" "${work_dir}"
 
+    # Pass PARALLEL_TESTS to the container
+    if [ -n "${PARALLEL_TESTS}" ] ; then
+        echo "PARALLEL_TESTS=\"${PARALLEL_TESTS}\"" > sdk_container/.env
```
pothos, line 141: Where does this get sourced?

t-lo: It's in the SDK container user's `.bashrc`, which is being generated during SDK container build: https://github.com/flatcar-linux/scripts/blob/main/sdk_lib/Dockerfile.sdk-import#L47-L48

ci-automation/test.sh
```diff
@@ -134,6 +136,12 @@ function test_run() {
     # Make the torcx artifacts available to test implementation
     __prepare_torcx "${arch}" "${vernum}" "${work_dir}"
 
+    # Pass PARALLEL_TESTS to the container
+    if [ -n "${PARALLEL_TESTS}" ] ; then
```
pothos, line 140:
```suggestion
    if [ -n "${PARALLEL_TESTS-}" ] ; then
```
To gracefully handle if it isn't set

pothos (approved):

t-lo:
Merged + cherry-picked to flatcar-3033, flatcar-3139, and flatcar-3165.

jepio:
So this doesn't work: last nights nightlies are specifying 20 still, even though PARALLEL_TEST is 8. http://jenkins.infra.kinvolk.io:8080/job/_testing_/job/container-test/job/test/246/console
```
+ sudo kola run --board=amd64-usr --parallel=20 --platform=qemu --qemu-bios=/usr/share/qemu/bios-256k.bin --qemu-image=flatcar_production_image.bin --tapfile=results-run-1.tap --torcx-manifest=../torcx_manifest.json '*'
```

PR #247, opened 2022-03-09
ci-automation: set images as official based on version

pothos:
The image needs to be set into official mode through a helper script
(see jenkins/images.sh) and the COREOS_OFFICIAL variable needs to be
set for prod_image_util.sh/build_image_util.sh/grub_install.sh.

## How to use

Pick for alpha, beta, stable

## Testing done

[passed](http://jenkins.infra.kinvolk.io:8080/job/_testing_/job/container-test/job/packages/993/cldsv/) and `/usr/share/update_engine/update-payload-key.pub.pem` is present while `update-payload-key.key.pem` is absent on the image because coreos-au-key got rebuild in the image job with the flag "official"

### DIFF
--- a/ci-automation/image.sh
+++ b/ci-automation/image.sh
@@ -55,10 +55,22 @@ function image_build() {
     local image="flatcar-images-${arch}"
     local image_container="${image}-${docker_vernum}"
 
+    local official_arg=""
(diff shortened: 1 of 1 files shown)


t-lo (approved):
Nice work, thank you Kai!

PR #249, opened 2022-03-10
ci-automation/tapfile_helper_lib.sh: only ASCII chars

t-lo:
This change removes all non-ASCII characters from test debug / error output when ingesting the output for inclusion in the TAP report. Jenkins TAP parser does not handle some unicode chars, leading to tap parser errors with e.g. Cilium output (which uses unicode).

Needs to be cherry-picked to flatcar-3033, flatcar-3139, and flatcar-3165.


pothos (approved):

t-lo:
Merged + cherry-picked to flatcar-3033, flatcar-3139, and flatcar-3165.

PR #250, opened 2022-03-10
jenkins/kola/aws: run tests on additional instance types

pothos:
There was a regression for an instance type which could have been
prevented by testing.
Add the same extended test logic used for Equinix Metal to AWS.

## How to use
Expected to fail due to https://github.com/flatcar-linux/Flatcar/issues/665

Port to all channels

## Testing done
[done](http://jenkins.infra.kinvolk.io:8080/job/os/job/kola/job/aws/1240/console)

### DIFF
--- a/jenkins/kola/aws.sh
+++ b/jenkins/kola/aws.sh
@@ -44,6 +44,32 @@ if [[ "${AWS_AMI_ID}" == "" ]]; then
   echo "Created new AMI ${AWS_AMI_ID} (will be removed after testing)"
 fi
 
+# Run the cl.internet test on multiple machine types only if it should run in general
+cl_internet_included="$(set -o noglob; bin/kola list --platform=aws --filter ${KOLA_TESTS} | { grep cl.internet || true ; } )"
+if [[ "${BOARD}" == "amd64-usr" ]] && [[ "${cl_internet_included}" != ""  ]]; then
+  for INSTANCE in m4.2xlarge; do
+    (
+    OUTPUT=$(timeout --signal=SIGQUIT 6h bin/kola run \
+    --parallel=8 \
+    --basename="${NAME}" \
+    --board="${BOARD}" \
+    --aws-ami="${AWS_AMI_ID}" \
+    --aws-region="${AWS_REGION}" \
+    --aws-type="${INSTANCE}" \
+    --aws-iam-profile="${AWS_IAM_PROFILE}" \
+    --platform=aws \
+    --channel="${GROUP}" \
+    --offering="${OFFER}" \
+    --tapfile="${JOB_NAME##*/}_validate_${INSTANCE}.tap" \
+    --torcx-manifest=torcx_manifest.json \
+    cl.internet 2>&1 || true)
+    echo "=== START $INSTANCE ==="
+    echo "${OUTPUT}" | sed "s/^/${INSTANCE}: /g"
+    echo "=== END $INSTANCE ==="
+    ) &
+  done
+fi
+
 # Do not expand the kola test patterns globs
 set -o noglob
 timeout --signal=SIGQUIT 6h bin/kola run \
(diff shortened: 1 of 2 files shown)


jenkins/kola/packet.sh
```diff
@@ -48,6 +48,7 @@ cl_internet_included="$(set -o noglob; bin/kola list --platform=packet --filter
 if [[ "${BOARD}" == "amd64-usr" ]] && [[ "${cl_internet_included}" != ""  ]]; then
   for INSTANCE in c3.medium.x86 m3.large.x86 s3.xlarge.x86 n2.xlarge.x86; do
     (
+    set +x
```
krnowak, line 51: This is probably a debugging leftover.

jenkins/kola/aws.sh
```diff
@@ -44,6 +44,33 @@ if [[ "${AWS_AMI_ID}" == "" ]]; then
   echo "Created new AMI ${AWS_AMI_ID} (will be removed after testing)"
 fi
 
+# Run the cl.internet test on multiple machine types only if it should run in general
```
krnowak, line 47: Should the `cl.internet` test be filtered out from `KOLA_TESTS` to avoid running it twice?

jenkins/kola/packet.sh
pothos: No, it's the second commit to disable the printing of the OUTPUT="Long output with newlines" statement

krnowak: Oh indeed. I didn't check the commits.

jenkins/kola/aws.sh
pothos: (We want to run the cl.internet service on all instance types)

jenkins/kola/aws.sh
```diff
@@ -44,6 +44,33 @@ if [[ "${AWS_AMI_ID}" == "" ]]; then
...
+    --aws-iam-profile="${AWS_IAM_PROFILE}" \
+    --platform=aws \
+    --channel="${GROUP}" \
+    --offering="${OFFER}" \
+    --tapfile="${JOB_NAME##*/}_validate_${INSTANCE}.tap" \
+    --torcx-manifest=torcx_manifest.json \
+    cl.internet 2>&1 || true)
+    echo "=== START $INSTANCE ==="
+    echo "${OUTPUT}" | sed "s/^/${INSTANCE}: /g"
+    echo "=== END $INSTANCE ==="
```
krnowak, line 69: Would it be less likely to have a mixed output if we did the output with one command? Feel free to ignore it if it does not matter.

```suggestion
    printf '=== START %s ===\n%s\n=== END %s ===\n' "$INSTANCE" "$(echo "${OUTPUT}" | sed "s/^/${INSTANCE}: /g")" "$INSTANCE"
```

jenkins/kola/aws.sh
krnowak: Ok, nevermind, now I have noticed that the extra cl.internet run is done of potentially different instance types.

krnowak (approved):

jenkins/kola/aws.sh
pothos: Yes, a bit less likely but I don't think we hit the case often where two runs exit on exactly the same time. But even then I'm not sure that it helps because two concurrent printfs will also fight with each other, or?

krnowak: I'd expect the submission to some buffer to be done under some lock or whatever, so there would be a chance that printf would do such a submission once. But who knows, let's get this in. If we have an issue with a mangled output, then we will worry about it.

krnowak:
> [t-lo](https://github.com/t-lo) assigned [krnowak](https://github.com/krnowak) [11 minutes ago](https://github.com/flatcar-linux/scripts/pull/250#event-6223793327)

? :)

t-lo:
> > [t-lo](https://github.com/t-lo) assigned [krnowak](https://github.com/krnowak) [11 minutes ago](https://github.com/flatcar-linux/scripts/pull/250#event-6223793327)
> 
> ? :)

Right? :smile: Just back-filling the new tactical board: https://github.com/orgs/flatcar-linux/projects/7/views/1 and wanted some issues / PRs in the "done" column.

PR #252, opened 2022-03-10
set_lsb_release: define Flatcar sysext level

pothos:
Sysext images have a compatibility matching mechanism that searches for
the matching OS version or custom sysext level setting. On Flatcar
there is just the OS version set in /etc/os-release until now which
means that sysext images can't easily be used together with autoupdates
that change the OS version.

Define a sysext level for Flatcar so that users can refer to it instead
of the OS version when they have images that don't rely on a particular
Flatcar version.
Here an example of the now possible metadata:
/etc/extensions/NAME/usr/lib64/extension-release.d/extension-release.NAME
  ID=flatcar
  SYSEXT_LEVEL=1.0
and a symlink `/etc/extensions/NAME/usr/lib→/etc/extensions/NAME/usr/lib64`
In the future the matching logic hopefully gets more flexible because
now it is just a string comparison. Also, the architecture is not
matched either for now - we should work with upstream to improve this.

Closes: https://github.com/flatcar-linux/Flatcar/issues/643

## How to use
Check that the above example works with `systemd-sysext merge` or `systemctl enable --now systemd-sysext`

## Testing done

[started](http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5091/cldsv/)


jepio (approved):

pothos:
> Ouch, when testing I realized that `/usr/lib` on Flatcar is just a symlink and now it gets mounted over and everything breaks^^

works when using an additional symlink in the sysext itself

krnowak:
Just a note - it probably won't be a symlink when we finally switch to the 17.1 profile portage is moaning about forever.

jepio:
its also not a symlink on arm64, so we need to fix this to get sysext ready.

pothos:
> Just a note - it probably won't be a symlink when we finally switch to the 17.1 profile portage is moaning about forever.

Does this mean `lib64` would be the symlink or to have two directories?

We have to solve this first or the sysext images get breaking changes later…

krnowak:
> > Just a note - it probably won't be a symlink when we finally switch to the 17.1 profile portage is moaning about forever.
> 
> Does this mean `lib64` would be the symlink or to have two directories?

From [news page about the profile](https://www.gentoo.org/support/news-items/2019-06-05-amd64-17-1-profiles-are-now-stable.html):

```
In the new profiles, the lib->lib64 compatibility symlink is removed.
64-bit libraries need to be installed directly to lib64.  /lib
and /usr/lib become real directories, that are used for cross-arch
and native non-library packages (gcc, clang) and 32-bit libraries
on the multilib profile (which improves compatibility with prebuilt x86
packages).
```

So IIUC that means that `/lib` won't be a symlink to `/lib64`, and `/usr/lib` won't be a symlink to `/usr/lib64`. So it has nothing to do with `/lib` actually pointing to `/usr/lib` or `/lib64` to `/usr/lib64` - that's a `split-usr` stuff, that will also need a look at, because I think I saw some message in systemd that it's going to be deprecated and removed…

> 
> We have to solve this first or the sysext images get breaking changes later…

PR #253, opened 2022-03-14
version.txt: Fix SDK version (to 3165.0.0)

t-lo:
Release automation does not yet update the SDK version on release. It remains stuck on the last nightly version of the main branch when a new Alpha is released. This change updates the SDK version manually; the root cause needs to be addressed in the release script helpers.


pothos:
I don't see the `alpha-3165.0.0` commit on this branch that updates this file, it should be pushed: https://github.com/flatcar-linux/flatcar-build-scripts/blob/master/tag-release#L72

pothos:
Ah, maybe because the `push_to_branch` argument is not set

pothos:
Can you cherry pick the release commit to include the submodule updates?

pothos:
> Ah, maybe because the `push_to_branch` argument is not set

Yeah, it uses the `PUSH` env var which got replaced by the argument.

pothos:
PR here: https://github.com/flatcar-linux/flatcar-build-scripts/pull/105

t-lo:
> Can you cherry pick the release commit to include the submodule updates?
Will do.

t-lo:
Hmmm, the `alpha-3165.0.0` release tag is not part of the `flatcar-3165` branch... https://github.com/flatcar-linux/scripts/tree/alpha-3165.0.0
I've updated coreos-overlay to match the `alpha-3165.0.0` commit in this PR.

pothos (approved):

PR #257, opened 2022-03-16
ci-automation/tapfile_helper_lib.sh: remove non-printable ASCII

t-lo:
Jenkins TAP file parser does not process non-printable ASCII characters but bails out. This change removes all ASCII < 0x1F, so non-printable characters are not included in the TAP report.

Fixes
```
   ...
    Caused by: unacceptable character '' (0x1B) special characters are not allowed
   ...
```

### DIFF
--- a/ci-automation/tapfile_helper_lib.sh
+++ b/ci-automation/tapfile_helper_lib.sh
@@ -127,7 +127,7 @@ function tap_ingest_tapfile() {
                         | LC_ALL=C sed -e 's/^Error: "--- FAIL: /"/' -e 's/^[[:space:]]*//' \
                               -e "s/[>\\\"']/_/g" -e 's/[[:space:]]/ /g' \
                               -e 's/.\{200\}/&\n/g' \
-                              -e 's/[^\x00-\x7F]/?/g' \
+                              -e 's/[^\x1F-\x7F]/?/g' \
                         >> "${error_message_file}"
                     continue
                 fi


ci-automation/tapfile_helper_lib.sh
```diff
@@ -127,7 +127,7 @@ function tap_ingest_tapfile() {
                         | LC_ALL=C sed -e 's/^Error: "--- FAIL: /"/' -e 's/^[[:space:]]*//' \
                               -e "s/[>\\\"']/_/g" -e 's/[[:space:]]/ /g' \
                               -e 's/.\{200\}/&\n/g' \
-                              -e 's/[^\x00-\x7F]/?/g' \
+                              -e 's/[^\x1F-\x7F]/?/g' \
```
pothos, line 130: If it gets out of hand we can use `tr -c '[:graph:][:space:]' '?'` instead of `LC_ALL=C` and the byte matching

jepio:
We could do this in kola?

ci-automation/tapfile_helper_lib.sh
pothos: The proposed discarding also includes `\t` which I think is not wanted

t-lo: I'm not taking any chances with the Jenkins TAP parser anymore :) It's my third fix for this issue.

t-lo:
> We could do this in kola?

It's Jenkins TAP parser specific, I'd consider this an ugly work-around. I'd rather not mess with the raw debug output in kola but retain the original output entirely.

ci-automation/tapfile_helper_lib.sh
t-lo: > The proposed discarding also includes `\t` which I think is not wanted

That's already done earlier by `s/[[:space:]]/ /g`; I believe that was necessary, too.

pothos (approved):

t-lo:
Merged and cherry-picked to 3033, 3139, and 3165.

PR #258, opened 2022-03-16
ci-automation/test.sh: use the latest kola from coreos-overlay

t-lo:
Make sure we use the kola version pinned in the submodule instead of the one shipped with the SDK. This allows us to update kola for release / maintenance branches while retaining the version pinning.

### DIFF
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -136,6 +136,13 @@ function test_run() {
     # Make the torcx artifacts available to test implementation
     __prepare_torcx "${arch}" "${vernum}" "${work_dir}"
 
+    # Ensure we run tests with an up-to-date kola.
+    # This covers the case of a mantle/kola commit ID in the ebuild being
+    #  newer than the mantle/kola included in the SDK.
+    ./run_sdk_container -x ./ci-cleanup.sh \
+        -n "${container_name}" -C "${sdk_image}" -v "${vernum}" \
+        sudo emerge mantle
+
     # Pass PARALLEL_TESTS to the container
     if [ -n "${PARALLEL_TESTS-}" ] ; then
         echo "PARALLEL_TESTS=\"${PARALLEL_TESTS}\"" > sdk_container/.env


ci-automation/test.sh
```diff
@@ -136,6 +136,13 @@ function test_run() {
     # Make the torcx artifacts available to test implementation
     __prepare_torcx "${arch}" "${vernum}" "${work_dir}"
 
+    # Ensure we run tests with an up-to-date kola.
+    # This covers the case of a mantle/kola commit ID in the ebuild being
+    #  newer than the mantle/kola included in the SDK.
+    ./run_sdk_container -x ./ci-cleanup.sh \
+        -n "${container_name}" -C "${sdk_image}" -v "${vernum}" \
+        sudo emerge mantle
```
pothos, line 144: Would update_chroot also work to only rebuild it if needed?

t-lo: I wouldn't pull it in there - that would impact regular SDK operations (development and build). Rather, I'd publish a new SDK when we have piled up enough changes to justify that.

t-lo: > I wouldn't pull it in there - that would impact regular SDK operations (development and build). 

Some implicit knowledge here I just realised: `update_chroot` is run every time when a new container is run for the first time, to make sure SDK and build target binary hosts are correct.

pothos: I wonder if with this direct call we always rebuild mantle even if we don't have to.

pothos: As far as I know there is no binary package stored inside the SDK nor do we publish them which means that it will rebuild each time in Jenkins.

t-lo: > I wonder if with this direct call we always rebuild mantle even if we don't have to.

Unfortunately yes, and there's no easy way around this because the `CROS_WORKON_COMMIT` we're using does not impact `emerge --update`.  `emerge --update` would update only if the version changed, but a change of `CROS_WORKON_COMMIT` does not count as a version change (even though it arguably _is_).

So currently, we can only rebuild every time.

pothos: I thought there is logic plugged in to emerge already for rebuilding when `CROS_WORKON_COMMIT` changes (an existing remote binary package would still be preferred, though).

t-lo: If there is then I was unable to find it - one would usually use `--update` for that but it doesn't work.

pothos (approved):

ci-automation/test.sh
pothos: Let's just rebuild then, I think the overhead should not be that big compared to the SDK container import itself

t-lo:
Merged and cherry-picked to 3033, 3139, 3165.

PR #265, opened 2022-03-28
jenkins/kola/do: use gzipped image file

jepio:
# jenkins/kola/do: use gzipped image file

The bzipped image file sometimes results in a timeout during decompression on
the remote side. Gzipped images are supposed to be better supported by
DigitalOcean, so use those.

## How to use

## Testing done

### DIFF
--- a/jenkins/kola/do.sh
+++ b/jenkins/kola/do.sh
@@ -9,11 +9,11 @@ set -o pipefail
 if [[ "${DOWNLOAD_ROOT}" == gs://flatcar-jenkins-private/* ]]; then
   echo "Fetching google/cloud-sdk"
   docker pull google/cloud-sdk > /dev/null
-  BUCKET_PATH="${DOWNLOAD_ROOT}/boards/${BOARD}/${FLATCAR_VERSION}/flatcar_production_digitalocean_image.bin.bz2"
+  BUCKET_PATH="${DOWNLOAD_ROOT}/boards/${BOARD}/${FLATCAR_VERSION}/flatcar_production_digitalocean_image.bin.gz"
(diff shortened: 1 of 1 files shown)


pothos (approved):
(for Alpha/main)

pothos:
Maybe we should backport the change now to align the build system in all channels. So far we had no problems coming up from the change, and misalignment always causes inconsistencies (e.g., the gzip change gets documented but LTS doesn't have it) and cherry-pick troubles later.

jepio:
The changes to compression handling? Yeah, we can do that. The reason this PR is not merged is because when I started testing it, the import still failed with a timeout several times in a row. So I lost some faith in whether bzip/gzip makes a real difference. Had to bump up the timeout instead (https://github.com/flatcar-linux/mantle/commit/d48d25de1cd3ba932ca63e7ef1f03a260f524762).

pothos:
I think it was even a problem without any timeout constraints but maybe it doesn't happen often anymore because the hardware running the decompression job got upgraded as well or has less load than before, or lbzip2 is used there now…

I'm still for merging but think that if we do so we should first cherry-pick everywhere.

krnowak:
Merging this as is now is pointless as this updates scripts in jenkins directory, which are rather unused.

pothos:
Ah, we now need to change `DIGITALOCEAN_IMAGE_URL_TEMPLATE` in `ci-automation/ci-config.env`

jepio:
I'm still not convinced this helps. When testing this I found that there were still occasional timeouts during the upload, same as when using bzip2. If no one objects I'd rather close.

PR #267, opened 2022-04-01
ci-container/test: add equinix-metal test script

tormath1:

<hr>

Closes: https://github.com/flatcar-linux/Flatcar/issues/671

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/vms/125/cldsv/

### DIFF
--- a/ci-automation/ci-config.env
+++ b/ci-automation/ci-config.env
@@ -53,3 +53,27 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+
+# -- Equinix Metal --
+EQUINIXMETAL_PARALLEL="${PARALLEL_TESTS:-4}"
+# Metro is a set of Equinix Metal regions
+EQUINIXMETAL_amd64_METRO="${EQUINIXMETAL_amd64_METRO:-SV}"
+EQUINIXMETAL_arm64_METRO="${EQUINIXMETAL_arm64_METRO:-DA}"
+# Name of the Equinix Metal image
+EQUINIXMETAL_IMAGE_NAME="flatcar_production_packet_image.bin.bz2"
+# Storage URL required to store user-data
+EQUINIXMETAL_STORAGE_URL="${EQUINIXMETAL_STORAGE_URL:-gs://flatcar-jenkins/mantle/packet}"
+# Equinix Metal default AMD64 instance type
+EQUINIXMETAL_amd64_INSTANCE_TYPE="${EQUINIXMETAL_amd64_INSTANCE_TYPE:-c3.small.x86}"
+# Space separated list of instance types. On those instances the
+# cl.internet kola test will be run if this test is selected to run.
+EQUINIXMETAL_amd64_MORE_INSTANCE_TYPES="c3.medium.x86 m3.large.x86 s3.xlarge.x86 n2.xlarge.x86"
+# Equinix Metal default ARM64 instance type
+EQUINIXMETAL_arm64_INSTANCE_TYPE="c3.large.arm"
+# Space separated list of instance types. On those instances the
+# cl.internet kola test will be run if this test is selected to run.
+EQUINIXMETAL_arm64_MORE_INSTANCE_TYPES=""
+
+# -- PXE --
+PXE_KERNEL_NAME="flatcar_production_pxe.vmlinuz"
+PXE_IMAGE_NAME="flatcar_production_pxe_image.cpio.gz"
--- a/ci-automation/vendor-testing/equinix_metal.sh
+++ b/ci-automation/vendor-testing/equinix_metal.sh
@@ -1,0 +1,119 @@
+#!/bin/bash
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -euo pipefail
+
+# Test execution script for the Equinix Metal vendor image.
+# This script is supposed to run in the SDK container.
+# This script requires "pxe" Jenkins job.
+
+work_dir="$1"; shift
+arch="$1"; shift
+vernum="$1"; shift
+tapfile="$1"; shift
+
+# $@ now contains tests / test patterns to run
+
+source ci-automation/ci_automation_common.sh
+# required for get_git_channel
+source sdk_lib/sdk_container_common.sh
+
+channel="$(get_git_channel)"
+
+mkdir -p "${work_dir}"
+cd "${work_dir}"
+
+# Equinix Metal ARM server are not yet hourly available in the default `SV` metro
+equinixmetal_metro_var="EQUINIXMETAL_${arch}_METRO"
+equinixmetal_metro="${!equinixmetal_metro_var}"
+
+EQUINIXMETAL_INSTANCE_TYPE_VAR="EQUINIXMETAL_${arch}_INSTANCE_TYPE"
+EQUINIXMETAL_INSTANCE_TYPE="${!EQUINIXMETAL_INSTANCE_TYPE_VAR}"
+MORE_INSTANCE_TYPES_VAR="EQUINIXMETAL_${arch}_MORE_INSTANCE_TYPES"
+MORE_INSTANCE_TYPES=( ${!MORE_INSTANCE_TYPES_VAR} )
+
+# The maximum is 6h coming from the ore GC duration parameter
+timeout=6h
+
+BASE_URL="http://${BUILDCACHE_SERVER}/images/${arch}/${vernum}"
+
+run_equinix_metal_kola_test() {
+    local instance_type="${1}"
+    local instance_tapfile="${2}"
+
+    timeout --signal=SIGQUIT "${timeout}" \
+        kola run \
+          --board="${arch}-usr" \
+          --basename="ci-${vernum/+/-}" \
+          --platform=equinixmetal \
+          --tapfile="${instance_tapfile}" \
+          --parallel="${EQUINIXMETAL_PARALLEL}" \
+          --torcx-manifest=../torcx_manifest.json \
+          --equinixmetal-image-url="${BASE_URL}/${EQUINIXMETAL_IMAGE_NAME}" \
+          --equinixmetal-installer-image-kernel-url="${BASE_URL}/${PXE_KERNEL_NAME}" \
+          --equinixmetal-installer-image-cpio-url="${BASE_URL}/${PXE_IMAGE_NAME}" \
+          --equinixmetal-metro="${equinixmetal_metro}" \
+          --equinixmetal-plan="${instance_type}" \
+          --equinixmetal-project="${EQUINIXMETAL_PROJECT}" \
+          --equinixmetal-storage-url="${EQUINIXMETAL_STORAGE_URL}" \
+          --gce-json-key=<(set +x; echo "${GCP_JSON_KEY}" | base64 --decode) \
+          --equinixmetal-api-key="${EQUINIXMETAL_KEY}" \
+          "${@}"
+
+    # compare the tested instance with the default instance type.
+    if [[ "${instance_type}" != "${EQUINIXMETAL_INSTANCE_TYPE}" ]]; then
+        sed --in-place "s/cl\.internet/${instance_type}\.cl\.internet/" "${instance_tapfile}"
+    fi
+}
+
+cl_internet_included="$(kola list --platform=equinixmetal --filter "${@}" | { grep cl.internet || : ; } )"
+
+# in case of rerun, we need to convert <instance-type>.cl.internet
+# to regular cl.internet tests on the correct instance type.
+instance_types=()
+for t in "${@}"; do
+    if [[ "${t}" =~ ".cl.internet" ]]; then
+        instance_types+=( "${t/\.cl\.internet/}" )
+        # cl_internet needs to run.
+        cl_internet_included="yes"
+    fi
+done
+# Remove any <instance-type>.cl.internet in ${@}
+set -o noglob
+set -- $(echo "$*" | sed 's/[^[:space:]]*\.cl\.internet//g')
+set +o noglob
+
+# empty array is seen as unbound variable.
+set +u
+[[ "${#instance_types}" -gt 0 ]] && MORE_INSTANCE_TYPES=( "${instance_types[@]}" )
+set -u
+
+run_more_tests=0
+
+[[ -n "${cl_internet_included}" ]] && [[ "${#MORE_INSTANCE_TYPES[@]}" -gt 0 ]] && run_more_tests=1
+
+if [[ "${run_more_tests}" -eq 1 ]]; then
+    for instance_type in "${MORE_INSTANCE_TYPES[@]}"; do
+	(
+            OUTPUT=$(set +x; run_equinix_metal_kola_test "${instance_type}" "validate_${instance_type}.tap" 'cl.internet' 2>&1 || :)
+            echo "=== START ${instance_type} ==="
+            echo "${OUTPUT}" | sed "s/^/${instance_type}: /g"
+            echo "=== END ${instance_type} ==="
+        ) &
+    done
+fi
+
+# Skip regular run if only <instance-type>.cl.internet were to be tested
+ARGS="$*"
+if [[ -n "${ARGS// }" ]]; then
+    set -x
+    run_equinix_metal_kola_test "${EQUINIXMETAL_INSTANCE_TYPE}" "${tapfile}" "${@}"
+    set +x
+fi
+
+if [[ "${run_more_tests}" -eq 1 ]]; then
+    wait
+    cat validate_*.tap >>"${tapfile}"
+fi


ci-automation/vendor-testing/equinixmetal.sh
```diff
@@ -0,0 +1,61 @@
...
+set -o noglob
+
+sudo kola run \
+    --board="${arch}-usr" \
+    --basename="ci-${vernum}" \
+    --parallel="${EQUINIXMETAL_PARALLEL}" \
+    --platform=equinixmetal \
+    --tapfile="${tapfile}" \
+    --torcx-manifest=../torcx_manifest.json \
+    --equinixmetal-image-url="https://bucket.release.flatcar-linux.net/flatcar-jenkins/${channel}/boards/${arch}-usr/${vernum}/flatcar_production_packet_image.bin.bz2" \
```
pothos, line 51: This should be the bincache URL, don't know if we have a helper function already?

ci-automation/vendor-testing/equinixmetal.sh
```diff
@@ -0,0 +1,61 @@
...
+# have that many free ARM servers, the whole tests will fail. With PARALLEL_TESTS=2
+# the total number of servers stays < 10.
+# In addition, we override `timeout` to 10 hours, because it takes more than 8 hours
+# to run all tests only with 2 tests in parallel.
+if [[ "${arch}" == "arm64" ]]; then
+  EQUINIXMETAL_METRO="DA"
+  EQUINIXMETAL_PARALLEL="2"
+  timeout=15h
+fi
+
```
pothos, line 40: Do we want to continue testing the other instance types in this file as well as done in the `for INSTANCE` loop or do you want to have a different file for that?

ci-automation/vendor-testing/equinixmetal.sh
sayanchowdhury: There is one `copy_from_buildcache` method used in the qemu.sh

ci-automation/vendor-testing/equinixmetal.sh
```diff
@@ -0,0 +1,61 @@
...
+set -x
+set -o noglob
+
+sudo kola run \
```
sayanchowdhury, line 44: Do we need to include the `cl.internet` test on the other hardware? cc @pothos

ci-automation/vendor-testing/equinixmetal.sh
```diff
@@ -0,0 +1,61 @@
...
+    --torcx-manifest=../torcx_manifest.json \
+    --equinixmetal-image-url="https://bucket.release.flatcar-linux.net/flatcar-jenkins/${channel}/boards/${arch}-usr/${vernum}/flatcar_production_packet_image.bin.bz2" \
+    --equinixmetal-installer-image-base-url="https://bucket.release.flatcar-linux.net/flatcar-jenkins/${channel}/boards/${arch}-usr/${vernum}" \
+    --equinixmetal-project="${EQUINIXMETAL_PROJECT}" \
+    --equinixmetal-storage-url=gs://flatcar-jenkins/mantle/packet \
+    --equinixmetal-metro="${EQUINIXMETAL_METRO}" \
+    --gce-json-key="${GCP_JSON_KEY}" \
+    --equinixmetal-api-key="${EQUINIXMETAL_KEY}"
+    $@
+
```
pothos, line 59: Since kola creates resources which it may no be able to clean up due to errors, we should extend `ci-automation/garbage_collect.sh` to do the `ore packet gc --duration 2h` call.

ci-automation/ci-config.env
```diff
@@ -45,3 +45,9 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+
+# -- Equinix Metal --
+
+EQUINIXMETAL_PARALLEL="${QEMU_PARALLEL}"
```
sayanchowdhury, line 51: For Azure, I've been using `QEMU_PARALLEL` too. Wondering if we should move out into a new var and use it?!

ci-automation/ci-config.env
```diff
@@ -45,3 +45,9 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
...
+# Metro is a set of Equinix Metal regions
+EQUINIXMETAL_METRO="DA"
```
pothos, line 53:
```suggestion
EQUINIXMETAL_METRO="SV"
```
to be near the bincache host

ci-automation/ci-config.env
pothos, line 51:
```suggestion
EQUINIXMETAL_PARALLEL="${${PARALLEL_TESTS:-4}}"
```

ci-automation/vendor-testing/equinixmetal.sh
pothos, line 44:
```suggestion
sudo timeout --signal=SIGQUIT "${timeout}" kola run \
```

ci-automation/vendor-testing/equinixmetal.sh
```diff
@@ -0,0 +1,61 @@
...
+# tests to provision >= 12 ARM servers at the same time. As the DA metro does not
+# have that many free ARM servers, the whole tests will fail. With PARALLEL_TESTS=2
+# the total number of servers stays < 10.
+# In addition, we override `timeout` to 10 hours, because it takes more than 8 hours
+# to run all tests only with 2 tests in parallel.
+if [[ "${arch}" == "arm64" ]]; then
+  EQUINIXMETAL_METRO="DA"
+  EQUINIXMETAL_PARALLEL="2"
+  timeout=15h
+fi
```
pothos, line 39:
```suggestion
else
  timeout=8h
fi
```
(the current default, not sure if it still is the best value)

ci-automation/vendor-testing/equinixmetal.sh
pothos: Yes, we need a new one to only give the URL

pothos:
More a note on later usage from Jenkins in a groovy job: I think we can use a lockable resource like done for VMware ESX to prevent running too many tests in parallel that exhaust the capacities.

ci-automation/vendor-testing/equinixmetal.sh
t-lo: This should work: `"https://${BUILDCACHE_SERVER}/images/${arch}/${vernum}/${QEMU_IMAGE_NAME}"`

ci-automation/ci-config.env
t-lo: Actually, `PARALLEL_TESTS` would be better (with a sane default if that variable is not set or empty), just like the qemu test does it: https://github.com/flatcar-linux/scripts/blob/main/ci-automation/ci-config.env#L47

t-lo: (e.g. `EQUINIXMETAL_PARALLEL="${PARALLEL_TESTS:-20}"` if "20" is a good parallel option for EM)

pothos:
You'll have to rebase to make use of the new container image

ci-automation/vendor-testing/equinixmetal.sh
pothos: see discussion (and implementation draft) in https://github.com/flatcar-linux/scripts/pull/273/files#r855301810 for testing different instance types

ci-automation/garbage_collect.sh
```diff
@@ -132,4 +132,12 @@ function garbage_collect() {
             echo "## (DRY_RUN=y so not doing anything) ##"
         fi
     done
+
+    ## Let's clear the cloud provider resources created during the tests.
+    if [ "$dry_run" != "y" ] ; then
+        vendors=($(find ci-automation/vendor-testing -not -name "qemu*" -type f -printf "%f "))
+        for vendor in "${vendors[@]}"; do
+            ore "${vendor%.*}" gc --duration 2h
```
pothos, line 140: Now I wonder how that works at the moment; is 2 hours not a bit too short for Equinix Metal tests, now that we have the reuse logic for instances?

Looks good, but the for loop also needs to pass the creds, as far as I know.

ci-automation/ci-config.env
```diff
@@ -53,3 +53,17 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+
+# -- Equinix Metal --
+
+EQUINIXMETAL_PARALLEL="${PARALLEL_TESTS:-4}"
+# Metro is a set of Equinix Metal regions
+EQUINIXMETAL_METRO="SV"
```
krnowak, line 61: What I did in AWS is that I had two variables for each arch, so in this case you could have:

```bash
EQUINIXMETAL_amd64_METRO="SV" # Or rather, `: ${EQUINIXMETAL_amd64_METRO:="SV"}`
EQUINIXMETAL_arm64_METRO="DA" # Or rather, `: ${EQUINIXMETAL_arm64_METRO:="DA"}`
```

Then in the vendor script you could do:
```bash
em_metro_var="EQUINIXMETAL_${arch}_METRO"
em_metro="${!em_metro_var}"
```

ci-automation/garbage_collect.sh
tormath1: Don't you think we could add this logic right _after_ the tests in the test job ?

Once all tests have ran for a provider, we cleanup resources for this very provider with ore - in this context, we still have access to the credentials.

pothos: The gc job should be run periodically. Just running it after the job would mean that over the weekend where no changes are pushed it would not run. It would also mean that it won't run for platforms where we trigger the tests manually and not nightly.

tormath1: I get your points. For the record, I just mentioned moving the garbage collection of the cloud provider resources not the other stuffs that we would keep as they are.

Tests should most of the time be CI only and a test script would:
* create the resources
* consume the resources
* delete the created resources

For manual testing, we should use the exact same scripts so if we garbage collect from the test script directly we should be good.

For manual testing _without_ using the scripts, it should up to the user to delete the created resources IMHO.

pothos: The garbage collection is different from the resource cleanup that kola does itself (or that the job does with `ore` to remove the uploaded image).
The cloud resource GC job should run periodically, for example, in the case where Jenkins job got terminated due to the Jenkins nodes rebooting while jobs were running. The leftover cloud resources would accumulate costs or in distrub execution of other tests because the instance limits are hit.

ci-automation/vendor-testing/equinix_metal.sh
```diff
@@ -0,0 +1,108 @@
...
+run_equinix_metal_kola_test "${EQUINIXMETAL_INSTANCE_TYPE}" "${tapfile}" "${@}"
+
+set +x
+
+if [[ "${run_more_tests}" -eq 1 ]]; then
+    wait
+    cat validate_*.tap >>"${tapfile}"
+fi
+
+rm --force ./gcp-key.json
```
pothos, line 108: The other tests use `trap 'cmd' EXIT` to set cleanup actions up, before kola is invoked

ci-automation/ci-config.env
```diff
@@ -53,3 +53,27 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
...
+EQUINIXMETAL_PARALLEL="${PARALLEL_TESTS:-4}"
+# Metro is a set of Equinix Metal regions
+EQUINIXMETAL_amd64_METRO="SV"
+EQUINIXMETAL_arm64_METRO="DA"
+# Name of the Equinix Metal image
+EQUINIXMETAL_IMAGE_NAME="flatcar_production_packet_image.bin.bz2"
+# Storage URL required to store user-data
+EQUINIXMETAL_STORAGE_URL="gs://flatcar-jenkins/mantle/packet"
+# Equinix Metal default AMD64 instance type
+EQUINIXMETAL_amd64_INSTANCE_TYPE="c3.small.x86"
```
pothos, line 67:
```suggestion
EQUINIXMETAL_amd64_INSTANCE_TYPE="${EQUINIXMETAL_amd64_INSTANCE_TYPE-c3.small.x86}"
```
I think it would be good to be able to overwrite it like we can do now in the original pipeline

ci-automation/ci-config.env
```diff
@@ -53,3 +53,27 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+
+# -- Equinix Metal --
+EQUINIXMETAL_PARALLEL="${PARALLEL_TESTS:-4}"
+# Metro is a set of Equinix Metal regions
+EQUINIXMETAL_amd64_METRO="SV"
+EQUINIXMETAL_arm64_METRO="DA"
```
pothos, line 61:
```suggestion
EQUINIXMETAL_amd64_METRO="${EQUINIXMETAL_amd64_METRO:-SV}"
EQUINIXMETAL_arm64_METRO="${EQUINIXMETAL_arm64_METRO:-DA}"
```

ci-automation/ci-config.env
```diff
@@ -53,3 +53,27 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
...
+# Name of the Equinix Metal image
+EQUINIXMETAL_IMAGE_NAME="flatcar_production_packet_image.bin.bz2"
+# Storage URL required to store user-data
+EQUINIXMETAL_STORAGE_URL="gs://flatcar-jenkins/mantle/packet"
```
pothos, line 65:
```suggestion
EQUINIXMETAL_STORAGE_URL="${EQUINIXMETAL_STORAGE_URL:-gs://flatcar-jenkins/mantle/packet}"
```

ci-automation/vendor-testing/equinix_metal.sh
```diff
@@ -0,0 +1,108 @@
...
+MORE_INSTANCE_TYPES_VAR="EQUINIXMETAL_${arch}_MORE_INSTANCE_TYPES"
+MORE_INSTANCE_TYPES=( ${!MORE_INSTANCE_TYPES_VAR} )
+
+# We override `EQUINIXMETAL_PARALLEL`, because running kola with EQUINIXMETAL_PARALLEL >= 4 causes the
+# tests to provision >= 12 ARM servers at the same time. As the DA metro does not
+# have that many free ARM servers, the whole tests will fail.
+# In addition, we override `timeout` to 10 hours, because it takes more than 8 hours
+# to run all tests only with 2 tests in parallel.
+if [[ "${arch}" == "arm64" ]]; then
+  EQUINIXMETAL_PARALLEL="2"
```
pothos, line 43:
```suggestion
```
Now covered by the Jenkins job

ci-automation/vendor-testing/equinix_metal.sh
```diff
@@ -0,0 +1,108 @@
...
+
+    # compare the tested instance with the default instance type.
+    if [[ "${instance_type}" != "${EQUINIXMETAL_INSTANCE_TYPE}" ]]; then
+        sed --in-place "s/cl\.internet/${instance_type}\.cl\.internet" "${instance_tapfile}"
+    fi
+}
+
+cl_internet_included="$(kola list --platform=equinixmetal --filter "${@}" | { grep cl.internet || : ; } )"
+run_more_tests=0
+
```
pothos, line 85: The rerun script will pass `${instance_type}.cl.internet` as test name to rerun, but it's not a valid test name and we have to cut the prefix away again.
We can also set `MORE_INSTANCE_TYPES` from the prefixes, otherwise we would rerun the test for all of them.

ci-automation/vendor-testing/equinix_metal.sh
```diff
@@ -0,0 +1,118 @@
...
+          --parallel="${EQUINIXMETAL_PARALLEL}" \
+          --torcx-manifest=../torcx_manifest.json \
+          --equinixmetal-image-url="${BASE_URL}/${EQUINIXMETAL_IMAGE_NAME}" \
+          --equinixmetal-installer-image-kernel-url="${BASE_URL}/${PXE_KERNEL_NAME}" \
+          --equinixmetal-installer-image-cpio-url="${BASE_URL}/${PXE_IMAGE_NAME}" \
+          --equinixmetal-metro="${equinixmetal_metro}" \
+          --equinixmetal-plan="${instance_type}" \
+          --equinixmetal-project="${EQUINIXMETAL_PROJECT}" \
+          --equinixmetal-storage-url="${EQUINIXMETAL_STORAGE_URL}" \
+          --gce-json-key="./gcp-key.json" \
```
pothos, line 68:
```suggestion
          --gce-json-key=<(echo "${GCP_JSON_KEY}" | base64 --decode) \
```
This should also work to avoid writing the file

ci-automation/vendor-testing/equinix_metal.sh
pothos: or see other suggestion of using a temporary FD with `<(cmd)`

ci-automation/garbage_collect.sh
pothos: I filed a PR here now: https://github.com/flatcar-linux/scripts/pull/306

You can drop this code because it would anyway not find `ore` (this code directly runs on Flatcar)

ci-automation/vendor-testing/equinix_metal.sh
```diff
@@ -0,0 +1,116 @@
...
+          --parallel="${EQUINIXMETAL_PARALLEL}" \
+          --torcx-manifest=../torcx_manifest.json \
+          --equinixmetal-image-url="${BASE_URL}/${EQUINIXMETAL_IMAGE_NAME}" \
+          --equinixmetal-installer-image-kernel-url="${BASE_URL}/${PXE_KERNEL_NAME}" \
+          --equinixmetal-installer-image-cpio-url="${BASE_URL}/${PXE_IMAGE_NAME}" \
+          --equinixmetal-metro="${equinixmetal_metro}" \
+          --equinixmetal-plan="${instance_type}" \
+          --equinixmetal-project="${EQUINIXMETAL_PROJECT}" \
+          --equinixmetal-storage-url="${EQUINIXMETAL_STORAGE_URL}" \
+          --gce-json-key=<(echo "${GCP_JSON_KEY}" | base64 --decode) \
```
pothos, line 65:
```suggestion
          --gce-json-key=<(set +x; echo "${GCP_JSON_KEY}" | base64 --decode) \
```

ci-automation/vendor-testing/equinix_metal.sh
```diff
@@ -0,0 +1,118 @@
...
+instance_types=()
+for t in "${@}"; do
+    if [[ "${t}" =~ ".cl.internet" ]]; then
+        instance_types+=( "${t/\.cl\.internet/}" )
+        # cl_internet needs to run.
+        cl_internet_included="yes"
+        # TODO: replace <instance-type>.cl.internet with cl.internet in ${@}
+    fi
+done
+
```
pothos, line 88:
```suggestion

# Remove <instance-type>.cl.internet in ${@}
set -o noglob
set -- $(echo "$*" | sed 's/[^[:space:]]*\.cl\.internet//g')
set +o noglob
```
We just need to remove the prefixed cl.internet names and don't need to insert cl.internet

pothos:
The general kola run would also have to be skipped if only an instance-specific test should be rerun.

pothos:
> The general kola run would also have to be skipped if only an instance-specific test should be rerun.

Pushed a fix for that and also moved the resetting of the arguments out of the loop over the arguments which was a bit confusing.

Going forward I think we can either merge it now or try to move the logic into https://github.com/flatcar-linux/scripts/pull/293
I'm more in favor of merging first.

pothos:
I ran another test and the outstanding problems are:

- [ ] when re-running tests only a single test each time gets passed to kola
- [ ] the final TAP report doesn't contain the instance tests, thus a failure is not noticed but visible in the output when searching for `FAIL:`

However, this is no regression compared to the existing Equinix Metal test and we can merge this as is.
Later it will be refactored with the logic from https://github.com/flatcar-linux/scripts/pull/293 / https://github.com/flatcar-linux/scripts/pull/273.

pothos (approved):

PR #274, opened 2022-04-19
ci-automation/azure: Add initial container tests infra for Azure

sayanchowdhury:
WIP PR

To test, I pass the Azure creds via the sdk_container/.env file, and current it fails with the issue

```
2022-04-19T10:10:15Z kola: creating flight for RunTests failed: Uploading blob failed: Parse footer field 'Cookie' failed: Invalid footer cookie data [0 0 0 0 0 0 0 0]
```

### DIFF
--- a/ci-automation/ci-config.env
+++ b/ci-automation/ci-config.env
@@ -121,7 +121,6 @@ AWS_PARALLEL="${PARALLEL_TESTS:-8}"
 # sdk_container/.env
 
 # -- Azure --
-: ${AZURE_HYPER_V_GENERATION:="V2"}
 : ${AZURE_IMAGE_NAME:="flatcar_production_azure_image.vhd"}
 : ${AZURE_MACHINE_SIZE:="Standard_D2s_v4"}
 : ${AZURE_MORE_MACHINE_SIZE:="Standard_D2s_v4"}
--- a/ci-automation/ci-config.env
+++ b/ci-automation/ci-config.env
@@ -119,3 +119,14 @@ VMWARE_ESX_PARALLEL="${PARALLEL_TESTS:-4}"
 AWS_PARALLEL="${PARALLEL_TESTS:-8}"
 # AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY should come from
 # sdk_container/.env
+
+# -- Azure --
+: ${AZURE_HYPER_V_GENERATION:="V2"}
+: ${AZURE_IMAGE_NAME:="flatcar_production_azure_image.vhd"}
+: ${AZURE_MACHINE_SIZE:="Standard_D2s_v4"}
+: ${AZURE_MORE_MACHINE_SIZE:="Standard_D2s_v4"}
+: ${AZURE_USE_GALLERY:=""}
+: ${AZURE_USE_PRIVATE_IPS:=true}
+: ${AZURE_VNET_SUBNET_NAME:="jenkins-vnet-westeurope"}
+AZURE_PARALLEL="${PARALLEL_TESTS:-20}"
+AZURE_LOCATION="${AZURE_LOCATION:-westeurope}"
--- a/ci-automation/vendor-testing/azure.sh
+++ b/ci-automation/vendor-testing/azure.sh
@@ -15,9 +15,10 @@ set -x
 source ci-automation/vendor_test.sh
 
 board="${CIA_ARCH}-usr"
-basename="ci-${CIA_VERNUM//+/-}"
+basename="ci-${CIA_VERNUM//+/-}-${CIA_ARCH}"
 azure_instance_type="${AZURE_MACHINE_SIZE}"
 azure_vnet_subnet_name="jenkins-vnet-${AZURE_LOCATION}"
+IS_AMD64=${CIA_ARCH//arm64/}
 
 azure_profile_config_file=''
 secret_to_file azure_profile_config_file "${AZURE_PROFILE}"
@@ -43,8 +44,12 @@ fi
 set -o noglob
 
 run_kola_tests() {
-    local instance_type="${1}"; shift
     local instance_tapfile="${1}"; shift
+    local hyperv_gen="v2"
+    if [ "${instance_type}" = "v1" ]; then
+        hyperv_gen="v1"
+        instance_type="${azure_instance_type}"
+    fi
 
     timeout --signal=SIGQUIT 20h \
 			kola run \
@@ -60,8 +65,8 @@ run_kola_tests() {
 			--torcx-manifest=${CIA_TORCX_MANIFEST} \
 			--tapfile="${instance_tapfile}" \
 			--azure-size=${instance_type} \
+			--azure-hyper-v-generation=${hyperv_gen} \
 			${AZURE_USE_GALLERY} \
-			${AZURE_HYPER_V_GENERATION:+--azure-hyper-v-generation=${AZURE_HYPER_V_GENERATION}} \
 			${azure_vnet_subnet_name:+--azure-vnet-subnet-name=${azure_vnet_subnet_name}} \
 			${AZURE_USE_PRIVATE_IPS:+--azure-use-private-ips=${AZURE_USE_PRIVATE_IPS}} \
 			"${@}"
@@ -76,7 +81,7 @@ run_kola_tests_on_instances \
     "${azure_instance_type}" \
     "${CIA_TAPFILE}" \
     "${CIA_FIRST_RUN}" \
-    "" \
+    ${IS_AMD64:+v1} \
     '--' \
     '' \
     '--' \
--- a/ci-automation/vendor-testing/azure.sh
+++ b/ci-automation/vendor-testing/azure.sh
@@ -44,6 +44,7 @@ fi
 set -o noglob
 
 run_kola_tests() {
+    local instance_type="${1}"; shift
     local instance_tapfile="${1}"; shift
     local hyperv_gen="v2"
     if [ "${instance_type}" = "v1" ]; then
@@ -52,24 +53,24 @@ run_kola_tests() {
     fi
 
     timeout --signal=SIGQUIT 20h \
-			kola run \
-			--board="${board}" \
-			--basename="${basename}" \
-			--parallel="${AZURE_PARALLEL}" \
-			--offering=basic \
-			--platform=azure \
-			--azure-image-file="${AZURE_IMAGE_NAME}" \
-			--azure-location="${AZURE_LOCATION}" \
-			--azure-profile="${azure_profile_config_file}" \
-			--azure-auth="${azure_auth_config_file}" \
-			--torcx-manifest=${CIA_TORCX_MANIFEST} \
-			--tapfile="${instance_tapfile}" \
-			--azure-size=${instance_type} \
-			--azure-hyper-v-generation=${hyperv_gen} \
-			${AZURE_USE_GALLERY} \
-			${azure_vnet_subnet_name:+--azure-vnet-subnet-name=${azure_vnet_subnet_name}} \
-			${AZURE_USE_PRIVATE_IPS:+--azure-use-private-ips=${AZURE_USE_PRIVATE_IPS}} \
-			"${@}"
+      kola run \
+      --board="${board}" \
+      --basename="${basename}" \
+      --parallel="${AZURE_PARALLEL}" \
+      --offering=basic \
+      --platform=azure \
+      --azure-image-file="${AZURE_IMAGE_NAME}" \
+      --azure-location="${AZURE_LOCATION}" \
+      --azure-profile="${azure_profile_config_file}" \
+      --azure-auth="${azure_auth_config_file}" \
+      --torcx-manifest=${CIA_TORCX_MANIFEST} \
+      --tapfile="${instance_tapfile}" \
+      --azure-size=${instance_type} \
+      --azure-hyper-v-generation=${hyperv_gen} \
+      ${AZURE_USE_GALLERY} \
+      ${azure_vnet_subnet_name:+--azure-vnet-subnet-name=${azure_vnet_subnet_name}} \
+      ${AZURE_USE_PRIVATE_IPS:+--azure-use-private-ips=${AZURE_USE_PRIVATE_IPS}} \
+      "${@}"
 }
 
 query_kola_tests() {
--- a/ci-automation/vendor-testing/azure.sh
+++ b/ci-automation/vendor-testing/azure.sh
@@ -41,7 +41,6 @@ if [[ "${CIA_ARCH}" == "arm64" ]]; then
   AZURE_USE_GALLERY="--azure-use-gallery"
 fi
 
-set -o noglob
 
 run_kola_tests() {
     local instance_type="${1}"; shift
--- a/ci-automation/vendor-testing/azure.sh
+++ b/ci-automation/vendor-testing/azure.sh
@@ -84,6 +84,6 @@ run_kola_tests_on_instances \
     "${CIA_FIRST_RUN}" \
     ${IS_AMD64:+v1} \
     '--' \
-    '' \
+    'cl.internet' \
     '--' \
     "${@}"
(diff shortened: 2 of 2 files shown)


ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,79 @@
...
+
+testscript="$(basename "$0")"
+
+# Fetch the Azure image if not present
+if [ -f "${AZURE_IMAGE_NAME}" ] ; then
+    echo "++++ ${testscript}: Using existing ${work_dir}/${AZURE_IMAGE_NAME} for testing ${vernum} (${arch}) ++++"
+else
+    echo "++++ ${testscript}: downloading ${AZURE_IMAGE_NAME} for ${vernum} (${arch}) ++++"
+    copy_from_buildcache "images/${arch}/${vernum}/${AZURE_IMAGE_NAME}.bz2" .
+    bunzip2 "${work_dir}/${AZURE_IMAGE_NAME}.bz2"
```
pothos, line 34:
```suggestion
    lbunzip2 "${work_dir}/${AZURE_IMAGE_NAME}.bz2"
```
We also use it in `qemu_update.sh`, I think it's ok to require it as dependency

pothos:
You'll have to rebase to make use of the new container image

sayanchowdhury:
Update:
The mentioned issue is resolved. The VHD files are uploaded, and the instance is created but ssh into the box fails.

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,81 @@
...
+    --azure-auth="${AZURE_AUTH_CREDENTIALS}" \
+    --tapfile="${tapfile}" \
+    --torcx-manifest=../torcx_manifest.json \
+    ${AZURE_USE_GALLERY} \
+    ${AZURE_MACHINE_SIZE_OPT} \
+    ${AZURE_HYPER_V_GENERATION:+--azure-hyper-v-generation=${AZURE_HYPER_V_GENERATION}} \
+    ${AZURE_VNET_SUBNET_NAME:+--azure-vnet-subnet-name=${AZURE_VNET_SUBNET_NAME}} \
+    ${AZURE_USE_PRIVATE_IPS:+--azure-use-private-ips=${AZURE_USE_PRIVATE_IPS}} \
+    $@
+
```
pothos, line 79: see https://github.com/flatcar-linux/scripts/pull/273/files#r855301810 for the discussion of having the `cl.internet` test run for different instance types (here it would be hyperv generation instead of instance type unless we know that there is some interesting instance type to test?)

ci-automation/ci-config.env
```diff
@@ -53,3 +53,14 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+
+# -- Azure --
+AZURE_HYPER_V_GENERATION="V1"
+AZURE_IMAGE_NAME="flatcar_production_azure_image.vhd"
+AZURE_LOCATION="westeurope"
```
pothos, line 60: The location should come from the Jenkins env var, which should allocate it from a lock

ci-automation/ci-config.env
```diff
@@ -53,3 +53,14 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
...
+AZURE_MACHINE_SIZE="Standard_D2s_v4"
+AZURE_OFFER="basic"
```
pothos, line 62: Do we need still need to test pro vs not pro or can we drop this?

ci-automation/ci-config.env
```diff
@@ -53,3 +53,14 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+
+# -- Azure --
+AZURE_HYPER_V_GENERATION="V1"
+AZURE_IMAGE_NAME="flatcar_production_azure_image.vhd"
+AZURE_LOCATION="westeurope"
+AZURE_MACHINE_SIZE="Standard_D2s_v4"
```
pothos, line 61: Would be good to be able to overwrite the instance type from the job

jepio:
> Update: The mentioned issue is resolved. The VHD files are uploaded, and the instance is created but ssh into the box fails.

The ssh would fail unless the test is run from the node with the azure label.

ci-automation/ci-config.env
sayanchowdhury: The way I was approaching was to overwrite the `sdk_container/.env` if the Jenkins params are passed. https://github.com/kinvolk/jenkins-os/blob/flatcar-master/container/test.groovy#L63

The ci-config.env would serve as the configuration for basic setup, or running the items locally.

pothos: That won't work because `ci-automation/test.sh` sources `ci-automation/ci_automation_common.sh` which sources `ci-automation/ci-config.env` and will then overwrite the values.

krnowak: Maybe we should write those env vars like:

```bash
: ${AZURE_HYPER_V_GENERATION:="V1"}
: ${AZURE_IMAGE_NAME:="flatcar_production_azure_image.vhd"}
: ${AZURE_LOCATION:="westeurope"}
: ${AZURE_MACHINE_SIZE:="Standard_D2s_v4"}
```

ci-automation/ci-config.env
```diff
@@ -82,3 +82,14 @@ GCE_IMAGE_NAME="flatcar_production_gce.tar.gz"
 GCE_GCS_IMAGE_UPLOAD="gs://flatcar-jenkins/developer/gce-ci"
 GCE_MACHINE_TYPE="${GCE_MACHINE_TYPE:-n1-standard-2}"
 GCE_PARALLEL="${PARALLEL_TESTS:-4}"
+
+# -- Azure --
+AZURE_HYPER_V_GENERATION="V1"
```
pothos, line 87: I would rather like the test to use the logic from https://github.com/flatcar-linux/scripts/pull/273 to run the cl.internet test with v1 and otherwise use v2 only from now on. Or is there any reason to prefer v1?

sayanchowdhury: This is a good suggestion, makes sense to test v2 on all and v1 on the selected few.

sayanchowdhury:
Don't merge. I've rebased the PR on top of the `dongsu/ciauto-aws` branch to pull in helper functions from the branch

pothos:
Maybe we can also test `Standard_D2ds_v5` which has a different NIC and in the past we had the problem that the interface rules didn't match this driver.

ci-automation/vendor_test.sh
```diff
@@ -102,6 +110,276 @@ CIA_TESTSCRIPT="${ciavts_testscript}"
...
+#
+# Typical use:
+#   filter_prefixed_tests tests_to_run "${instance_type}" "${@}"
+#   if [[ "${#tests_to_run[@]}" -gt 0 ]]; then …; fi
+#
+# Parameters:
+# 1 - name of an array variable where the filtering results will be stored
+# 2 - prefix
+# @ - test names
+function filter_prefixed_tests() {
```
pothos, line 151: This function and the one below got fixed in `main`, would be good to rebase before testing

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,76 @@
...
+    --platform=azure \
+    --azure-image-file="${AZURE_IMAGE_NAME}" \
+    --azure-location="${AZURE_LOCATION}" \
+    --azure-profile="${AZURE_PROFILE}" \
+    --azure-auth="${AZURE_AUTH_CREDENTIALS}" \
+    --torcx-manifest=${CIA_TORCX_MANIFEST} \
+    --tapfile="${instance_tapfile}" \
+    --azure-size=${instance_type} \
+    ${AZURE_USE_GALLERY} \
+    ${AZURE_HYPER_V_GENERATION:+--azure-hyper-v-generation=${AZURE_HYPER_V_GENERATION}} \
```
pothos, line 57: you can look at the gce test where "gvnic" as instance type was used to set a kola parameter and we can do the same here for v1 vs v2

ci-automation/vendor-testing/aws.sh
```diff
@@ -0,0 +1,79 @@
...
+
+vmdk='flatcar_production_ami_vmdk_image.vmdk'
+
+if [[ "${AWS_AMI_ID}" == "" ]]; then
+    if [[ -f "${vmdk}" ]]; then
+        echo "++++ ${CIA_TESTSCRIPT}: using existing ${vmdk} for ${CIA_VERNUM} (${CIA_ARCH}) ++++"
+    else
+        echo "++++ ${CIA_TESTSCRIPT}: downloading ${vmdk}.bz2 for ${CIA_VERNUM} (${CIA_ARCH}) ++++"
+        copy_from_buildcache "images/${CIA_ARCH}/${CIA_VERNUM}/${vmdk}.bz2" .
+        lbunzip2 "${vmdk}.bz2"
```
pothos, line 31: Since we have the problem with the disk space, how about making this file sparse by trying something like this:
```
cp --sparse=auto <(lbzcat "${vmdk}.bz2") "${vmdk}"
rm "${vmdk}.bz2"
```

pothos: (I meant proposing this for Azure and this is the AWS file, anyway, Sayan picked it up: https://github.com/flatcar-linux/scripts/pull/274/commits/b9bbda63ee1df1f57a39d523897686a1fe5dd677)

pothos: And funny observation: with btrfs and FS compression this wouldn't have bubbled up even though it is not as efficient as sparse files but zeros compress quite ok.

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,83 @@
...
+
+# Test execution script for the qemu vendor image.
+# This script is supposed to run in the SDK container.
+
+# $@ now contains tests / test patterns to run
+
+source ci-automation/vendor_test.sh
+
+board="${CIA_ARCH}-usr"
+basename="ci-${CIA_VERNUM//+/-}"
```
pothos, line 18:
```suggestion
basename="ci-${CIA_VERNUM//+/-}-${CIA_ARCH}"
```
Since the basename is used per test we should prevent resource name clashes when both architectures get tested.

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,83 @@
...
+query_kola_tests() {
+    shift; # ignore the instance type
+    kola list --platform=azure --filter "${@}"
+}
+
+run_kola_tests_on_instances \
+    "${azure_instance_type}" \
+    "${CIA_TAPFILE}" \
+    "${CIA_FIRST_RUN}" \
+    "" \
```
pothos, line 79: For an MVP we can ignore running for multiple instance types but let's focus on testing both v1 and v2 at least.
```suggestion
    ${IS_AMD64:+v1} \
```
I would assume it makes sense to test on v2 by default and just run the cl.internet test on v1?

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,83 @@
...
+# Test execution script for the qemu vendor image.
+# This script is supposed to run in the SDK container.
+
+# $@ now contains tests / test patterns to run
+
+source ci-automation/vendor_test.sh
+
+board="${CIA_ARCH}-usr"
+basename="ci-${CIA_VERNUM//+/-}"
+azure_instance_type="${AZURE_MACHINE_SIZE}"
```
pothos, line 19:
```suggestion
azure_instance_type="${AZURE_MACHINE_SIZE}"
IS_AMD64=${CIA_ARCH//arm64/}
```

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,83 @@
...
+
+if [[ "${CIA_ARCH}" == "arm64" ]]; then
+  AZURE_USE_GALLERY="--azure-use-gallery"
+fi
+
+set -o noglob
+
+run_kola_tests() {
+    local instance_type="${1}"; shift
+    local instance_tapfile="${1}"; shift
```
pothos, line 47:
```suggestion
    local instance_tapfile="${1}"; shift
    local hyperv_gen="v2"
    if [ "${instance_type}" = "v1" ]; then
        hyperv_gen="v1"
        instance_type="${azure_instance_type}"
    fi
```

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,83 @@
...
+    --platform=azure \
+    --azure-image-file="${AZURE_IMAGE_NAME}" \
+    --azure-location="${AZURE_LOCATION}" \
+    --azure-profile="${azure_profile_config_file}" \
+    --azure-auth="${azure_auth_config_file}" \
+    --torcx-manifest=${CIA_TORCX_MANIFEST} \
+    --tapfile="${instance_tapfile}" \
+    --azure-size=${instance_type} \
+    ${AZURE_USE_GALLERY} \
+    ${AZURE_HYPER_V_GENERATION:+--azure-hyper-v-generation=${AZURE_HYPER_V_GENERATION}} \
```
pothos, line 64:
```suggestion
    --azure-hyper-v-generation=${hyperv_gen} \
```

ci-automation/ci-config.env
```diff
@@ -102,3 +102,30 @@ DIGITALOCEAN_PARALLEL="${PARALLEL_TESTS:-8}"
...
+: ${AWS_arm64_MORE_INSTANCE_TYPES:=""}
+: ${AWS_IAM_PROFILE:="ciauto-test"}
+: ${AWS_REGION:="us-east-1"}
+: ${AWS_AMI_ID:=""}
+AWS_PARALLEL="${PARALLEL_TESTS:-8}"
+# AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY should come from
+# sdk_container/.env
+
+# -- Azure --
+: ${AZURE_HYPER_V_GENERATION:="V2"}
```
pothos, line 124:
```suggestion
```
(see the other suggestion as why this wouldn't be a global setting then)

pothos:
Looking at the test run in http://192.168.42.7:8080/job/container/job/test/774/consoleFull I see that the v1 test didn't get started probably due to `mkdir _kola_temp: file exists`

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,89 @@
...
+    copy_from_buildcache "images/${CIA_ARCH}/${CIA_VERNUM}/${AZURE_IMAGE_NAME}.bz2" .
+    cp --sparse=always <(lbzcat "${AZURE_IMAGE_NAME}.bz2") "${AZURE_IMAGE_NAME}"
+    rm "${AZURE_IMAGE_NAME}.bz2"
+fi
+
+if [[ "${CIA_ARCH}" == "arm64" ]]; then
+  AZURE_USE_GALLERY="--azure-use-gallery"
+fi
+
+set -o noglob
```
pothos, line 44:
```suggestion
```
Not sure why this is here, maybe it made later glob usage like `merge_tap_files "${main_tapfile}" 'instance_'*'_validate.tap'` fail

ci-automation/vendor-testing/azure.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -euo pipefail
+
+set -x
```
pothos, line 8:
```suggestion
```
Was used for debugging

pothos (approved):
After a small fixup commit it works now as expected, I'll use the GitHub _squash and merge_ button as Sayan is on holiday.

pothos:
While cherry-picking the PR for 3033, 3139, 3185 I realized that github didn't squash the commits in the branch but only during merge… hm, sorry, at least something learned^^ For 3227 and 3248 I picked the merged commit ID directly.

PR #275, opened 2022-04-19
Run CI container pipeline kola tests with the new mantle image

pothos:
The SDK container does not exist for arm64 and is quite heavy. We
currently also resort to a unconditional rebuilding of mantle inside
the SDK.
Use the new mantle container image to run kola tests, and pin its
version through a text file that gets updated by GitHub Actions.

## How to use

Port to flatcar-3139, flatcar-3185, flatcar-3200

## Testing done

- GitHub Action PRs were created as expected: https://github.com/pothos/scripts/pulls
- arm64 qemu(_update) tests passed on the arm64 Jenkins build executor
- amd64 qemu(_update) tests passed

↑ not needed

### DIFF
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -118,35 +118,23 @@ function test_run() {
     local docker_vernum
     docker_vernum="$(vernum_to_docker_image_version "${vernum}")"
 
-    # Get SDK from either the registry or import from build cache
-    local sdk_version="${FLATCAR_SDK_VERSION}"
-    local sdk_name="flatcar-sdk-${arch}"
-    local docker_sdk_vernum="$(vernum_to_docker_image_version "${sdk_version}")"
-
-    docker_image_from_registry_or_buildcache "${sdk_name}" "${docker_sdk_vernum}"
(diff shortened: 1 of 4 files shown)


ci-automation/test.sh
```diff
@@ -118,54 +118,43 @@ function test_run() {
...
-    ./run_sdk_container -x ./ci-cleanup.sh \
-        -n "${container_name}" -C "${sdk_image}" -v "${vernum}" \
-        sudo emerge mantle
-
     local tap_merged_summary="results-${image}.tap"
     local tap_merged_detailed="results-${image}-detailed.tap"
     local retry=""
     local success=false
+    # Also prune images to remove old mantle images
+    echo "docker rm -f '${container_name}' ; docker image prune -a -f --filter until=24h" >> ./ci-cleanup.sh
```
jepio, line 137: We have the `docker-prune` service running as a cron job on all nodes that does this:
```
ExecStart=/usr/bin/docker system prune -fa --filter=until=12h
```
so I would leave only the `docker rm -f`

jepio (approved):

pothos:
I'll also backport this to all channels

PR #287, opened 2022-04-21
[WIP] Vendor test for VMware

krnowak:
This is to close https://github.com/flatcar-linux/Flatcar/issues/672.

~Does not work, feel free to pick it up.~ To test:

- Prepare the config file for kola. It's the secret sauce. It looks more or less like the following:

```json
{
        "default": {
                "user": "USERNAME",
                "password": "PASSWORD",
                "server": "IP_ADDRESS",
                "static_ips": NUMBER,
                "first_static_ip": "IP_ADDRESS",
                "first_static_ip_private": "IP_ADDRESS",
                "gateway": "IP_ADDRESS",
                "gateway_private": "IP_ADDRESS",
                "subnet_size": NUMBER
        }
}
```
- Encode the config file with base64 and put it into `sdk_container/.env` (like `export VMWARE_ESX_CREDS=<BASE64_ENCODED_CONFIG>`).
- From `<SCRIPTS_REPO>` directory do:
  - `bash` (this is to avoid losing a bash session and a tab in terminal emulator due to any failure and exit because of `set -e` done in the following step)
  - `. ci-automation/test.sh`
  - `test_run amd64 vmware cl.internet`

### DIFF
--- a/ci-automation/ci-config.env
+++ b/ci-automation/ci-config.env
@@ -95,3 +95,10 @@ GCE_PARALLEL="${PARALLEL_TESTS:-4}"
 DIGITALOCEAN_PARALLEL="${PARALLEL_TESTS:-8}"
 # DIGITALOCEAN_TOKEN_JSON env var is used for credentials, and should
 # come from sdk_container/.env. It must be base64-encoded.
+
+# -- VMware ESX --
+
+: ${VMWARE_ESX_IMAGE_NAME:='flatcar_production_vmware_ova.ova'}
+VMWARE_ESX_PARALLEL="${PARALLEL_TESTS:-4}"
+# VMWARE_ESX_CREDS should come from sdk_container/.env and must be
+# base64-encoded.
--- a/ci-automation/vendor-testing/vmware.sh
+++ b/ci-automation/vendor-testing/vmware.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+# Copyright (c) 2022 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+set -euo pipefail
+
+# Test execution script for the VMware ESX vendor image.
+# This script is supposed to run in the mantle container.
+
+source ci-automation/vendor_test.sh
+
+# We never ran VMware ESX on arm64, so for now fail it as an
+# unsupported option.
+if [[ "${CIA_ARCH}" == "arm64" ]]; then
+    echo "1..1" > "${CIA_TAPFILE}"
+    echo "not ok - all qemu tests" >> "${CIA_TAPFILE}"
+    echo "  ---" >> "${CIA_TAPFILE}"
+    echo "  ERROR: ARM64 tests not supported on VMware ESX." | tee -a "${CIA_TAPFILE}"
+    echo "  ..." >> "${CIA_TAPFILE}"
+    break_retest_cycle
+    exit 1
+fi
+
+# Fetch image if not present.
+if [ -f "${VMWARE_ESX_IMAGE_NAME}" ] ; then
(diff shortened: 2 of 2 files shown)


ci-automation/ci-config.env
```diff
@@ -53,3 +53,8 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+# -- VMware ESX --
+
+ESX_IMAGE_NAME='flatcar_production_vmware_ova.ova'
+ESX_PARALLEL="${PARALLEL_TESTS:-4}"
```
pothos, line 59: with the other PR to set up the creds we now have two places where it's defined, I think since the default here depends on the address setup we could also use `1` here or error out if not defined

sdk_lib/sdk_container_common.sh
```diff
@@ -47,6 +47,10 @@ function yell() {
 # Guess the SDK version from the current git commit.
 #
 function get_git_version() {
+    if [[ -f '.git_version' ]]; then
+        cat '.git_version'
```
pothos, line 51: If that is useful in general, how about an env var instead?

ci-automation/vendor-testing/vmware.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+set -x
+
+# krnowak: This is what the old script in jenkins did. I'm not sure we
+# should be backporting this - it seems to be very specific for our
+# setup.
+#
+# # Delete every VM that is running because we'll use all available spots
+# ore esx --esx-config-file "${ESX_CONFIG_FILE}" remove-vms || :
```
pothos, line 55: The static IP allocation inside kola assumes exclusive usage when `StaticIPs != 0`. We can guard this command here by a check of the json file's value with `jq`.

sdk_lib/sdk_container_common.sh
krnowak: Not sure about usefulness - it's just my setup that does not work well with the script (I use git worktree functionality extensively). Still, it's a hack and my gut feeling is that this should be stored as a file somewhere in the scripts repo. Environment variable can be a bit finicky when involved with a bunch of sudos and dockers along the way.

ci-automation/vendor-testing/vmware.sh
krnowak: How does that play with the already running tests in CI when someone wants to reproduce something by running the tests locally?

pothos: If you want to use it locally, you should also "reserve" the lock under http://192.168.42.7:8080/lockable-resources/ or your VMs will compete with the CI VMs over the same IP address…

The removal ensures that the CI isn't entering a broken state if there are left over VMs from manual usage or a forcefully canceled Jenkins job.

pothos:
for testing it in Jenkins you will need https://github.com/flatcar-linux/scripts/pull/305

krnowak:
Will need https://github.com/flatcar-linux/mantle/pull/333.

pothos:
> Doing a test run: http://192.168.42.7:8080/job/container/job/vms/188

Restarted: http://192.168.42.7:8080/job/container/job/test/489/console

krnowak:
This seems to work like a charm.

ci-automation/ci-config.env
krnowak: Now it is expected that the creds will come through the sdk_container/.env or some other way.

sdk_lib/sdk_container_common.sh
krnowak: It's already a part of boilerplate stuff that is already in main.

ci-automation/vendor-testing/vmware.sh
krnowak: This is done.

pothos (approved):

PR #288, opened 2022-04-21
ci-automation: Initial test script for Digital Ocean

krnowak:
Not tested a lot, yet.

Closes https://github.com/flatcar-linux/Flatcar/issues/669.


ci-automation/vendor-testing/digitalocean.sh
```diff
@@ -0,0 +1,71 @@
...
+# We never ran Digital Ocean on arm64, so for now fail it as an
+# unsupported option.
+if [[ "${arch}" == "arm64" ]]; then
+    echo "1..1" > "${tapfile}"
+    echo "not ok - all digital ocean tests" >> "${tapfile}"
+    echo "  ---" >> "${tapfile}"
+    echo "  ERROR: ARM64 tests not supported on Digital Ocean." | tee -a "${tapfile}"
+    echo "  ..." >> "${tapfile}"
+    exit 1
+fi
```
krnowak, line 33: Would be nice to have a way to tell the parent script to not even try rerunning the tests.

ci-automation/ci-config.env
```diff
@@ -53,3 +53,12 @@ QEMU_BIOS="/usr/share/qemu/bios-256k.bin"
 # Published by vms.sh as part of the qemu vendor build.
 QEMU_UEFI_BIOS="flatcar_production_qemu_uefi_efi_code.fd"
 
+# -- Digital Ocean --
+
+# Use the "@ARCH@", "@CHANNEL@" and "@VERNUM@" placeholders. They will
+# be replaced.
+DO_IMAGE_URL="https://${BUILDCACHE_SERVER}/images/@ARCH@/@VERNUM@/flatcar_production_digitalocean_image.bin.bz2"
+DO_CONFIG_FILE='do-token.json'
```
pothos, line 61: This name should be aligned with the credential parameter of the Jenkins job and use something like `="${NAME-"do-token.json"}"` if you want to provide a default here.

The other values `DO_REGION`, `DO_PARALLEL`, and `DO_MACHINE_SIZE` should also only fallback to defaults defined here and prefer what the Jenkins job sets up as env vars.

ci-automation/vendor-testing/digitalocean.sh
```diff
@@ -0,0 +1,71 @@
...
+
+channel="$(get_git_channel)"
+if [[ "${channel}" = 'developer' ]]; then
+    channel='alpha'
+fi
+image_name="ci-${vernum//+/-}"
+testscript="$(basename "$0")"
+image_url="${DO_IMAGE_URL//@ARCH@/${arch}}"
```
pothos, line 41: The Equinix Metal test needs similar logic, I would say we should add a helper function to get the URL that is used by the copy_from_bincache function (or how it's called)

pothos:
Thanks, not to forget: should also be covered by some new GC job that calls `ore … gc …`

ci-automation/vendor-testing/digitalocean.sh
krnowak: Done.

ci-automation/ci-config.env
krnowak: Done.

krnowak:
This is blocked on #293.

ci-automation/ci_automation_common.sh
```diff
@@ -224,3 +226,75 @@ function docker_image_from_registry_or_buildcache() {
...
+function secret_to_file() {
+    local config_var_name="${1}"; shift
+    local secret="${1}"; shift
+    local tmpfile=$(mktemp)
+    local -n config_ref="${config_var_name}"
+    local fd
+
+    exec {fd}<>"${tmpfile}"
+    rm -f "${tmpfile}"
+    echo "${secret}" >&${fd}
```
pothos, line 297:
```suggestion
    echo "${secret}" | base64 --decode >&${fd}
```
Otherwise it fails with `10:55:08  Error: could not create DigitalOcean client: couldn't read DigitalOcean config: invalid character 'e' looking for beginning of value` in http://192.168.42.7:8080/job/container/job/test/451/console
I think we can assume that the file secrets are base64 encoded because this is the case now with Jenkins. If the need arises to be able to opt out of base64 decoding, we can add it later through an env var in the CI config env file or similar.

krnowak:
Tried running it again in http://192.168.42.7:8080/job/container/job/vms/182/console, but now got `curl: (22) The requested URL returned error: 404` out of blue. I'm not sure what this error is referring to.

pothos:
> Tried running it again in http://192.168.42.7:8080/job/container/job/vms/182/console, but now got `curl: (22) The requested URL returned error: 404` out of blue. I'm not sure what this error is referring to.

The rebased "scripts" branch uses the version.txt that got pushed from the nightly just a few hours ago but it didn't yet complete the image build yet (still ongoing): http://192.168.42.7:8080/job/container/job/image/156/

krnowak:
Looks like images should already be here, let me run the tests again then.

krnowak:
Running again: http://192.168.42.7:8080/job/container/job/vms/185

krnowak:
Tests seem to be running alright.

ci-automation/vendor-testing/digitalocean.sh
krnowak: I'll do it in a separate PR.

krnowak:
Cherry-picked to 3227, 3200, 3185, 3139 and 3033.

PR #293, opened 2022-04-22
ci-automation: Reduce boilerplate in vendor tests

krnowak:
~Move the common setup to the test_vendor.sh script, which will then
source the vendor script.~

Move the common setup to the vendor_test.sh script, which will be
sourced by the vendor scripts.

With this PR I'm trying to fix the following issues:

- Stop repeating the same code in every vendor test (taking the parameters, creating the directory, entering it, calling `basename $0` to get the test script name, and so on). With this PR, the vendor scripts can assume being in the expected directory already and can use the provided variables instead of hardcoding them (like `../torcx-manifest.json`).
- Make channel information reliably available (so it works in my git-worktree based setup too).
- Drop unnecessary noglob settings.
- Drop unnecessary sudo before calling kola.

Tested with qemu. Currently the qemu test fail on my computer for unrelated reasons (kolet returns 1, same on current main).

### DIFF
--- a/ci-automation/vendor-testing/qemu.sh
+++ b/ci-automation/vendor-testing/qemu.sh
@@ -6,63 +6,49 @@
 set -euo pipefail
 
 # Test execution script for the qemu vendor image.
-# This script is supposed to run in the SDK container.
+# This script is supposed to run in the mantle container.
 
-work_dir="$1"; shift
-arch="$1"; shift
-vernum="$1"; shift
-tapfile="$1"; shift
-
-# $@ now contains tests / test patterns to run
-
-source ci-automation/ci_automation_common.sh
-
-mkdir -p "${work_dir}"
-cd "${work_dir}"
-
-testscript="$(basename "$0")"
+source ci-automation/vendor_test.sh
 
 # ARM64 qemu tests only supported on UEFI
-if [ "${arch}" = "arm64" ] && [ "${testscript}" != "qemu_uefi.sh" ] ; then
-    echo "1..1" > "${tapfile}"
-    echo "not ok - all qemu tests" >> "${tapfile}"
-    echo "  ---" >> "${tapfile}"
-    echo "  ERROR: ARM64 tests only supported on qemu_uefi." | tee -a "${tapfile}"
-    echo "  ..." >> "${tapfile}"
+if [ "${CIA_ARCH}" = "arm64" ] && [ "${CIA_TESTSCRIPT}" != "qemu_uefi.sh" ] ; then
+    echo "1..1" > "${CIA_TAPFILE}"
+    echo "not ok - all qemu tests" >> "${CIA_TAPFILE}"
+    echo "  ---" >> "${CIA_TAPFILE}"
+    echo "  ERROR: ARM64 tests only supported on qemu_uefi." | tee -a "${CIA_TAPFILE}"
+    echo "  ..." >> "${CIA_TAPFILE}"
     exit 1
 fi
 
 # Fetch image and BIOS if not present
 if [ -f "${QEMU_IMAGE_NAME}" ] ; then
-    echo "++++ ${testscript}: Using existing ${work_dir}/${QEMU_IMAGE_NAME} for testing ${vernum} (${arch}) ++++"
+    echo "++++ ${CIA_TESTSCRIPT}: Using existing ./${QEMU_IMAGE_NAME} for testing ${CIA_VERNUM} (${CIA_ARCH}) ++++"
 else
-    echo "++++ ${testscript}: downloading ${QEMU_IMAGE_NAME} for ${vernum} (${arch}) ++++"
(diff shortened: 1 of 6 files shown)


ci-automation/vendor-testing/qemu.sh
```diff
@@ -3,66 +3,48 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-set -euo pipefail
-
 # Test execution script for the qemu vendor image.
 # This script is supposed to run in the SDK container.
```
pothos, line 7:
```suggestion
# This script is supposed to run in the mantle container and is sourced from "test_vendor.sh".
```

pothos: Ah, the other files have a similar statement and we can copy it instead of using my suggestion

ci-automation/test_vendor.sh
```diff
@@ -0,0 +1,114 @@
...
+# CIA_ARCH:
+#   Architecture to test. Currently it is either "amd64" or "arm64".
+#
+# CIA_TAPFILE:
+#   Where the TAP reports should be written. Usually just passed to
+#   kola throught the --tapfile parameter.
+#
+# CIA_CHANNEL:
+#   A channel. Either "alpha", "beta", "stable" or "lts". Usually
+#   passed to kola through the --channel parameter.
```
pothos, line 40:
```suggestion
#   A channel. Either "alpha", "beta", "stable" or "lts". Used to
#   find the last release for the update check.
```
Normally there is no channel filtering in kola because we use the filtering by version.

ci-automation/vendor-testing/qemu_update.sh
```diff
@@ -3,76 +3,57 @@
...
-if [ "${ON_CHANNEL}" = "developer" ]; then
-    # For main/dev builds we compare to last alpha release
-    ON_CHANNEL="alpha"
-fi
-if [ "${ON_CHANNEL}" = "lts" ]; then
+if [ "${CIA_CHANNEL}" = "lts" ]; then
     echo "Updating from previous LTS is not supported yet (needs creds), fallback to Stable"
-    ON_CHANNEL="stable"
+    CIA_CHANNEL="stable"
 fi
```
pothos, line 28: I opened another PR that removes this, in case you see git conflicts due to that being merged earlier it's ok remove this block.

ci-automation/test_vendor.sh
```diff
@@ -0,0 +1,114 @@
...
+CIA_CHANNEL="${ciavts_channel}"
+CIA_TESTSCRIPT="${ciavts_testscript}"
+CIA_GIT_VERSION="${ciavts_git_version}"
+CIA_BUILD_TYPE="${ciavts_type}"
+CIA_TORCX_MANIFEST="${ciavts_torcx_manifest}"
+
+# Unset all variables with ciavts_ prefix now.
+unset -v "${!ciavts_@}"
+
+source "${vendor_test}" "${@}"
```
pothos, line 114: Thinking about it, we already have a long chain of scripts that call scripts. Couldn't we let the vendor test source this file here instead?

ci-automation/vendor-testing/qemu.sh
krnowak: I updated the description.

ci-automation/vendor-testing/qemu_update.sh
krnowak: Thanks for the heads-up.

ci-automation/test_vendor.sh
krnowak: Yeah, we could. I made the change.

krnowak:
I renamed `test_vendor.sh` to `vendor_test.sh`, because the file stopped being an executable, so verb form of the filename stopped fitting.

pothos:
Should I do a test run on Jenkins?

krnowak:
> Should I do a test run on Jenkins?

@pothos: Yes, please. I'll check if I can run GCE and equinix metal tests locally.

pothos:
> > Should I do a test run on Jenkins?
> 
> @pothos: Yes, please. I'll check if I can run GCE and equinix metal tests locally.

Here started for the regular `qemu` test script: http://192.168.42.7:8080/job/container/job/test/445/parameters/ - the same works for other test scripts, except when there is no vendor image built because the nightly doesn't build all of them. In this case you start the VMs job and specify the platform name there as vendor image.

pothos:
Started the test for Equinix Metal and GCE: http://192.168.42.7:8080/job/container/job/vms/172/cldsv/

We can wait a bit and see that it gets started, then merge.

pothos (approved):

pothos:
Please cherry-pick to flatcar-3033, 3139, 3185, 3227 (and 3200 while the old Alpha still gets tested as nightly)

krnowak:
Cherry-picked.

PR #299, opened 2022-04-26
jenkins/kola/qemu.sh: create missing folders

pothos:
This ports a similar change from "main" to create the directory for
amd64 if it doesn't exist because mantle is not installed in the SDK.

## Testing done

[started](http://192.168.42.7:8080/job/os/job/manifest/5510/cldsv/)


jepio (approved):

jepio:
We missed the other occurrence in qemu_uefi.sh.

pothos:
> We missed the other occurrence in qemu_uefi.sh.

Thanks, pushed it

PR #321, opened 2022-05-06
update_sdk_container_image: work around sandbox permission errors

t-lo:
This change temporarily disables the Gentoo sandbox when updating the SDK to work around sandbox permission errors some pakage builds (like e.g. GO) run into.

Fixes e.g.
```
Building Go cmd/dist using /usr/lib/go-bootstrap. (go1.5.3 linux/amd64)
 * /var/tmp/portage/sys-apps/sandbox-2.12/work/sandbox-2.12/libsandbox/trace.c:do_peekstr():125: failure (Operation not permitted):
 * ISE:do_peekstr:process_vm_readv(6863, 0x00007ffe4a502180{0x00007f01abd3e010, 0x570}, 1, 0x00007ffe4a502190{0x000000c820012a90, 0x570}, 1, 0) failed: Operation not permitted
 * ERROR: dev-lang/go-1.17.8::coreos failed (compile phase):
```

Needs cherry-picking to flatcar-3033, flatcar-3139, flatcar-3185, flatcar-3200, and flatcar-32227.

### DIFF
--- a/sdk_lib/Dockerfile.sdk-update
+++ b/sdk_lib/Dockerfile.sdk-update
@@ -4,8 +4,18 @@ FROM ${BASE}
 COPY --chown=sdk:sdk sdk_container/ /mnt/host/source
 COPY --chown=sdk:sdk . /mnt/host/source/src/scripts
 
+# Disable all sandboxing for SDK updates since some core packages
+#  (like GO) fail to build from a permission error otherwise.
+RUN cp /home/sdk/.bashrc /home/sdk/.bashrc.bak
+RUN echo 'export FEATURES="-sandbox -usersandbox -ipc-sandbox -network-sandbox -pid-sandbox"' \
+        >> /home/sdk/.bashrc
+
 RUN chown sdk:sdk /mnt/host/source
 RUN /home/sdk/sdk_entry.sh ./update_chroot --toolchain_boards="amd64-usr arm64-usr"
 
 RUN /home/sdk/sdk_entry.sh ./setup_board --board="arm64-usr" --regen_configs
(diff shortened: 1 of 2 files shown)


krnowak (approved):
Looks good, but does this need an issue filed about reverting this change?

t-lo:
> Looks good, but does this need an issue filed about reverting this change?

I don't think so, we haven't used `update_sdk_container_image` in the past. We just started using it so some bumps are expected.

krnowak:
> > Looks good, but does this need an issue filed about reverting this change?
> 
> I don't think so, we haven't used `update_sdk_container_image` in the past. We just started using it so some bumps are expected.

Ok, I was just wondering if there is something about our packages that could be changed to keep the sandbox enabled. I think that we have currently our own way of bootstrapping go (or really outdated), which may be the reason it isn't playing nice with sandbox, compared to what currently Gentoo has.

t-lo:
Gentoo upstream has acknowledged this behaviour and recommends running privileged builds: https://github.com/gentoo/gentoo-docker-images/issues/98#issuecomment-735281173 (this requires `docker buildx` which we don't ship with Flatcar, hence it's unavailable on Jenkins worker nodes).

This is not an issue when running the SDK container since we run it privileged. But when building via `Dockerfile`, the temporary build container runs without necessary privileges. Since we don't want to require `buildx` fir the SDK build, the only way around this seems to be temporarily disabling the sandbox.

t-lo:
Merged and cherry-picked to flatcar-3033, flatcar-3139, flatcar-3185, flatcar-3200, and flatcar-3227.

PR #328, opened 2022-05-09
New major release: 3227.0.0

sayanchowdhury:


krnowak (approved):

pothos:
This PR makes the nightly build use the Alpha SDK, not really what we want and it will be overwritten in the next nightly anyway

pothos:
Found the place in the release instructions that was outdated, will update it in another PR.

PR #331, opened 2022-05-11
ci-automation/vendor-testing/equinix_metal.sh: Use test framework

pothos:
The test framework from the AWS PR allows us to align the logic which
also addresses some bugs we had here.
Port the Equinix Metal test over to the new framework (and also use
different test basenames per architecture while at it which could
otherwise result in clashes).

## How to use

backport

## Testing done

[ongoing](http://192.168.42.7:8080/job/container/job/test/558/consoleFull)
The s3.xlarge.x86 instance fails which is a known issue and cl.locksmith.cluster also fails which I'm not sure we have seen before.
A problem seems to be that the test for n2.xlarge.x86 gets re-run even though it didn't fail.
Maybe the job also hangs at the end, as I see no new output for a while now.←Eventually there was progress, so all good.

### DIFF
--- a/ci-automation/vendor-testing/equinix_metal.sh
+++ b/ci-automation/vendor-testing/equinix_metal.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 
 # Test execution script for the Equinix Metal vendor image.
 # This script is supposed to run in the mantle container.
-# This script requires "pxe" Jenkins job.
+# This script requires the PXE images to be built.
 
 source ci-automation/vendor_test.sh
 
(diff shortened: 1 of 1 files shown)


pothos:
@krnowak A problem seems to be that the test for `n2.xlarge.x86` gets re-run even though it didn't fail.

tormath1 (approved):

krnowak:
I think that the problem with unnecessary rerunning of the passed tests was that I did not reset the `instance_tests` variable back to an empty array. So `cl.internet` tests get picked up for `n2.xlarge.x86` because in earlier iteration `instance_tests` gets it for `s3.xlarge.x86`. Looking at the `EQUINIXMETAL_amd64_MORE_INSTANCE_TYPES` variable, `s3.xlarge.x86` is processed before `n2.xlarge.x86`, but after `c3.medium.x86` and `m3.large.x86`, so that is in line with my observation.

pothos:
Thanks for having a look! I think you can file a new PR for that.

krnowak:
@pothos : https://github.com/flatcar-linux/scripts/pull/332

krnowak (approved):

PR #332, opened 2022-05-11
ci-automation: Do not rerun tests on unrelated instances

krnowak:
We forgot to clear the array with instance tests to rerun, so the list
grew from one iteration to another when going over all the instance
types. I did not spot it before, because I tested it with only one
extra instance.

### DIFF
--- a/ci-automation/vendor_test.sh
+++ b/ci-automation/vendor_test.sh
@@ -369,7 +369,7 @@ function run_kola_tests_on_instances() {
         fi
     done
 
-    local -a main_tests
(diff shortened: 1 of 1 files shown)


pothos (approved):

krnowak:
Tested locally, works. Fixed a potential issue with unbound variable in the meantime.

pothos:
This wasn't cherry-picked, will do now.

pothos:
No problem, it was just for the record of why I pushed something to the branches directly

PR #333, opened 2022-05-23
ci-automation: Push version file early

pothos:
When a nightly build is started that pushes the version file to the
branch it was doing so only at the end of the build, causing the push
to fail if something else got merged in between.
Push the version file early by generating it the same way it would be
generated by the run_sdk_container/bootstrap_sdk_container scripts.
In the case of the SDK the version file gets the same version for the
OS and the SDK and a few commands were removed. Note that the scripts
will still rewrite the file but it should be a no-op.

Fixes https://github.com/flatcar-linux/Flatcar/issues/723

## How to use

Observe that a nightly build pushes the expected values in the version file to the branch at the start of the build. One case is the SDK job, the other the packages job.

## Testing done

TODO

### DIFF
--- a/ci-automation/packages.sh
+++ b/ci-automation/packages.sh
@@ -122,6 +122,13 @@ function packages_build() {
     local docker_vernum="$(vernum_to_docker_image_version "${vernum}")"
     local packages_container="flatcar-packages-${arch}-${docker_vernum}"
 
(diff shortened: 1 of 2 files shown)


pothos:
Cherry-picking for 3033, 3139, 3185, 3227, 3248

PR #334, opened 2022-05-23
ci-automation: Sign the artifacts

krnowak:
For now it relies on the SIGNER environment variable not being empty
to perform the signing.

Comes together with https://github.com/kinvolk/jenkins-os/pull/229.

### DIFF
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -175,7 +175,8 @@ function docker_image_to_buildcache() {
     local tarball="$(basename "$image")-${version}.tar.gz"
 
     $docker save "${image}":"${version}" | $PIGZ -c > "${tarball}"
-    copy_to_buildcache "containers/${version}" "${tarball}"
+    sign_artifacts "${SIGNER:-}" "${tarball}"
+    copy_to_buildcache "containers/${version}" "${tarball}"*
 }
 # --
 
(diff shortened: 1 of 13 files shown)


pothos (commented):
At what point is the generic image signed and copied to the buildcache? (Edit: maybe we can look at https://github.com/flatcar-linux/Flatcar/issues/740 the same time)

ci-automation/ci_automation_common.sh
```diff
@@ -299,3 +300,36 @@ function secret_to_file() {
...
+function sign_artifacts() {
+    local signer="${1}"; shift
+    # rest of the parameters are directories/files to sign
+    local to_sign=()
+    local file
+    local files
+
+    if [[ -z "${signer}" ]]; then
+        return
+    fi
```
pothos, line 314: How about handling `[ "$*" = "*" ]` when there were no files and the glob didn't expand?

pothos: Maybe also for the case where `*` is the last character of the string like `sometarball*` when the caller uses `"${dest_tarball}"*`

krnowak:
> At what point is the generic image signed and copied to the buildcache? (Edit: maybe we can look at [flatcar-linux/Flatcar#740](https://github.com/flatcar-linux/Flatcar/issues/740) the same time)

It looks to me that the "generic image" is basically a tarball made by "docker save". I patched `docker_image_to_buildcache` to sign the tarball before uploading it.

pothos:
Ok, it's probably covered by the vms job at the moment, the copying to buildcache is done in the end and the generic image seems to be copied along with the newly generated images.

jenkins/container/image.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# expects arch to be in environment
+#
+# signing_key and signer may be absent
+
+set -exuo pipefail
+
+source ci-automation/image.sh
```
pothos, line 9: Given that the `ci-automation/image.sh` probably won't be sourced by something else, how about moving all this into it instead of creating new files?

pothos: same for packages/vms/sdk(_container)

krnowak: I'll have a look, but I'll let the CI to finish it's test run first.

krnowak: New files are gone now.

jenkins/container/packages.sh
```diff
@@ -0,0 +1,12 @@
...
+# expects version, arch, coreos_ref, portage_ref to be in environment
+#
+# signing_key and signer may be absent
+
+set -exuo pipefail
+
+source ci-automation/packages.sh
+source jenkins/container/gpg_setup.sh
+
+packages_build "${version}" "${arch}" "${coreos_ref}" "${portage_ref}"
```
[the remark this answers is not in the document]
krnowak: These are gone now.

krnowak:
Test run: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/89

krnowak:
Test run succeeded for arm64, failed for amd64 because we run out of space apparently. Hopefully GC will clean some things up.

pothos (approved):
Thanks

krnowak:
Cherry-picked to 3255 3227 3139 3033.

PR #338, opened 2022-05-31
build_library: fix manually git protocol in Github URI

dongsupark:
Since Github does not allow unauthenticated git protocol in the URI, we need to manually replace `git://` with `https://` in `CROS_WORKON_REPO`.

Without that, the SDK stage1 would simply fail, because at that moment only the ancient version of baselayout, 3.6.8-r2 is available, which does not have even the [URI fix](https://github.com/flatcar-linux/coreos-overlay/commit/97074d28e514).
As a result, `git clone git://github.com/...`  will simply fail.

## Testing done

CI running: http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/5767/

### DIFF
--- a/build_library/catalyst.sh
+++ b/build_library/catalyst.sh
@@ -261,6 +261,18 @@ build_stage() {
         return
     fi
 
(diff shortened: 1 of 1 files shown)


pothos:
Thanks for working on this. We can pick this up again and cover the remaining ebuild files if we want to build a new 3033 SDK in the future.

PR #348, opened 2022-06-16
Make qemu_template.sh work on macOS

martinborgman:
# Make qemu_template.sh work on macOS

Line 14; The nproc command is only available on systems with GNU coreutils installed. The getconf _NPROCESSORS_ONLN alternative will work on a wider range of UNIX systems.
Line 114; The mktemp syntax used is not standaard and will only work with the GNU version of mktemp.
Line 159; added hvf (macOS) and tcg (no acceleration) options as a fallback. By doing this qemu-system-x86_64 will try to use kvm, but when it fails try hvf, and when that fails switch to the tcg accelerator.

## How to use

The proposed change should not change anything on systems where this script already worked, but is should now also work on macOS x86_64 and M1 systems using a recent version of qemu 

## Testing done

Changes tested on MacPro 5.1 running macOS 12.3.1 and QEMU emulator version 6.2.0 and 7.0.0
Also tested on a MacBook Pro 16-inch 2021 running macOS 13.0 Beta and QEMU emulator version 7.0.0

### DIFF
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -11,7 +11,7 @@ VM_MEMORY=
 VM_CDROM=
 VM_PFLASH_RO=
 VM_PFLASH_RW=
-VM_NCPUS="$(nproc)"
+VM_NCPUS="$(getconf _NPROCESSORS_ONLN)"
 SSH_PORT=2222
 SSH_KEYS=""
 CLOUD_CONFIG_FILE=""
@@ -111,7 +111,7 @@ write_ssh_keys() {
 
 
 if [ -z "${CONFIG_IMAGE}" ]; then
-    CONFIG_DRIVE=$(mktemp -t -d flatcar-configdrive.XXXXXXXXXX)
+    CONFIG_DRIVE=$(mktemp -d)
     ret=$?
     if [ "$ret" -ne 0 ] || [ ! -d "$CONFIG_DRIVE" ]; then
         echo "$0: mktemp -d failed!" >&2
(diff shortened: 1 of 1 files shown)


build_library/qemu_template.sh
```diff
@@ -111,7 +111,7 @@ write_ssh_keys() {
 
 
 if [ -z "${CONFIG_IMAGE}" ]; then
-    CONFIG_DRIVE=$(mktemp -t -d flatcar-configdrive.XXXXXXXXXX)
+    CONFIG_DRIVE=$(mktemp -d -t flatcar-configdrive)
```
pothos, line 114: Removing the Xs breaks GNU mktemp. Does this work on MacOS?

```suggestion
    TEMPSUFFIX=""
    if mktemp --version | grep -q GNU; then
      TEMPSUFFIX=".XXXXXXXXXX"
    fi
    CONFIG_DRIVE=$(mktemp -d -t "flatcar-configdrive${TEMPSUFFIX}")
```

martinborgman: Yes it does, but simply adds the random part at the end like "flatcar-configdrive.XXXXXXXXXX.FkrR1lVh".

pothos: With my suggestion or with the way it was before?

jepio: How about we leave the X's and just accept that the way the name is constructed is slightly different. At the end of the day, we only need it to be slightly unique on both platform, no reason to overcomplicate things.

pothos: Ok, updated my suggestion

martinborgman:
Let me first make clear that there is no standard mktemp in UNIX there is only a mktemp() C function and its exact behavior differs between UNIX implementations. One can use these C functions from the shell by passing them to m4 which comes standard on UNIX systems but not on Linux. But none of this fixes our issue here.
Our problem is that the GNU implementation and the macOS freebsd implementation have a different behavior for the -t parameter.
However if we remove the -t parameter and the template, both mktemp versions wil create a directory using the default template tmp.XXXXXXXX in the users temp directory.
If we remove the -t parameter and keep the template, both mktemp versions wil create a directory using the given template in the current directory.
My suggestion is to remove both the -t parameter and the template. The functionality we loose is the flatcar-configdrive part in the directory name.

build_library/qemu_template.sh
martinborgman: I tried to make it even simpler by removing the -t and the template.

pothos (approved):
Thanks, can you squash the commits?

pothos:
This will be part of the next major Alpha release, and transition to Stable as usual but nothing prevents people to use the Alpha script for Stable.

PR #354, opened 2022-06-24
Add CI workflow

gabriel-samfira:
# Add GitHub CI workflow

Currently the flatcar CI is not publicly viewable. this adds some friction for external contributors, as we rely on someone with access to jenkins, to shuttle information from the CI, into the PR.

This change adds a new workflow, that builds packages, the image and VM images, then uploads the resulting artifacts to github.

This is the first step in creating a set of workflows that are on par with what already exists in Jenkins. Next steps will include:

  * Create a workflow that pulls the artifacts that were built, and tests them using Kola
  * Create workflows in dependency repos (portage-stable, coreos-overlay), that will trigger a build in this repository with the proposed changes to the dependencies.

The current "glue" CI scripts were used as a base fr this workflow. Initially, this workflow sourced the existing scripts and called a function within those scripts, however, some changes were needed to work around the fact that there is no need to upload anything to bincache, and changing those existing scripts would have added a number of ugly special cases.

Cherry picking only needed bits helps keep the workflow clean, while allowing us to extend it with any future jobs or steps that might be needed, without working around old behavior. 

## How to use

This workflow will be run automatically on any pull request. To limit automatic workflow runs only to members of the team, change the repo settings under Settings --> Actions --> General --> "Fork pull request workflows from outside collaborators" --> "Require approval for all outside collaborators"

You can also run this workflow manually, by going to Action --> Run build --> Run Workflow. Running the workflow manually, also allows you to customize inputs.

## Testing done

This change was tested in a fork. Results can be seen here:

https://github.com/gabriel-samfira/scripts/actions/runs/2553458796

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,286 @@
+name: "Run build"
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      bincache_server:
+        description: |
+          Bincache server.
+        default: "bincache.flatcar-linux.net"
+        required: true
+      image_formats:
+        description: |
+          Space-separated vendor formats to build.
+        required: true
+        default: qemu_uefi
+      portage_remote:
+        description: |
+          The remote we should pull portage-stable from. This defaults to whatever the submodule is set to in this repo.
+          If triggered by a change in the portage repo, please set this to the remote which is proposing a change.
+        required: false
+      portage_ref:
+        description: |
+          This is the ref we will use to pull the changes from the portage_remote.
+        required: false
+      coreos_remote:
+        description: |
+          The remote we should pull coreos-overlay from. This defaults to whatever the submodule is set to in this repo.
+          If triggered by a change in the portage repo, please set this to the remote which is proposing a change.
+        required: false
+      coreos_ref:
+        description: |
+          This is the ref we will use to pull the changes from the coreos_remote.
+        required: false
+
+permissions: {}
+
+jobs:
+  packages:
+    name: "Build Flatcar packages"
+    runs-on:
+      - debian
+      - build
+      - self-hosted
+      - x64
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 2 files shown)


pothos:
Thanks so much!

> This workflow will be run automatically on any pull request. To limit automatic workflow runs only to members of the team, change the repo settings under Settings --> Actions --> General --> "Fork pull request workflows from outside collaborators" --> "Require approval for all outside collaborators"

I just did that but we can discuss if we relax this for first contributors.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,286 @@
+name: "Run build"
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      bincache_server:
+        descriotion: |
```
pothos, line 8:
```suggestion
        description: |
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,286 @@
...
+                  -v "${vernum}" \
+                  ./image_to_vm.sh --format "${format}" --board="${arch}-usr" \
+                      --from "${CONTAINER_IMAGE_ROOT}/${arch}-usr/latest" \
+                      --image_compression_formats=bz2
+          done
+
+          # copy resulting images
+          ./run_sdk_container -n "${vms_container}" \
+              -v "${vernum}" \
+              cp --reflink=auto -R "${CONTAINER_IMAGE_ROOT}/${arch}-usr/" "./${images_out}"
```
pothos, line 268:
```suggestion
              cp --reflink=auto -R "${CONTAINER_IMAGE_ROOT}/${arch}-usr/" "./${images_out}"
```
In case we have space problems on the runner, would it work to move here?

gabriel-samfira: There should be enough space to run this workflow comfortably if we remove containers once we're done with them. If we need more space on the runners, we can always tweak the flavors.

pothos:
Will backport down to flatcar-3033

PR #355, opened 2022-06-27
ci-automation: Use the package container for VM image building

pothos:
- ci-automation: Use the package container for VM image building  
    The image job builds an image container that is multiple GBs big and
    takes >10 mins to be loaded in the vms job. The vms job can also do its
    work by running from the packages container from the packages job when
    it fetchs the built image from bincache first and assuming the images
    job copies it there.
    Skip generating the image container and instead use the packages
    container for VM image building by copying the image folder first to
    bincache and then retrieving it from there. While reworking this we
    also address the issue that the VMs container had used the same name
    for both architectures, causing a race when both run in parallel on
    the same worker.
- ci-automation: Run package-diff to report image changes    
    The original pipeline has package-diff commands to print out image
    differences compared to the last release. This is used for the release
    Go/No-Go QA checks.
    Add the same logic to the new pipeline.

Closes https://github.com/flatcar-linux/Flatcar/issues/740

## Testing done

[here](http://192.168.42.7:8080/job/container/job/packages/679/cldsv/) and checked that the expected files are in https://bincache.flatcar-linux.net/images/amd64/9999.1.2+noimagecontainer-kai/ and the workspace has both the `latest` folder with the resulting images and the `latest-input` folder with the generic image files from the previous job.

### DIFF
--- a/ci-automation/image.sh
+++ b/ci-automation/image.sh
@@ -104,5 +104,41 @@ function _image_build_impl() {
 
     sign_artifacts "${SIGNER}" "images/latest/"*
     copy_to_buildcache "images/${arch}/${vernum}/" "images/latest/"*
+
+    (
+    set +x
+    # Don't fail the whole job
+    set +e
+    echo "==================================================================="
+    export BOARD_A="${arch}-usr"
+    export FROM_A="release"
+    export VERSION_A="current"
+    if [ "${channel}" = "developer" ]; then
+            export CHANNEL_A="alpha"
+    else
+            export CHANNEL_A="${channel}"
+    fi
+    export FROM_B="file://${PWD}/images/latest"
+    # Use the directory directly (and BOARD_B and CHANNEL_B are unused)
+    export VERSION_B="."
+    echo "== Image differences compared to ${CHANNEL_A} ${VERSION_A} =="
+    rm -f package-diff
+    curl -fsSLO --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 "https://raw.githubusercontent.com/flatcar-linux/flatcar-build-scripts/master/package-diff"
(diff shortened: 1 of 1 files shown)


pothos:
Will rework this to use separate folders for the generic image as input and the VM images as output. Otherwise gpg complains when files already have signatures.

ci-automation/image.sh
```diff
@@ -98,7 +95,50 @@ function _image_build_impl() {
...
+            -v "${vernum}" \
+            mv "${CONTAINER_IMAGE_ROOT}/${arch}-usr/" "./${images_out}/"
+    cd "images/latest"
+    sign_artifacts "${SIGNER}" *
+    copy_to_buildcache "images/${arch}/${vernum}/" *
+
+    (
+    set +x
+    # Don't fail the whole job
+    set +e
```
krnowak, line 111: Would this work (to limit the scope of a different working directory to the subshell)?

```suggestion
    sign_artifacts "${SIGNER}" "images/latest/"*
    copy_to_buildcache "images/${arch}/${vernum}/" "images/latest/"*

    (
    set +x
    # Don't fail the whole job
    set +e
    cd "images/latest"
```

pothos: I've added two wip commits and started a new build to test if it really uploads the correct paths http://192.168.42.7:8080/job/container/job/packages/691/cldsv/

pothos: Seems to work https://bincache.flatcar-linux.net/images/amd64/9999.1.2+noimagecontainer-kai2/

pothos: (Had to deal with way too many git conflicts…)

krnowak: Thanks for doing it despite the conflicts. :)

krnowak (approved):

PR #356, opened 2022-06-28
.github: Run CI workflow only for PR events

pothos:
Building for the branch push event causes two builds per PR and is not
needed anyway (we have nightly builds for the main branch).
Only consider PR events to trigger the CI build.


krnowak (approved):

pothos:
Backporting as well…

PR #357, opened 2022-06-28
jenkins/manifest: Generalize manifest ID overriding

krnowak:
This should allow kola to properly filter tests for dev builds based
on maintenance branches.

This will work if https://github.com/kinvolk/jenkins-os/pull/237 gets merged together with this PR.

Need to find out how to test it, though.


krnowak:
Closing. It's a dead end, and we won't have this problem in the new pipeline.

PR #359, opened 2022-06-28
Add retag-for-jenkins helper

pothos:
When testing "scripts" changes with Jenkins and a leaf job fails, we
currently would have to restart the whole build or update the git tag
manually. Since dealing with the git tag is not straightforward, add
a helper to update it.

## How to use

See Usage

## Testing done

TODO


retag-for-jenkins
```diff
@@ -0,0 +1,34 @@
...
+git checkout --recurse-submodules "${TAG}"
+git reset --hard "${BRANCH}"
+echo "${BUILD_PATCH}" | git am -3 || {
+  git checkout "${TAG}" -- sdk_container/.repo/manifests/version.txt
+  git add sdk_container/.repo/manifests/version.txt
+  git am --continue
+  # This does not handle submodule conflicts: It should use the one
+  # from the TAG (similar to version.txt) unless an explicit new
+  # reference was specified
+} || { echo "Failed to resolve conflict" >&2 ; exit 1 ; }
```
krnowak, line 31: Should we call `git am --abort` here too?

retag-for-jenkins
```diff
@@ -0,0 +1,34 @@
...
+  echo
+  echo "TODO: Add feature to update coreos-overlay/portage-stable submodule refs"
+  exit 1
+fi
+
+TAG="$1"
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+git fetch --tags origin
+BUILD_PATCH=$(git format-patch --output=/dev/stdout "${TAG}~1..${TAG}")
+git checkout --recurse-submodules "${TAG}"
```
krnowak, line 22: I'm wondering if it wouldn't be better to do it in a temporary worktree to avoid messing with the current worktree. Like `git workplace add -b tmp-branch /tmp/foo "${TAG}"; cd /tmp/foo; …`?

retag-for-jenkins
```diff
@@ -0,0 +1,34 @@
...
+  git checkout "${TAG}" -- sdk_container/.repo/manifests/version.txt
+  git add sdk_container/.repo/manifests/version.txt
+  git am --continue
+  # This does not handle submodule conflicts: It should use the one
+  # from the TAG (similar to version.txt) unless an explicit new
+  # reference was specified
+} || { echo "Failed to resolve conflict" >&2 ; exit 1 ; }
+git tag -d "${TAG}"
+git tag "${TAG}"
+git push --force origin "${TAG}"
```
krnowak, line 34:
```suggestion
git push --force-with-lease origin "${TAG}"
```

retag-for-jenkins
pothos: I thought one can manually continue and select the right submodule ref

pothos:
```suggestion
} || { echo "Failed to resolve conflict, continue manually" >&2 ; exit 1 ; }
```
Does this help?

retag-for-jenkins
pothos: I've never used `git workplace`, will have to look into it - but if it's not a well-known thing I would rather keep the helper at a level where most people can understand what's going on and intervene manually if needed.

retag-for-jenkins
pothos: Wait, this did not work and `--force` is needed

retag-for-jenkins
krnowak: Sure, fair enough.

retag-for-jenkins
krnowak: Weird, but ok.

retag-for-jenkins
krnowak: Adding `, continue manually` is fine too.

retag-for-jenkins
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
+
+if [ $# -lt 1 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 TAG"
```
krnowak, line 6: You could add a paragraph when the script comes in handy. Basically what you wrote in the commit message.

pothos: Yes, thanks

krnowak (approved):

PR #365, opened 2022-07-01
ci-automation: Only store compressed images

pothos:
The new build pipeline compresses images already but uploaded both the
compressed and uncompressed files because the whole build folder gets
uploaded.
Add a new flag "--only_store_compressed" to the image generation which
deletes the uncompressed file after compression is done. Uncompressed
images are still supported if specified in the flag
"image_compression_formats".

Closes https://github.com/flatcar-linux/Flatcar/issues/793

## Testing done

TODO


build_library/release_util.sh
```diff
@@ -113,6 +115,13 @@ compress_disk_images() {
                     processed_format["${format}"]=1
                 fi
             done
+            # If requested, delete the input file after compression (only if 'none' is not part of the formats)
```
krnowak, line 118: Could you expand on why two names are excluded here?

pothos: Yes, good idea

sayanchowdhury (approved):

PR #366, opened 2022-07-04
Add concurency setting to CI workflow

gabriel-samfira:
This change will prompt the CI workflow to cancel a currently running job for a PR, if a new push is sent to that PR while the job is till running. This should prevent duplicate jobs running for the same PR simultaneously.

Signed-off-by: Gabriel Adrian Samfira <gsamfira@cloudbasesolutions.com>


jepio (approved):

pothos:
Backporting to ensure that PRs against older branches run don't run with an old workflow setup

PR #367, opened 2022-07-06
.github/ci: add kola qemu test

tormath1:
this initial attempt runs right after the "packages" jobs and downloads
the resulting artifacts.

The QEMU image is unzipped then the kola test is running using the
ci-automation scripts.


.github/workflows/ci.yaml
```diff
@@ -287,3 +287,47 @@ jobs:
...
+          set -euo pipefail
+
+          source ci-automation/test.sh
+
+          # TODO: REMOVE ME (for debugging purposes)
+          ls -ali
+          find . -type f -name "*.bz2"
+
+          # extract the image.
+          bzip2 --decompress scripts/images/**/flatcar_production_qemu_image.img.bz2
```
pothos, line 327: This is now done in the test itself (got merged this afternoon)

.github/workflows/ci.yaml
```diff
@@ -287,3 +287,47 @@ jobs:
...
             scripts/images/**/flatcar_test_update.gz
             scripts/ebuild_logs.tar.xz
+  test:
+    name: "Run kola tests"
+    needs: packages
+    runs-on:
+      - debian
+      - build
+      - self-hosted
+      - x64
```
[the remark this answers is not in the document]
gabriel-samfira: I added ```amd64``` as a tag to the pool configured for kola tests. Instead of ```x64``` you can just add ```${{ matrix.arch }}```. This, together with the ```kola``` tag instead of the ```build``` tag, will make sure the tests are run on the proper architecture.
```suggestion
      - ${{ matrix.arch }}
```

.github/workflows/ci.yaml
```diff
@@ -287,3 +287,47 @@ jobs:
             scripts/images/**/flatcar_production_*.sh
             scripts/images/**/flatcar_test_update.gz
             scripts/ebuild_logs.tar.xz
+  test:
+    name: "Run kola tests"
+    needs: packages
+    runs-on:
+      - debian
+      - build
```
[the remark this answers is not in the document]
gabriel-samfira:
```suggestion
      - kola
```

.github/workflows/ci.yaml
```diff
@@ -287,3 +287,47 @@ jobs:
...
+      matrix:
+        arch: ["amd64", "arm64"]
+    defaults:
+      run:
+        working-directory: scripts
+
+    steps:
+      # it downloads from the previous linked job.
+      - name: Download artifacts
+        uses: actions/upload-artifact@v3
```
gabriel-samfira, line 309: This should probably be ```actions/download-artifact@v3```
```suggestion
        uses: actions/download-artifact@v3
```

gabriel-samfira:
The ```actions/download-artifact``` cannot download artifacts from arbitrary workflow runs. It only works if you want to download artifacts uploaded by the current workflow. To download artifacts from arbitrary workflow runs, this action seems to be what we need:

https://github.com/dawidd6/action-download-artifact

However, I have not tested it. But it seems to do what we need. We can use it to create a separate workflow, maybe even a reusable workflow (reacts to the ```workflow_call``` event) to run kola tests against artifacts generated by any workflow run.

tormath1:
> The `actions/download-artifact` cannot download artifacts from arbitrary workflow runs. It only works if you want to download artifacts uploaded by the current workflow. To download artifacts from arbitrary workflow runs, this action seems to be what we need:
> 
> https://github.com/dawidd6/action-download-artifact
> 
> However, I have not tested it. But it seems to do what we need. We can use it to create a separate workflow, maybe even a reusable workflow (reacts to the `workflow_call` event) to run kola tests against artifacts generated by any workflow run.

Yeah I notified this limitation - but I don't get yet why we would need to run tests from a given workflow? To me these things needs to be done one after the other?

gabriel-samfira:
> Yeah I notified this limitation - but I don't get yet why we would need to run tests from a given workflow? To me these things needs to be done one after the other?

Agreed. Especially considering that if we want to simply re-run failed jobs, github allows us to do that.

One scenario that may be valid is if you make changes to tests, then want to re-run just the tests against an existing build. This way you don't have to wait for 2 hours just to rebuild the whole thing.

.github/workflows/ci.yaml
```diff
@@ -287,3 +287,47 @@ jobs:
...
+      - ${{ matrix.arch }}
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: ["amd64", "arm64"]
+    defaults:
+      run:
+        working-directory: scripts
+
+    steps:
```
gabriel-samfira, line 306: Considering this is a new job, running on a fresh runner, you will need to also clone the repo, check out submodules, etc:

https://github.com/flatcar-linux/scripts/blob/main/.github/workflows/ci.yaml#L77-L81

This might also be of interest in case some submodules are from different remotes:

https://github.com/flatcar-linux/scripts/blob/main/.github/workflows/ci.yaml#L109-L130

gabriel-samfira: Although, I am not sure if the submodules are needed to run kola tests. You can probably skip updating remotes and pulling the refs. Just the stuff from the first link should be enough  (i think).

gabriel-samfira: *sigh*

The runners are vanilla. So you'll have to install the needed bits for now:

https://github.com/flatcar-linux/scripts/blob/main/.github/workflows/ci.yaml#L58-L75

You can pick an choose which binaries make sense and add any additional packages the tests need to run.

PR #368, opened 2022-07-07
build_library: set correct SELinux contexts in final images

tormath1:
We need to basically run `restorecon -R -v /` on all files in the
final rootfs, before finalizing the image. Without doing that, nearly
every file will have `unlabeled_t` as its context. Then with
recent versions of selinux-base, some critical actions would not work
correctly in SELinux enforcing mode, e.g., loading Kernel modules.

To be able to run `restorecon`, `/sys/fs/selinux` has to be mounted.
Without that, it will silently skip relabelling.

<hr>

Initially proposed by @dongsupark

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -697,7 +697,25 @@ EOF
 
   # Build the selinux policy
   if pkg_use_enabled coreos-base/coreos selinux; then
-      sudo chroot "${root_fs_dir}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+    local did_mount_sys=0, did_mount_sys_fs_selinux=0
+    if ! mountpoint -q "${root_fs_dir}/sys"; then
+      did_mount_sys=1
+      sudo mount -t sysfs none "${root_fs_dir}/sys"
+    fi
+    if ! mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
+      did_mount_sys_fs_selinux=1
+      sudo mount -t selinuxfs none "${root_fs_dir}/sys/fs/selinux"
+    fi
+
+    sudo chroot "${root_fs_dir}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+    sudo chroot "${root_fs_dir}" bash -c "find / -maxdepth 1 | egrep -v 'boot|proc|sys' | xargs restorecon -R || true"
+
+    if [[ ${did_mount_sys_fs_selinux} -eq 1 ]]; then
+      sudo umount "${root_fs_dir}/sys/fs/selinux"
+    fi
+    if [[ ${did_mount_sys} -eq 1 ]]; then
+      sudo umount "${root_fs_dir}/sys"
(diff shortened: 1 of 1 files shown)


build_library/build_image_util.sh
```diff
@@ -697,7 +697,25 @@ EOF
 
   # Build the selinux policy
   if pkg_use_enabled coreos-base/coreos selinux; then
```
pothos, line 699: Please move this block before `write_contents` and `sudo fstrim`, otherwise the image content list is not correct and the fstrim doesn't have the full effect to reduce the image size to a minimum.

build_library/build_image_util.sh
```diff
@@ -697,7 +697,25 @@ EOF
...
+      did_mount_sys=1
+      sudo mount -t sysfs none "${root_fs_dir}/sys"
+    fi
+    if ! mountpoint -q "${root_fs_dir}/sys/fs/selinux"; then
+      did_mount_sys_fs_selinux=1
+      sudo mount -t selinuxfs none "${root_fs_dir}/sys/fs/selinux"
+    fi
+
+    sudo chroot "${root_fs_dir}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+    sudo chroot "${root_fs_dir}" bash -c "find / -maxdepth 1 | egrep -v 'boot|proc|sys' | xargs restorecon -R || true"
```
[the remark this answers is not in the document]
tormath1: This is going to disappear. I'm exploring a new approach: using `setfiles` directly (instead of `restorecon`). This only needs to read the `file_contexts` generated from the `semodule` compilation step.
```suggestion
    sudo setfiles -Dv -r "${BOARD_ROOT}" "${BOARD_ROOT}"/etc/selinux/mcs/contexts/files/file_contexts "${BOARD_ROOT}"
```

build_packages
```diff
@@ -301,6 +301,11 @@ test_image_content "${BOARD_ROOT}"
 # upload packages if enabled
 upload_packages
 
+# build and label the system before generating torcx image.
+info "Label the build root"
+sudo chroot "${BOARD_ROOT}" bash -c "cd /usr/share/selinux/mcs && semodule -s mcs -i *.pp"
+sudo setfiles -Dv -r "${BOARD_ROOT}" "${BOARD_ROOT}"/etc/selinux/mcs/contexts/files/file_contexts "${BOARD_ROOT}"
```
jepio, line 307: `setfiles` is a cool idea, but doing it in `build_packages` is not useful, because the images are not built from the files in `BOARD_ROOT`. This would have to happen somewhere in `build_image` and in `build_torcx_store`.

jepio: Also came across this: https://github.com/gardenlinux/gardenlinux/blob/main/bin/makepart#L20-L22

pothos:
> Closed in favor of #666 (imp)

Can you rather rebase? My PR only did the labeling for `/etc` which was what I needed to prevent overlay upcopies but most of your PR is still relevant, to fix any `/usr` labels and also produce correct torcx labels.

PR #374, opened 2022-07-11
Make the kola test workflow reusable

gabriel-samfira:
* Adds a reusable workflow that can run tests
  * Adapts the ci.yaml to use reusable workflow
  * Creates a new workflow that helps trigger tests using an arbitrary workflow run.

In this PR, the artifact download action is from my own fork, because it includes a fix that streams the artifact directly to disk. The upstream action downloads to memory and also un-archives the artifact in memory, which leads to huge memory usage in our case. I created a PR for this with the upstream project here:

https://github.com/dawidd6/action-download-artifact/pull/175

Signed-off-by: Gabriel Adrian Samfira <gsamfira@cloudbasesolutions.com>

### DIFF
--- a/.github/workflows/run-kola-tests.yaml
+++ b/.github/workflows/run-kola-tests.yaml
@@ -1,0 +1,137 @@
+name: "Run kola tests"
+on:
+  workflow_call:
+    inputs:
+      arch:
+        type: string
+        default: '["amd64", "arm64"]'
+        description: |
+          The architectures to run tests for. This is an array of architectures
+          that can be deserialized in JSON and results in an array of strings.
+        required: true
+      workflow_name_or_id:
+        type: string
+        required: true
+        default: ci.yaml
+        description: |
+          The workflow ID from where we'll download the artifacts to be tested.
+      workflow_run_id:
+        type: string
+        required: true
+        description: |
+          The run ID of the workflow specified in workflow_name_or_id
+      parallel_tests:
+        type: string
+        required: true
+        default: "2"
+        description: |
+          The number of parallel tests to run.
+
+jobs:
+  tests:
+    name: "Build Flatcar packages"
+    runs-on:
+      - self-hosted
+      - debian
+      - kola
+      - ${{ matrix.arch }}
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: ${{fromJSON(github.event.inputs.arch)}}
+
+    steps:
+      - name: Prepare machine
+        shell: bash
+        working-directory: ${{ github.workspace }}
+        run: |
+          sudo rm /bin/sh
+          sudo ln -s /bin/bash /bin/sh
+          sudo apt-get install -y ca-certificates curl gnupg lsb-release qemu-system qemu-user-static git bzip2 jq dnsmasq
+          sudo systemctl stop dnsmasq
+          sudo systemctl mask dnsmasq
+
+          # Install Docker-CE
+          sudo mkdir -p /etc/apt/keyrings
+          curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+          echo \
+            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/debian \
+            $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+          sudo apt-get update
+          sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
+
+          # Install latest Go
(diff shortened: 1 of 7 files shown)


pothos (commented):
Looks good, thanks!
For cherry-picking the changes in `ci-automation/*` should be done in a separate commit.
I'll look into the kola VM profile RAM situation - why was 8 GB swap chosen?

gabriel-samfira:
> Looks good, thanks! For cherry-picking the changes in `ci-automation/*` should be done in a separate commit. I'll look into the kola VM profile RAM situation - why was 8 GB swap chosen?

A little over what 3 VMs of 2.5 GB memory would consume. It gives a chance for tests to finish before an out of memory condition is reached. Hitting swap will slow down newly spawned tests to allow already spawned tests to finish and free up memory.

Without swap, some VMs will be killed when memory is exhausted.

.github/workflows/run-kola-tests.yaml
```diff
@@ -0,0 +1,137 @@
...
+          sudo systemctl mask dnsmasq
+
+          # Install Docker-CE
+          sudo mkdir -p /etc/apt/keyrings
+          curl -fsSL https://download.docker.com/linux/debian/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
+          echo \
+            "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/debian \
+            $(lsb_release -cs) stable" | sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
+          sudo apt-get update
+          sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-compose-plugin
```
tormath1, line 61: Is `docker-compose-plugin` necessary ?

.github/workflows/run-kola-tests.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+          # Install latest Go
```
tormath1, line 63: I think we can skip the Go installation because we're using Kola from the Docker version, so no need to build it from sources. :)

.github/workflows/run-kola-tests.yaml
```diff
@@ -0,0 +1,137 @@
...
+        run: |
+          exec 2>&1
+          set +x
+          set -euo pipefail
+
+          # extract the image.
+          IMG_ARCHIVE=$(readlink -f images/**/flatcar_production_image.bin.bz2)
+          bzip2 --decompress ${IMG_ARCHIVE}
+
+          # export the QEMU_IMAGE_NAME to avoid to download it.
```
tormath1, line 115: this comment can be moved after - on top of the `export QEMU_IMAGE_NAME=...`

.github/workflows/run-kola-tests.yaml
```diff
@@ -0,0 +1,141 @@
...
+            echo 'kernel.unprivileged_bpf_disabled=0' | sudo tee /etc/sysctl.d/enable_unprivileged_bpf.conf
+
+            sudo sysctl --system
+            sysctl -p
+
+            # Enable extra SWAP
+            sudo fallocate /swap.img -l 8GiB
+            sudo chmod 600 /swap.img
+            sudo mkswap /swap.img
+            sudo swapon /swap.img
```
pothos, line 73:
```suggestion
            if [ "${{ matrix.arch }}" == "amd64" ];then
              # Enable extra SWAP
              sudo fallocate /swap.img -l 8GiB
              sudo chmod 600 /swap.img
              sudo mkswap /swap.img
              sudo swapon /swap.img
            fi
```

ci-automation/vendor-testing/qemu.sh
```diff
@@ -49,6 +49,7 @@ kola run \
     --qemu-image="${QEMU_IMAGE_NAME}" \
     --tapfile="${CIA_TAPFILE}" \
     --torcx-manifest="${CIA_TORCX_MANIFEST}" \
+    --qemu-skip-mangle \
```
pothos, line 52: For backporting it would be great to have this change in a separate commit - we try to align `ci-automation/` for the branches but not the `.github` folder.

tormath1 (approved):

pothos (approved):

pothos:
A fix for the failed test is prepared in https://github.com/flatcar-linux/bootengine/pull/47

I just realize that we even should backport the .github changes (and before that the previous PR), otherwise the a PR against an older scripts branch like `flatcar-3227` would run without the changes done here, or?

gabriel-samfira:
> I just realize that we even should backport the .github changes (and before that the previous PR), otherwise the a PR against an older scripts branch like `flatcar-3227` would run without the changes done here, or?

Yes. The workflows would need to be backported to branches where you push PRs to as well as far as I remember.

gabriel-samfira:
> He, I was a bit late with picking #356 and it already started the workflow builds for all branches :D

no worries. They can easily be canceled in the "Actions" tab of the repository. :D

PR #375, opened 2022-07-11
ci-automation: Show changes by finding the previous channel

pothos:
The image comparison was done against the old release in the channel
we release to instead of the previous release with the same major
version. This means when a channel transition happens we see a large
diff instead of the diff against the previous release. While not bad
for finding problems, this is normally not needed. However, we want
to have two changelogs generated, one against the old release in the
channel we relese to and one against the previous release with the same
major version when a transition happens. There was no changelog
printing yet, and this is added now.

## How to use

backport

(when doing https://github.com/flatcar-linux/Flatcar/issues/794 this whole code block should gets its own function)

## Testing done

output is available in the [image job](http://192.168.42.7:8080/job/container/job/image/430/console)

### DIFF
--- a/ci-automation/image.sh
+++ b/ci-automation/image.sh
@@ -115,33 +115,65 @@ function _image_build_impl() {
     echo "==================================================================="
     export BOARD_A="${arch}-usr"
     export FROM_A="release"
-    export VERSION_A="current"
     if [ "${channel}" = "developer" ]; then
-            export CHANNEL_A="alpha"
+            NEW_CHANNEL="alpha"
     else
-            export CHANNEL_A="${channel}"
+            NEW_CHANNEL="${channel}"
     fi
+    NEW_CHANNEL_VERSION_A=$(curl -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 "https://${NEW_CHANNEL}.release.flatcar-linux.net/${BOARD_A}/current/version.txt" | grep -m 1 FLATCAR_VERSION= | cut -d = -f 2)
+    MAJOR_A=$(echo "${NEW_CHANNEL_VERSION_A}" | cut -d . -f 1)
+    MAJOR_B=$(echo "${FLATCAR_VERSION}" | cut -d . -f 1)
+    # When the major version for the new channel is different, a transition has happened and we can find the previous release in the old channel
+    if [ "${MAJOR_A}" != "${MAJOR_B}" ]; then
+        case "${NEW_CHANNEL}" in
+          lts)
+            CHANNEL_A=stable
+            ;;
+          stable)
+            CHANNEL_A=beta
+            ;;
+          *)
+            CHANNEL_A=alpha
+            ;;
+        esac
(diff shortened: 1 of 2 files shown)


krnowak:
Changes looks alright, but I'd like to see the output first. CI seems to fail, because of the wrong version string: `ERROR: invalid version '9999.99.99-kai-print-changelog', must start with 'main-', 'alpha-', 'beta-', 'stable-', or 'lts-'`

krnowak:
The output has:
```
12:30:42  Changes since alpha-3277.0.0
12:30:42  
12:30:42  ## Security fixes:
12:30:42  
12:30:42  fatal: invalid object name 'alpha-9999.99.99-kai-print-changelog'.
```

Is there an issue with your changes?

pothos:
I only had releases in mind which have tags for the submodules but dev builds not, I'll try to get the right submodule reference with `git ls-tree`

pothos:
Fix prepared in https://github.com/flatcar-linux/flatcar-build-scripts/pull/122

ci-automation/image.sh
```diff
@@ -115,33 +115,65 @@ function _image_build_impl() {
...
     echo
     echo "Image file size change (includes /boot, /usr and the default rootfs partitions), compared to ${CHANNEL_A} ${VERSION_A}:"
-    FILE=flatcar_production_image_contents.txt CALCSIZE=1 ./package-diff "${VERSION_A}" "${VERSION_B}"
+    FILE=flatcar_production_image_contents.txt CALCSIZE=1 flatcar-build-scripts/package-diff "${VERSION_A}" "${VERSION_B}"
     echo
     BASE_URL="http://${BUILDCACHE_SERVER}/images/${arch}/${vernum}"
     echo "Image URL: ${BASE_URL}/flatcar_production_image.bin.bz2"
+    echo
+    # Provide a python3 command for the CVE DB parsing
+    export PATH="$PATH:$PWD/scripts/ci-automation/python-bin"
```
krnowak, line 170: Why not running `show-changes` inside the SDK container? It should have python and the packaging package already. Too much fuss with setting everything up?

pothos: I thought about it but wasn't sure. Maybe we move this changelog printing to an own job and then it's faster to have it runnable without the SDK container.
The SDK container also adds all these
```
10:47:34  ###### Starting stopped container 'flatcar-images-amd64-9999.99.99-kai-print-changelog' ######
10:47:34  flatcar-images-amd64-9999.99.99-kai-print-changelog
10:47:34  usermod: no changes
10:47:34  bash: cannot set terminal process group (24): Inappropriate ioctl for device
10:47:34  bash: no job control in this shell
10:47:34  flatcar-images-amd64-9999.99.99-kai-print-changelog
10:47:34  
10:47:34  ###### Writing versionfile 'sdk_container/.repo/manifests/version.txt' to SDK '3292.0.0+nightly-20220706-2100', OS '9999.99.99+kai-print-changelog'. ######
10:47:34  
```
lines each time it is invoked^^

pothos: Tried it, the SDK's python doesn't have pip nor the `packaging` module

krnowak: Interesting, because `packaging` is a dependency of portage, so it should be a part of SDK already (https://github.com/flatcar-linux/portage-stable/blob/main/dev-python/packaging/packaging-21.3-r2.ebuild).

Nevermind, we can tackle this later.

krnowak (approved):

PR #376, opened 2022-07-11
ci-automation: Generate digests for artifacts

krnowak:
This PR adds generation of DIGESTS files and their armored ASCII file counterparts for the built artifacts (images, container tarballs). The only artifacts for which the DIGESTS files are not created are Gentoo packages in the "push packages" job. Not sure if it's worthwhile, because the Packages file already serves as a kinda-sorta-DIGESTS-file, and it's signed by us.

## How to use

Run the build (or use the CI build from below), see if the generated artifacts have DIGESTS and asc files next to them for sdk, packages, image and vms jobs.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/129/cldsv/

### DIFF
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -187,7 +187,8 @@ function docker_image_to_buildcache() {
     local tarball="$(basename "$image")-${version}.tar.gz"
 
     $docker save "${image}":"${version}" | $PIGZ -c > "${tarball}"
-    sign_artifacts "${SIGNER:-}" "${tarball}"
+    create_digests "${SIGNER:-}" "${tarball}"
+    sign_artifacts "${SIGNER:-}" "${tarball}"*
     copy_to_buildcache "containers/${version}" "${tarball}"*
 }
 # --
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -332,30 +332,67 @@ function sign_artifacts() {
     # rest of the parameters are directories/files to sign
     local to_sign=()
     local file
-    local files
 
     if [[ -z "${signer}" ]]; then
         return
     fi
 
-    for file; do
-        files=()
-        if [[ -d "${file}" ]]; then
-            readarray -d '' files < <(find "${file}" ! -type d -print0)
-        elif [[ -e "${file}" ]]; then
-            files+=( "${file}" )
-        fi
-        for file in "${files[@]}"; do
-            if [[ "${file}" =~ \.(asc|gpg|sig)$ ]]; then
-                continue
-            fi
-            to_sign+=( "${file}" )
-        done
-    done
+    list_files to_sign 'asc,gpg,sig' "${@}"
+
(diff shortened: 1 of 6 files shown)


krnowak:
See the output:
- images amd64: http://bincache.flatcar-linux.net/images/amd64/9876.0.0+krnowak-digests/
- images arm64: http://bincache.flatcar-linux.net/images/arm64/9876.0.0+krnowak-digests/
- containers: http://bincache.flatcar-linux.net/containers/9876.0.0-krnowak-digests/
- sdk: http://bincache.flatcar-linux.net/sdk/amd64/9876.0.0+krnowak-digests/

Digests for SDK are most likely [generated by catalyst](https://github.com/flatcar-linux/scripts/blob/056752222eb167bd6713078594577d4f2ff0cae6/build_library/catalyst.sh#L59). Maybe we should disable it to have digests generated by one routine. But this can happen when we disable the old pipeline.

krnowak:
Right now the script creates digests of almost all the files (so not only images, but also package listings, READMEs, scripts, JSON files and so on). Not sure if this is a problem, though.

pothos:
Thanks, looks good already! A minor problem I see is that the filenames in the DIGEST file have a path prefix now which they shouldn't have (e.g., there is `images/latest/flatcar-container.tar.gz`)

pothos:
I think it's ok to skip creating the summary digest files as they don't follow a clear scheme

krnowak:
> Thanks, looks good already! A minor problem I see is that the filenames in the DIGEST file have a path prefix now which they shouldn't have (e.g., there is `images/latest/flatcar-container.tar.gz`)

Well spotted, thanks. I think I only checked DIGESTS for containers, and those look good. :P

krnowak:
Updated, I have kicked off a rebuild of a vms job (http://jenkins.infra.kinvolk.io:8080/job/container/job/vms/454/) - this should overwrite some of the DIGESTS files with a fixed versions.

krnowak:
> Updated, I have kicked off a rebuild of a vms job (http://jenkins.infra.kinvolk.io:8080/job/container/job/vms/454/) - this should overwrite some of the DIGESTS files with a fixed versions.

Scratch that. First, I forgot to push my branch. Second, vms job pick up scripts repo from a tag, instead of my branch, so it will recreate the same bad DIGESTS files.

krnowak:
Started another SDK build - http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/132/

pothos:
> > Updated, I have kicked off a rebuild of a vms job (http://jenkins.infra.kinvolk.io:8080/job/container/job/vms/454/) - this should overwrite some of the DIGESTS files with a fixed versions.
> 
> Scratch that. First, I forgot to push my branch. Second, vms job pick up scripts repo from a tag, instead of my branch, so it will recreate the same bad DIGESTS files.

You can update the tag with `./retag-for-jenkins TAG` from your branch.

pothos (approved):

pothos:
Backported down to 3033

PR #378, opened 2022-07-13
generate SLSA provenance by default

jepio:
# generate SLSA provenance by default

Enable SLSA provenance report generation by default for board packages (including toolchains). SLSA provenance generation happens through a profile.bashrc file in profiles/coreos/base, it is enabled by defining GENERATE_SLSA_PROVENANCE=true in make.conf (or environment).

Digging deeper, provenance generation is enabled during `./build_packages` but also during `./build_toolchains`, because that is where we get binpkgs for board glibc/gcc and their dependencies (some 23 packages) from. An alternative way is available: `./setup_board --nousepkg --nogetbinpkg --board=amd64-usr` has been revived and can be used to skip `./build_toolchains`.

## How to use

Build a full image starting from toolchains (or sdk in case of container pipeline).

## Testing done

http://jenkins.infra.kinvolk.io:8080/job/os/job/manifest/6068/cldsv/

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -265,12 +265,15 @@ query_available_package() {
             tail -n 1
 }
 
-# Generate a list of packages installed in an image.
-# Usage: image_packages /image/root
-image_packages() {
-    local profile="${BUILD_DIR}/configroot/etc/portage/profile"    
+# List packages installed directly in portages package database
+image_packages_portage() {
     ROOT="$1" PORTAGE_CONFIGROOT="${BUILD_DIR}"/configroot \
         equery --no-color list --format '$cpv::$repo' '*'
+}
+# List packages implicitly contained in rootfs, such as in torcx packages or
+# initramfs.
+image_packages_implicit() {
+    local profile="${BUILD_DIR}/configroot/etc/portage/profile"
 
     # We also want to list packages that only exist in the initramfs.
     # Approximate this by listing build dependencies of coreos-kernel that
@@ -302,6 +305,13 @@ image_packages() {
     [ -z "${FLAGS_torcx_manifest}" ] ||
     torcx_manifest::sources_on_disk "${FLAGS_torcx_manifest}" |
     while read pkg ; do query_available_package "${pkg}" ; done
+}
+
+# Generate a list of packages installed in an image.
+# Usage: image_packages /image/root
+image_packages() {
+  image_packages_portage "$1"
+  image_packages_implicit "$1"
 }
 
 # Generate a list of installed packages in the format:
@@ -511,6 +521,35 @@ EOF
     sudo gzip -9 "${root_fs_dir}"/usr/share/licenses/common/*
 }
 
+# Add /usr/share/SLSA reports for packages indirectly contained within the rootfs
+# If the package is available in BOARD_ROOT accesses it from there, otherwise
+# needs to download binpkg.
+# Reports for torcx packages are also included when adding the torcx package to
+# rootfs.
+insert_extra_slsa() {
+  info "Inserting additional SLSA file"
+  local rootfs="$1"
+  for atom in $(image_packages_implicit "$rootfs"); do
+    pkg="${atom%::*}"
+    pkg="${pkg/\//_}.json.xz"
+    if [ -f "${BOARD_ROOT}/usr/share/SLSA/${pkg}" ]; then
+      info "Found ${atom} in BOARD_ROOT"
+      sudo cp "${BOARD_ROOT}/usr/share/SLSA/${pkg}" "${rootfs}/usr/share/SLSA/"
+      continue
+    fi
+    # let's not die if SLSA information is missing
+    pkgversion=$( (get_binary_pkg "=${atom}" 2>/dev/null ) || true)
+    binpkg="$(portageq-${BOARD} pkgdir)/${pkgversion}.tbz2"
+    if [ -f "${binpkg}" ]; then
+      info "Found ${atom} at ${binpkg}"
+      qtbz2 -O -t "${binpkg}" | \
+        sudo tar -C "${rootfs}" -xj --wildcards './usr/share/SLSA'
+      continue
+    fi
+    warn "Missing SLSA information for ${atom}"
+  done
+}
+
 # Add an entry to the image's package.provided
 package_provided() {
     local p profile="${BUILD_DIR}/configroot/etc/portage/profile"    
@@ -600,7 +639,7 @@ finish_image() {
           local casDigest="$(torcx_manifest::get_digest "${FLAGS_torcx_manifest}" "${pkg}" "${version}")"
           sudo cp "${FLAGS_torcx_root}/pkgs/${BOARD}/${pkg}/${casDigest}/${pkg}:${version}.torcx.tgz" \
             "${root_fs_dir}${on_disk_path}"
-
+          sudo tar xf "${root_fs_dir}${on_disk_path}" -C "${root_fs_dir}" --wildcards "./usr/share/SLSA"
           if [[ "${version}" == "${default_version}" ]]; then
             # Create the default symlink for this package
             sudo ln -fns "${on_disk_path##*/}" \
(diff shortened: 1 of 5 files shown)


jepio:
The provenance report generation assumes that it's running from within the sdk docker container. Generated reports are in `$ROOT/usr/share/SLSA`.

pothos:
I didn't fully understand it yet: Is calling `setup_board` after the SDK is built with the annotation data now necessary or not?

jepio:
> I didn't fully understand it yet: Is calling `setup_board` after the SDK is built with the annotation data now necessary or not? 

If the toolchains are built with annotation, then no changes to the existing image build flow are necessary.

With un-annotated toolchains, it would be necessary to run `./setup_board --force --nousepkg --nogetbinpkg` to generate the annotations. This didn't work when @t-lo tried it, because it had suffered from bitrot but it now works again.

> Edit: E.g., do we need to add it to `ci-automation/packages.sh` or is this just for development purposes when the used SDK does not have the annotation data for the pre-existing board (packages)?

It's only for development with unannotated sdk containers. The current failure in the github workflow is because it does not rebuild sdk/toolchains and I didn't add that `./setup_board --nousepkg --nogetbinpkg --force` call.

jepio:
> Probably because bootengine uses cros-workon, which sets `EGIT_REPO_URI` conditionally during unpack, instead of unconditionally at some top level. I think that this conditional setting of `EGIT_REPO_URI` variable is already forgotten by the time the post install hook is run. We could either fix cros-workon eclass in some way, or add a special case (detecting if `CROS_WORKON_REPO`, `CROS_WORKON_PROJECT` and maybe `CROS_WORKON_COMMIT` are set) to the script in coreos overlay.

That part works fine, but bootengine is not actually present in the image, the entry in the packages.txt list comes from here: https://github.com/flatcar-linux/scripts/blob/main/build_library/build_image_util.sh#L278-L290. Bootengine only consists of dracut modules and the update-bootengine script, nothing used at runtime.

pothos:
Should we merge it with the missing bootengine info or is there a way to add a workaround somewhere?

jepio:
> Should we merge it with the missing bootengine info or is there a way to add a workaround somewhere?

bootengine info is now included thanks to https://github.com/flatcar-linux/scripts/pull/378/commits/f14382827d49fb19c01b85b278a91022e4a3a4f3. So this is in mergeable/reviewable shape.

t-lo:
If we're building provenance by default should we set `GENERATE_SLSA_PROVENANCE` in `build_packages`?
(I'm currently running my tests with this variable exported manually)

jepio:
> If we're building provenance by default should we set `GENERATE_SLSA_PROVENANCE` in `build_packages`? (I'm currently running my tests with this variable exported manually)

The change to `./setup_board` sets this variable in `$BOARD_ROOT/etc/portage/make.conf`, so this will be applied by default but you would need a new SDK to test this locally.

t-lo:
> > If we're building provenance by default should we set `GENERATE_SLSA_PROVENANCE` in `build_packages`? (I'm currently running my tests with this variable exported manually)
> 
> The change to `./setup_board` sets this variable in `$BOARD_ROOT/etc/portage/make.conf`, so this will be applied by default but you would need a new SDK to test this locally.

Ah gotcha, thanks. I went with with `./build_sdk_container_image` (which builds the container image based on the "classic" tarball, and builds toolchains in the process) and I do indeed see a `GENERATE_SLSA_PROVENANCE="true"` in `/build/amd64-usr/etc/portage/make.conf`.

jepio:
Building here: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/156/

t-lo:
> Building here: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/156/

This currently fails at the `sdk-comtainer` step in `build_toolchains` with `fatal: unsafe repository ('/mnt/host/source/src/scripts' is owned by someone else)` (i.e. [here](https://github.com/flatcar-linux/scripts/blob/slsa-provenance-by-default/build_toolchains#L36)). This git security feature is what made me use `cat` in the [provenance implementation](https://github.com/flatcar-linux/coreos-overlay/pull/2027/files#diff-45695e997e08abb9c418b7a41bb439644568ceec70dd7da7a1da188fb56c69f0R117).

t-lo:
@jepio The PR looks good to me but currently fails CI. I did local testing and did not encounter the CI breakage (see previous comment).

Local tests done:
1. Rebuilt SDK container based on PR source tree, using `flatcar-sdk-amd64-3277.0.0.tar.bz2` as base SDK tarball.
    1. Verified toolchain provenance was generated in `/build/amd64-usr/usr/share/SLSA/`
2. Ran `build_packages` in custom SDK container from 1.
    1. Verified package provenance was generated in `/build/amd64-usr/usr/share/SLSA/`
3. Ran `build_image` and `image_to_vm.sh`.
    1. Started VM in qemu and saved a listing of `/usr/share/SLSA`
    ```shell
     ssh -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" \
        core@localhost -p 2222 'ls -1 /usr/share/SLSA' > slsa-list.txt
    ````
    2. Using `flatcar_production_image_packages.txt` from the image build and the above generated SLSA directory listing, checked for missing packages in the SLSA directory listing
    ```shell
    for line in $(sed -e 's/:.*//' -e 's:/:_:' flatcar_production_image_packages.txt); do
        if grep -q $line slsa-list.txt; then
            continue
        else
            echo "Missing package $line in SLSA"
        fi
    done
    ```
    This came up empty.
    2. Again using `flatcar_production_image_packages.txt` from the image build and the above generated SLSA directory listing, checked for for files present in the directory listing but not the package list
    ```shell
    for line in $(sed -e 's:_:/:' -e 's/\.json.*//' slsa-list.txt); do
        if grep -q $line flatcar_production_image_packages.txt; then
            continue
        else
            echo "spurious slsa $line missing from package list"
        fi
    done
    ```
    This came up with a single hit, `app-torcx/docker-20.10`, which is entirely reasonable since this is a torcx package not contained in the base image packages list.

PR LGTM as far as I am concerned as soon as the CI issue is fixed.

jepio:
> This currently fails at the `sdk-container` step in `build_toolchains` with `fatal: unsafe repository ('/mnt/host/source/src/scripts' is owned by someone else)` (i.e. [here](https://github.com/flatcar-linux/scripts/blob/slsa-provenance-by-default/build_toolchains#L36)). This git security feature is what made me use `cat` in the [provenance implementation](https://github.com/flatcar-linux/coreos-overlay/pull/2027/files#diff-45695e997e08abb9c418b7a41bb439644568ceec70dd7da7a1da188fb56c69f0R117).

Ugh, I actually hit the same issue as well and that's why i open-coded this:
https://github.com/flatcar-linux/coreos-overlay/pull/2027/commits/5d479a0a0bcdf805ad1341f7451d8d8486324696#diff-45695e997e08abb9c418b7a41bb439644568ceec70dd7da7a1da188fb56c69f0R142

I thought something like this should work: `git -c 'safe.directory=*' ...` but for some reason it doesn't, even though `git -c ... config -l` confirms the value is set correctly.

I've pushed a fix and have restarted the CI.

krnowak:
Image build seems to have failed with ([see the output](http://jenkins.infra.kinvolk.io:8080/job/container/job/image/506/consoleText)):
```
[2022-07-27T16:26:01.414Z] INFO    build_image: Inserting additional SLSA file
[2022-07-27T16:26:10.159Z] INFO    build_image: Found sys-kernel/bootengine-0.0.38-r13::coreos at /build/arm64-usr/var/lib/portage/pkgs/sys-kernel/bootengine-0.0.38-r13.tbz2
[2022-07-27T16:26:10.159Z] qtbz2: ignoring parent with unknown repo in profile /mnt/host/source/src/third_party/coreos-overlay/profiles/coreos/amd64: 
[2022-07-27T16:26:10.159Z] qtbz2: ignoring parent with unknown repo in profile /mnt/host/source/src/third_party/coreos-overlay/profiles/coreos/amd64/sdk: 
[2022-07-27T16:26:10.159Z] tar: ./usr/share/SLSA: Not found in archive
[2022-07-27T16:26:10.160Z] tar: Exiting with failure status due to previous errors
[2022-07-27T16:26:10.160Z] ERROR   build_image: script called: build_image '--board=arm64-usr' '--group=developer' '--output_root=/home/sdk/build/images' '--only_store_compressed' '--torcx_root=/home/sdk/build/torcx' 'prodtar' 'container'
```

krnowak (commented):
The job succeeded, I think it's good to take it in. After resolving the conflict, that is. :)

krnowak (approved):

PR #384, opened 2022-07-15
.github: Use lxc containers for amd64 kola tests, too

pothos:
While we moved the arm64 tests to lxc containers, amd64 stayed on VMs
which were not easy to scale up.
Now the GitHub Action runner is running on lxc containers and we can
spawn more VMs in parallel because it has no memory limit.


gabriel-samfira (approved):
LGTM.

pothos:
backported down to 3033

PR #386, opened 2022-07-18
ci-automation: Move git tagging into own script

pothos:
When the build system runs the packages jobs for both architectures in
parallel and has to create a new tag, tagging fails due to the race in
the tagging.
Move the git tagging to its own script that is run from a new top-level
job that starts the packages jobs for both architectures.

## How to use

backport

## Testing done

I've started a build in Jenkins and it created the `main-9999.0.0-kai-tag-in-packages-all-arch` tag with the desired submodules pinned.

### DIFF
--- a/ci-automation/packages-tag.sh
+++ b/ci-automation/packages-tag.sh
@@ -1,0 +1,120 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# >>> This file is supposed to be SOURCED from the repository ROOT. <<<
+#
+# packages_tag() should be called w/ the positional INPUT parameters below.
+
+# build tag automation stub.
+#   This script will update the versionfile with the OS packages version to build,
+#    and will add a version tag (see INPUT) to the scripts repo.
+#
+# PREREQUISITES:
+#
+#   1. SDK version is recorded in sdk_container/.repo/manifests/version.txt
+#   2. SDK container is either
+#       - available via ghcr.io/flatcar-linux/flatcar-sdk-[ARCH]:[VERSION] (official SDK release)
+#       OR
+#       - available via build cache server "/containers/[VERSION]/flatcar-sdk-[ARCH]-[VERSION].tar.gz"
+#         (dev SDK)
+#
+# INPUT:
+#
+#   1. Version of the TARGET OS image to build (string).
+#       The version pattern '(alpha|beta|stable|lts)-MMMM.m.p' (e.g. 'alpha-3051.0.0')
+#         denotes a "official" build, i.e. a release build to be published.
+#       Use any version diverging from the pattern (e.g. 'alpha-3051.0.0-nightly-4302') for development / CI builds.
+#       A tag of this version will be created in the scripts repo and pushed upstream.
+#
+#
+# OPTIONAL INPUT:
+#
+#   2. coreos-overlay repository tag to use (commit-ish).
+#       Optional - use scripts repo sub-modules as-is if not set.
+#       This version will be checked out / pulled from remote in the coreos-overlay git submodule.
+#       The submodule config will be updated to point to this version before the TARGET SDK tag is created and pushed.
+#
+#   3. portage-stable repository tag to use (commit-ish).
+#       Optional - use scripts repo sub-modules as-is if not set.
+#       This version will be checked out / pulled from remote in the portage-stable git submodule.
+#       The submodule config will be updated to point to this version before the TARGET SDK tag is created and pushed.
+#
+# OUTPUT:
+#
+#   1. Updated scripts repository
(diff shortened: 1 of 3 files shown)


ci-automation/packages-tag.sh
```diff
@@ -0,0 +1,120 @@
...
+        push_branch="true"
+        local existing_tag=""
+        existing_tag=$(git tag --points-at HEAD) # exit code is always 0, output may be empty
+        # If the found tag is a release or nightly tag, we stop this build if there are no changes
+        if [[ "${existing_tag}" =~ ^(stable|alpha|beta|lts)-[0-9.]+(|-nightly-[-0-9]+)$ ]]; then
+          local ret=0
+          git diff --exit-code "${existing_tag}" || ret=$?
+          if [ "$ret" = "0" ]; then
+            echo "Stopping build because there are no changes since tag ${existing_tag}" >&2
+            return 0
```
krnowak, line 102: The point of this piece of code was to avoid doing the packages build if there were no changes. This optimization is now gone. Do we want to keep it in the `packages.sh` script?

ci-automation/packages-tag.sh
```diff
@@ -0,0 +1,120 @@
...
+       && [[ "$(git -C sdk_container/src/third_party/coreos-overlay/ rev-parse --abbrev-ref HEAD)" =~ ^flatcar-[0-9]+$ ]] \
+       && [[ "$(git -C sdk_container/src/third_party/portage-stable/ rev-parse --abbrev-ref HEAD)" =~ ^flatcar-[0-9]+$ ]] ; then
+        push_branch="true"
+        local existing_tag=""
+        existing_tag=$(git tag --points-at HEAD) # exit code is always 0, output may be empty
+        # If the found tag is a release or nightly tag, we stop this build if there are no changes
+        if [[ "${existing_tag}" =~ ^(stable|alpha|beta|lts)-[0-9.]+(|-nightly-[-0-9]+)$ ]]; then
+          local ret=0
+          git diff --exit-code "${existing_tag}" || ret=$?
+          if [ "$ret" = "0" ]; then
```
krnowak, line 100:
```suggestion
          if [[ ${ret} -eq 0 ]]; then
```

ci-automation/packages-tag.sh
```diff
@@ -0,0 +1,120 @@
...
+            echo "Stopping build because there are no changes since tag ${existing_tag}" >&2
+            return 0
+          elif [ "$ret" = "1" ]; then
```
krnowak, line 103:
```suggestion
          elif [[ ${ret} -eq 1 ]]; then
```

ci-automation/packages.sh
```diff
@@ -10,8 +10,6 @@
 
 # OS image binary packages build automation stub.
 #   This script will use an SDK container to build packages for an OS image.
-#   It will update the versionfile with the OS packages version built,
-#    and will add a version tag (see INPUT) to the scripts repo.
```
krnowak, line 14: Prerequisites below probably should get an entry about the built version being recorded in versionfile.

ci-automation/packages-tag.sh
pothos: I have moved it here, meaning that the packages-tag job already stops without invoking the packages job.

pothos: Ok, we can also use a different file with an explicit name like `.skip-build`

krnowak: `skip-build` sounds good.

ci-automation/packages-tag.sh
krnowak: Didn't know about `[[ foo -eq 0 ]]` form (without the dollar sign). Thanks, TIL.

> When treating it as number, then this would be nicer, or? What I don't like with bash arithmetic is that it evaluates the empty string (or even alphabetic strings without `set -eu`) to 0 in this comparison which silently swallows this type error. The current way of string comparison is stricter.

I'd say that having a variable that supposed to be a number and it actually has no value means that we botched something earlier. I'd treat it as an undefined behaviour. :P Not that it happens here (can `$?` be empty?), so eh. Not insisting, both ways work.

pothos: Have updated it now

krnowak (approved):
Some nitpicking in comments.

ci-automation/packages-tag.sh
```diff
@@ -0,0 +1,120 @@
...
+    local sdk_version="${FLATCAR_SDK_VERSION}"
+
+    if [ -n "${coreos_git}" ] ; then
+        update_submodule "coreos-overlay" "${coreos_git}"
+    fi
+    if [ -n "${portage_git}" ] ; then
+        update_submodule "portage-stable" "${portage_git}"
+    fi
+
+    # Create new tag in scripts repo w/ updated versionfile + submodules.
```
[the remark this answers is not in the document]
pothos: When an article is used for `tag` we also need `the` for `scripts repo` or it sounds strange

PR #425, opened 2022-08-24
Cover Equinix Metal m3.small.x86 instances in release test

pothos:
The new m3.small instance does not have official Flatcar support yet
but we can already cover it in our PXE boot release tests.
The c3.small instances are legacy and m3.small is the new smallest
type.

## How to use

- Test that it works
- Backport

## Testing done

started something [here](http://192.168.42.7:8080/job/container/job/test/2076/console)

### DIFF
--- a/ci-automation/ci-config.env
+++ b/ci-automation/ci-config.env
@@ -69,7 +69,7 @@ EQUINIXMETAL_STORAGE_URL="${EQUINIXMETAL_STORAGE_URL:-gs://flatcar-jenkins/mantl
 EQUINIXMETAL_amd64_INSTANCE_TYPE="${EQUINIXMETAL_amd64_INSTANCE_TYPE:-c3.small.x86}"
 # Space separated list of instance types. On those instances the
 # cl.internet kola test will be run if this test is selected to run.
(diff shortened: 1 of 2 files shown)


pothos:
Test failed, need to try a manual PXE boot

pothos:
Manually it works but I noticed that the OOB console is empty

pothos:
Got it working with `console=ttyS1,115200n8` - in the current form we would add this to grub.cfg with a check for this platform but no idea how we could identify it

pothos:
> Test failed, need to try a manual PXE boot

Only the first test failed, the rerun was actually ok

krnowak:
> Got it working with `console=ttyS1,115200n8` - in the current form we would add this to grub.cfg with a check for this platform but no idea how we could identify it

The [oem-packet grub config](https://github.com/flatcar-linux/coreos-overlay/blob/main/coreos-base/oem-packet/files/grub.cfg) seems to already have this setting. Isn't it used?

pothos:
> The [oem-packet grub config](https://github.com/flatcar-linux/coreos-overlay/blob/main/coreos-base/oem-packet/files/grub.cfg) seems to already have this setting. Isn't it used?

I had to make it unconditional, maybe the platform needs a different condition

krnowak (approved):

PR #438, opened 2022-08-31
Support running under UID 500

pothos:
When started by the Flatcar core user, the SDK failed to use UID 500
because inside the SDK there already is the core user from nss-altfiles
with the same ID. This way, the SDK user was continuing with UID 1000
and had permission errors.
Allow to reuse an existing ID for the SDK user. To take effect without
creating new images, the sdk_entry.sh file from the scripts repo is
used instead of the internal copy. However, this only works when
usermod doesn't find a process that uses this ID, and we had a race
between the SDK entry points called by "docker start" and by
"docker exec". The race is unwanted anyway because we don't want to
execute the commands while setup_board is still running. Solve it by
setting the entrypoint for "docker start" to "sleep infinity" in
"docker create".

## How to use

Backport

## Testing done

Tested on Flatcar manually
TODO: run build pipeline

↑ not needed


run_sdk_container
```diff
@@ -129,7 +129,8 @@ if [ -z "$stat" ] ; then
        -e SDK_GROUP_ID="$(id -g)" \
        --name="$name" \
        --hostname="$hostname" \
-       "${container_image_name}"
+       --entrypoint /bin/sleep \
+       "${container_image_name}" infinity
```
t-lo, line 133: I don't think this is a good idea. It will leave the container running in the background forever..

t-lo (changes requested):
Looks good to me except for the unrelated (to the UID) "sleep" change - this has potential side effects and will change the behaviour of `run_sdk_container` (container remains running in the background).

I would propose to split off that part into a separate change.

t-lo:
Also, we likely want to back-port this change to Alpha, Beta, Stable, and LTS 2022.

run_sdk_container
```diff
@@ -146,4 +147,4 @@ fi
 # Workaround: The SDK expects to be able to write to /etc/hosts
 $docker exec "$name" sh -c 'cp /etc/hosts /etc/hosts2; umount /etc/hosts ; mv /etc/hosts2 /etc/hosts'
 
-$docker exec $tty -i "$name" /home/sdk/sdk_entry.sh "$@"
+$docker exec $tty -i "$name" /mnt/host/source/src/scripts/sdk_lib/sdk_entry.sh "$@"
```
krnowak, line 150: This change is unrelated to allowing UID 500, is it? Maybe make a separate commit for it with some explanation. I suppose this is useful for testing changes in sdk_entry.sh without recreating SDK container.

run_sdk_container
krnowak: Maybe use `/bin/true`?

pothos: So, `sleep 10` instead? This would also work and ensure `docker exec` has enough overlap. Currently the sdk_entry script runs like 1 second which normally enough to have `docker exec` work but under high load it won't work. The container has to run for `docker exec` to succeed. Since we have the cleanup, I thought inifinity is not that bad and the user at least sees the container in `docker ps` while a stopped container is only shown with `docker ps -a`.

run_sdk_container
pothos: My idea was to fix it this way without having to create new SDK containers first. I can split it up.

krnowak: Sounds good. Please make it a separate commit, no need for a separate PR.

run_sdk_container
```diff
@@ -129,7 +129,8 @@ if [ -z "$stat" ] ; then
        -e SDK_GROUP_ID="$(id -g)" \
        --name="$name" \
        --hostname="$hostname" \
-       "${container_image_name}"
+       --entrypoint /bin/sleep \
+       "${container_image_name}" 10
```
t-lo, line 133: Won't this circumvent the user id / group id change now? That's done by passing env variables to `sdk_entry.sh` which is only done at container creation time; however we change the entrypoint here to sleep, so `sdk_entry.sh` is never called with the `SDK_USER_ID` and `SDK_GROUP_ID` variables set?

t-lo (changes requested):

run_sdk_container
t-lo: (I would strongly prefer moving the sleep change to a separate PR)

run_sdk_container
pothos: I'll try to keep the same behavior by running `bash -l`

pothos: That works

t-lo (approved):
LGTM now, thanks for taking care of this and for the fast turn-around on the PR, Kai!

krnowak (approved):

PR #439, opened 2022-08-31
sdk_container: publish the SDK on a Docker registry

tormath1:

<hr>

On the CI side, we need to generate a PAT for the Flatcar buildbot and pass it from Jenkins to the script.

Related to: https://github.com/flatcar-linux/Flatcar/issues/739

### DIFF
--- a/ci-automation/sdk_container.sh
+++ b/ci-automation/sdk_container.sh
@@ -30,10 +30,16 @@
 #   4. SIGNING_KEY. Environment variable. The artifact signing key.
 #        Defaults to nothing if not set - in such case, artifacts will not be signed.
 #        If provided, SIGNER environment variable should also be provided, otherwise this environment variable will be ignored.
+
+#   5. REGISTRY_USERNAME. Environment variable. The username to use for Docker registry login.
+#        Defaults to nothing if not set - in such case, SDK container will not be pushed.
+#
+#   6. REGISTRY_PASSWORD. Environment variable. The password to use for Docker registry login.
+#        Defaults to nothing if not set - in such case, SDK container will not be pushed.
 #
 # OUTPUT:
 #
-#   1. SDK container image of the new SDK, published to buildcache.
+#   1. SDK container image of the new SDK, published to buildcache and optionally on a Docker registry.
 #   2. "./ci-cleanup.sh" with commands to clean up temporary build resources,
 #        to be run after this step finishes / when this step is aborted.
 #   3. If signer key was passed, signatures of artifacts from point 1, pushed along to buildcache.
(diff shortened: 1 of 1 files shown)


ci-automation/sdk_container.sh
```diff
@@ -75,5 +81,19 @@ function _sdk_container_build_impl() {
...
+
+    # If the registry password or the registry username is not set, we leave early.
+    [[ -z "${REGISTRY_PASSWORD}" ]] || [[ -z "${REGISTRY_USERNAME}" ]] && return
+
+    # We assert that we are correctly logged into the container registry.
+    echo "${REGISTRY_PASSWORD}" | docker login "${CONTAINER_REGISTRY}" -u "${REGISTRY_USERNAME}" --password-stdin
+
+    # Docker images are pushed in the container registry.
+    for a in all amd64 arm64; do
+      docker push "${CONTAINER_REGISTRY}/flatcar-sdk-${a}":"${docker_vernum}"
```
pothos, line 96: For the future: In case we build something releasable every night, I wouldn't publish it directly here in this job but do the SDK publishing in the release job which is run manually - otherwise we clutter the registry too much.
For now it's ok to do it here, no problem.

pothos: Or as Gabriel suggested, we use garbage collection to remove the unused images after one week, only keeping those that are used in the version.txt files of the releases under CHANNEL.release.flatcar-linux.net

tormath1: quickly checked, but I don't see a limitation for GH Container Registry? In that case, we could simply push each built SDK, without pruning them.

pothos: Deleting development SDKs makes sense to give people a chance to mirror the registry efficiently and also helps to reduce the environmental impact ;)

tormath1: I just been through the `release` job - I think that would make more sense to move theses lines directly into the release script.

pothos: As you like. One argument for doing it here would be that we could skip the tar ball import to create the local Docker image which takes a couple of minutes. Not sure if this needs some other changes in the pipeline and we would still want to keep the possibility to work without a registry, I think?

pothos: Should we move this to https://github.com/flatcar/scripts/pull/475?

tormath1: Yes, let's do this.

tormath1:
closed in favor of: https://github.com/flatcar/scripts/pull/475

PR #445, opened 2022-09-01
ci-automation: Prepare new release job

pothos:
- ci-automation: Prepare release job
    The old pipeline had a release job where mantle's plume release tool
    was invoked to publish the cloud images.
    Implement a release job in the new pipeline with the same goals and
    eventually even more automation.
- ci-automation: Move image change report to own file
    To review the image changes and the changelog more easily and in case
    of fixes, iterate over it without rebuilding the image, move this logic
    to its own file where a new job could call it.

## How to use

This is currently more a draft that outlines what should be done where and you are invited to pick up parts, either by merging this and having follow-up PRs or by pushing your changes here.
 
I think it makes sense to backport each PR directly instead of waiting to have all implemented.

## Testing done

Ran the jobs on Jenkins. Since it was a dev build, there were no big changes printed compared to the last Alpha from this week. The release job is currently starting the mantle container with a no-op.


pothos:
I've tested it in the current form, I think it would be nice to merge it now and iterate over it in follow-up PRs which may be even focusing on different aspects, e.g., one PR adds plume, one adds the cloud-formation update, etc.

ci-automation/image-changes.sh
```diff
@@ -0,0 +1,133 @@
+#!/bin/bash
+#
+# Copyright (c) 2021 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# >>> This file is supposed to be SOURCED from the repository ROOT. <<<
+#
+# image_build() should be called w/ the positional INPUT parameters below.
```
krnowak, line 9:
```suggestion
# image_changes() should be called w/ the positional INPUT parameters below.
```

ci-automation/image-changes.sh
```diff
@@ -0,0 +1,133 @@
...
+
+# Binary OS image build automation stub.
+#   This script will build the OS image from a pre-built packages container.
```
krnowak, line 12:
```suggestion
# OS image differences display stub.
#   This script will display the differences between the last released image and the currently built one.
```

ci-automation/image-changes.sh
```diff
@@ -0,0 +1,133 @@
...
+#
+# PREREQUISITES:
+#
+#   1. SDK version and OS image version are recorded in sdk_container/.repo/manifests/version.txt
+#   2. Scripts repo version tag of OS image version to be built is available and checked out.
+#   3. The generic Flatcar image must be present in build cache server.
```
krnowak, line 18:
```suggestion
#   1. Artifacts describing the built image (kernel config, contents, packages, etc.) must be present in build cache server.
#   2. Scripts repo version tag of OS image version to be built is available and checked out.
```

ci-automation/image-changes.sh
```diff
@@ -0,0 +1,133 @@
...
+#
+# OPTIONAL INPUT:
+#
+#   1. SIGNER. Environment variable. Name of the owner of the artifact signing key.
+#        Defaults to nothing if not set - in such case, artifacts will not be signed.
+#        If provided, SIGNING_KEY environment variable should also be provided, otherwise this environment variable will be ignored.
+#
+#   2. SIGNING_KEY. Environment variable. The artifact signing key.
+#        Defaults to nothing if not set - in such case, artifacts will not be signed.
+#        If provided, SIGNER environment variable should also be provided, otherwise this environment variable will be ignored.
```
krnowak, line 32: This can be dropped.

ci-automation/image-changes.sh
```diff
@@ -0,0 +1,133 @@
...
+#
+# OUTPUT:
+#
+#   1. Currently the script prints the image differences compared to the last release and the changelog for the release notes but doesn't store it yet in the buildcache.
+#   2. "./ci-cleanup.sh" with commands to clean up temporary build resources,
+#        to be run after this step finishes / when this step is aborted.
+#   3. If signer key was passed, signatures of artifacts from point 1, pushed along to buildcache.
+#   4. DIGESTS of the artifacts from point 1, pushed to buildcache. If signer key was passed, armored ASCII files of the generated DIGESTS files too, pushed to buildcache.
```
krnowak, line 40: This can be dropped.

krnowak (commented):
Some nitpicks for the image changes job. Docs in the release job will be amended in follow up PRs.

krnowak (approved):

PR #452, opened 2022-09-05
*: Allow specifying extra URLs for torcx packages

krnowak:
Torcx manifest may contain paths and URLs as locations of
packages. There are two kinds of packages - vendored and
extra. Vendored packages normally have two locations - path to the
directory inside the image where the package is (which is why it's
called vendored), and a URL to the package on some remote
server. Extra packages only have a URL. But the URLs are added only
when we tell the build_torcx_store script to upload the packages at
the same time, which is what the old build pipeline was doing. With
the new pipeline, the upload happens as a separate step, thus the
upload is disabled when invoking build_torcx_store, and so the
packages are not getting URLs set. This change went unnoticed, because
a kola test checking the generated torcx manifest was only checking if
there is at least one location, either path or URL, and all the new
releases have no extra packages, only vendored ones.

When backporting the new pipeline to old LTS, the kola tests started
to fail, because old LTS had one extra package, and this is how I
noticed the problem.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/269/cldsv/

### DIFF
--- a/ci-automation/packages.sh
+++ b/ci-automation/packages.sh
@@ -80,6 +80,18 @@ function _packages_build_impl() {
     local vernum="${FLATCAR_VERSION}"
     local docker_vernum="$(vernum_to_docker_image_version "${vernum}")"
     local packages_container="flatcar-packages-${arch}-${docker_vernum}"
+    local torcx_pkg_url="https://${BUILDCACHE_SERVER}/images/${arch}/${vernum}/torcx"
+
+    source sdk_lib/sdk_container_common.sh
+
+    if is_official "${vernum}"; then
+        # A channel returned by get_git_channel should not ever be
+        # "developer" here, because it's an official build done from
+        # one of the maintenance branches. So if the channel happens
+        # to be "developer", then you are doing it wrong (releasing
+        # from the main branch?).
+        torcx_pkg_url="https://$(get_git_channel).release.flatcar-linux.net/${arch}-usr/${vernum}/torcx"
+    fi
 
     # Build packages; store packages and torcx output in container
     ./run_sdk_container -x ./ci-cleanup.sh -n "${packages_container}" -v "${vernum}" \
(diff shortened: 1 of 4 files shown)


ci-automation/packages.sh
```diff
@@ -80,6 +80,18 @@ function _packages_build_impl() {
...
+
+    source sdk_lib/sdk_container_common.sh
+
+    if is_official "${vernum}"; then
+        # A channel returned by get_git_channel should not ever be
+        # "developer" here, because it's an official build done from
+        # one of the maintenance branches. So if the channel happens
+        # to be "developer", then you are doing it wrong (releasing
+        # from the main branch?).
+        torcx_pkg_url="https://$(get_git_channel).release.flatcar-linux.net/${arch}-usr/${vernum}/torcx"
```
pothos, line 93: Does this need to be available during the kola tests already?

krnowak: Hm, well spotted. At first I thought that the test does a shallow check of the file structure, but it actually tries to download the packages with ignition and runs some basic docker tests on those.

So this won't fly - old pipeline was doing the uploading while right after building the torcx store, and it uploaded right away at to the final location. New pipeline uploads the packages to the staging area (bincache) and for releases my changes would expect the packages to be already on the origin server. I'll sleep with it.

krnowak: The torcx package URL needs to stay like I wrote it, because for releases, the packages will eventually be on origin, not in bincache.  And while writing it I noticed that we have [a `__prepare_torcx` function](https://github.com/flatcar-linux/scripts/blob/6d8d88cbe4db9a998dcacfdeb831d708fe525292/ci-automation/test.sh#L77-L100) that already mangles the downloaded torcx manifest by adding the URLs to the bincache, because otherwise the test would be failing. Interesting.

So I plan to change this mangling a bit to replace "release" URLs with "bincache" URLs.

krnowak:
CI passed. Not sure how to test the official build case.

pothos (approved):

krnowak:
Cherry-picked to 3033, 3346, 3277 and 3227.

PR #464, opened 2022-09-08
bootstrap_sdk_container: Fix a check for an official build

krnowak:
It should happen after we had a chance of processing input parameters, otherwise we were always operating on empty values which resulted in always having an unofficial build.


krnowak:
Tested locally.

pothos (approved):

krnowak:
Cherry picked to 3227, 3277, 3346 and 3033.

PR #511, opened 2022-09-27
.github: Use lts-info

krnowak:
We can use lts-info to optionally update mantle container in old LTS branch too.


pothos (approved):

krnowak:
Do we cherry-pick the workflow changes to maintenance branches too?

pothos:
That's only needed for PR workflows

PR #513, opened 2022-09-27
ci-automation: Fallback also to the mirror for container download

pothos:
When there is no SDK container image in the registry, the fallback looks at bincache but bincache isn't backed up and may be cleaned of old releases. While this won't be the regular case, the container image registry may be unavailable (or renamed as happened now), or people would like to rerun the image job which relies on the packages container.

## Testing done

None

### DIFF
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -212,11 +212,16 @@ function docker_image_from_buildcache() {
         return
     fi
 
+    # First try bincache then release to allow a bincache overwrite
     local url="https://${BUILDCACHE_SERVER}/containers/${version}/${tgz}"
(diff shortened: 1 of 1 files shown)


ci-automation/ci_automation_common.sh
```diff
@@ -212,11 +212,16 @@ function docker_image_from_buildcache() {
         return
     fi
 
+    # First try bincache then release to allow a bincache overwrite
     local url="https://${BUILDCACHE_SERVER}/containers/${version}/${tgz}"
+    local url_release="https://mirror.release.flatcar-linux.net/containers/"
```
krnowak, line 217: `url_release` should point to a file:

```suggestion
    local url_release="https://mirror.release.flatcar-linux.net/containers/${version}/${tgz}"
```

krnowak (approved):

PR #522, opened 2022-10-04
ci-automation/release: Only upload SDK if a new one was built

pothos:
A release includes an SDK if its SDK version is the release version. Only then we need to upload a new SDK container image.

CC @tormath1 

## How to use

Backport (to avoid unnecessary tar ball import, the upload gets skipped anyway)

## Testing done

none


tormath1 (approved):

krnowak (approved):

krnowak:
Was this also cherry-picked to 2605?

krnowak:
Alright, thanks. Must have typed something wrong here, because I couldn't see it. Probably did `git log flatcar-lts-2605` rather than `git log origin/flatcar-lts-2605`.

PR #530, opened 2022-10-13
ci-automation: Add an environment variable to skip build shortcuts

krnowak:
This will be used for the "run all tests" days in Jenkins. So even if no changes happened since last nightly build, a build will be done and the tests will be run.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/nightly/168/cldsv/

Made on [my test branch with some extra changes for testing](https://github.com/flatcar/scripts/tree/flatcar-999999).

Changes made for testing:
- limited nightly builds to alpha (because today is Tuesday and "run all tests" day today is for alpha channel),
- hardcoded flatcar-999999 scripts branch (can't be krnowak/flatcar-3374-test, for example, this gets checked),
- no packages job gets launched, just a message is printed
- packages tag script returns before modifying versionfile and pushing a new tag

### DIFF
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -507,3 +507,17 @@ function apply_local_patches() {
   done
 }
 # --
+
+# Returns 0 if passed value is either "1" or "true" or "t", otherwise
+# returns 1.
+function bool_is_true() {
+    case "${1}" in
+        1|true|t|yes|y)
+            return 0
+            ;;
+        *)
+            return 1
+            ;;
+    esac
+}
+# --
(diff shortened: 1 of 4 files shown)


krnowak:
Needs backporting to lts-2605 3033 3227 3346 3374.

jepio (commented):
Let's keep environment variable definitions in `ci-automation/ci-config.env`, they are basically all there right now.

krnowak:
Updated.

ci-automation/ci_automation_common.sh
```diff
@@ -507,3 +507,17 @@ function apply_local_patches() {
   done
 }
 # --
+
+# Returns 0 if passed value is either "1" or "true" or "t", otherwise
+# returns 1.
+function bool_is_true() {
+    case "${1}" in
+        1|true|t)
```
pothos, line 515:
```suggestion
        1|true|t|yes|y)
```
If thee values are allowed I might get confused which and also try `y` ;D

pothos (approved):

ci-automation/ci_automation_common.sh
krnowak: That's a good idea, thanks.

krnowak:
Cherry picked to 3033, 3227, 3374 and 3402.

PR #538, opened 2022-10-26
ci-automation: Support local patches

pothos:
For embargoed releases it is useful to apply patches locally to build with them before they are public. This allows to push the same patches to the repo during the Flatcar release at the embargo lift. The result is the same (as long as the scripts patches did not change parts of the setup logic that was running before they got applied), we can just build earlier and thus do the Flatcar release directly on the embargo lift instead of having to wait with the build because it would require the patches to be in the repos.

## How to use

Backport to all channels.

Used with the same Jenkins parameters that we used in the old pipeline.

## Testing done

Verified with:
[0001-Test-patch-for-scripts.patch.txt](https://github.com/flatcar/scripts/files/9870461/0001-Test-patch-for-scripts.patch.txt)
[0001-Test-patch-for-coreos-overlay.patch2.txt](https://github.com/flatcar/scripts/files/9871340/0001-Test-patch-for-coreos-overlay.patch2.txt)
The resulting image has both the changes from the scripts patch as well as the overlay patch.

### DIFF
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -475,3 +475,35 @@ function list_files() {
     done
 }
 # --
+
+# Looks for ../scripts.patch, ../overlay.patch, ../portage.patch and
+# applies them to the current repo or the respective sub-module checkout.
+function apply_local_patches() {
+  local patch_files=(../scripts.patch ../overlay.patch ../portage.patch)
+  local patch_file
+  local patch_id
+  local dirarg
+  echo "Looking for local patches ${patch_files[*]}"
+  for patch_file in "${patch_files[@]}"; do
+    if [ "${patch_file}" = "../scripts.patch" ]; then
+      dirarg=()
+    elif [ "${patch_file}" = "../overlay.patch" ]; then
+      dirarg=("-C" "sdk_container/src/third_party/coreos-overlay/")
+    elif [ "${patch_file}" = "../portage.patch" ]; then
+      dirarg=("-C" "sdk_container/src/third_party/portage-stable/")
+    else
+      echo "wrong case: unexpected ${patch_file}"
+      exit 1
+    fi
+    patch_id=$(test -e "${patch_file}" && { cat "${patch_file}" | git patch-id | cut -d ' ' -f 1 ; } || true)
+    if [ "${patch_id}" != "" ]; then
+      if git "${dirarg[@]}" log --no-merges -p HEAD | git patch-id | cut -d ' ' -f 1 | grep -q "${patch_id}"; then
+        echo "Skipping already applied ${patch_file}"
+      else
+        echo "Applying ${patch_file}"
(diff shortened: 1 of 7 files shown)


ci-automation/packages-tag.sh
```diff
@@ -120,5 +126,6 @@ function _packages_tag_impl() {
       create_versionfile "$sdk_version" "$version"
     )
     update_and_push_version "${version}" "${push_branch}"
+    apply_local_patches
```
krnowak, line 129: Isn't applying local patches in `packages-tag.sh` pointless? They will be applied in `packages.sh`, just before the build.

ci-automation/ci_automation_common.sh
```diff
@@ -475,3 +475,27 @@ function list_files() {
     done
 }
 # --
+
+# Looks for ../scripts.patch, ../overlay.patch, ../portage.patch and
+# applies them to the current repo or the respective sub-module checkout.
+function apply_local_patches() {
```
krnowak, line 481: Should this function have a check if these patches were already applied?

ci-automation/packages-tag.sh
pothos: Yes, I just wondered whether it still makes sense to have it when someone copies this job to do something else

ci-automation/ci_automation_common.sh
pothos: Will add

pothos: Did a rerun with the changes in http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/640/console

ci-automation/ci_automation_common.sh
```diff
@@ -475,3 +475,34 @@ function list_files() {
...
+  local patch_files=(../scripts.patch ../overlay.patch ../portage.patch)
+  local patch_file
+  local patch_id
+  echo "Looking for local patches ${patch_files[*]}"
+  for patch_file in "${patch_files[@]}"; do
+    if [ "${patch_file}" = "../scripts.patch" ]; then
+      dirarg=()
```
krnowak, line 488: `dirarg` needs to be declared as local somewhere earlier.

ci-automation/packages-tag.sh
krnowak: Maybe applying embargoed patches should be a separate step (a separate script in `ci-automation` directory) so that "someone" could just invoke it before doing this "something else". On the other hand, you have already merged the jenkins changes and you also have made applying those patches idempotent when they were already applied (so invoking ci-automation scripts locally won't fail there), so let's avoid the hassle for now.

ci-automation/ci_automation_common.sh
pothos: Thanks

krnowak (approved):

PR #540, opened 2022-10-28
vm_image_util: Use qemu-img for streamOptimized VMDKs

pothos:
To create the streamOptimized VMDK format we use the vmdk-convert
    utility. Nowadays qemu-img can also create this format directly,
    for supported formats see:
    https://www.qemu.org/docs/master/system/qemu-block-drivers.html

## Testing done

http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/653/cldsv/

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -587,12 +587,7 @@ _write_hdd_disk() {
 }
 
 _write_vmdk_stream_disk() {
-    # requires two pass conversion, qemu-img doesn't properly support the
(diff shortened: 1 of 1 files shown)


pothos:
Even if that doesn't make a difference for AWS I think it's still good in case it works well with VMware and then I'll reword the commit message.

pothos:
Looks like it doesn't work - the ESX VM test fails with: `uploading disks to ResourcePool: POST "/sdk": 400 Bad Request`

PR #565, opened 2022-11-08
common: Extend liblto link fixup function

krnowak:
We need to fix another symlink created by gcc-config, so extend the function that was doing it for some other links and rename it - it's not about only liblto any more.

Needs to be merged together with https://github.com/flatcar/portage-stable/pull/380 and https://github.com/flatcar/coreos-overlay/pull/2266.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/387/cldsv

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -189,7 +189,7 @@ emerge_to_image() {
   # Make sure profile.env has been generated
   sudo -E ROOT="${root_fs_dir}" env-update --no-ldconfig
 
-  fixup_liblto_softlinks "$root_fs_dir"
+  fixup_gcc_config_softlinks "${root_fs_dir}"
 
   # TODO(marineam): just call ${BUILD_LIBRARY_DIR}/check_root directly once
   # all tests are fatal, for now let the old function skip soname errors.
--- a/build_packages
+++ b/build_packages
@@ -295,7 +295,7 @@ fi
 
 eclean-$BOARD -d packages
 
-fixup_liblto_softlinks "${BOARD_ROOT}"
+fixup_gcc_config_softlinks "${BOARD_ROOT}"
 
 info "Checking build root"
 test_image_content "${BOARD_ROOT}"
--- a/common.sh
+++ b/common.sh
@@ -976,22 +976,37 @@ clean_qemu_static() {
   esac
 }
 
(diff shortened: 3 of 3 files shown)


thesamesam:
This looks like something we want to fix upstream. Please tell us about issues like this!

krnowak:
@thesamesam: Eventually I wanted to go upstream with it, but honestly I'm not sure if this is a problem with gcc-config or rather with our quirky setup. Thus weren't sure if there should be a fix for gcc-config, and what it would be. I mean, the immediate thought that I have here is "why don't we make those symlinks relative (so `${CHOST}-cc` -> `${CHOST}-gcc`, or a more complicated for liblto plugins)?". I can file an issue about it on Gentoo bugzilla.

thesamesam:
> @thesamesam: Eventually I wanted to go upstream with it, but honestly I'm not sure if this is a problem with gcc-config or rather with our quirky setup. Thus weren't sure if there should be a fix for gcc-config, and what it would be. I mean, the immediate thought that I have here is "why don't we make those symlinks relative (so `${CHOST}-cc` -> `${CHOST}-gcc`, or a more complicated for liblto plugins)?". I can file an issue about it on Gentoo bugzilla.

No worries. I just don't want downstream folks to be suffering if we've made a mistake or doing something wrong. The existence of https://github.com/gentoo/gentoo/pull/25340 made me suspicious as to whether {gcc,binutils}-config handles other bits wrong too. Would you mind filing a bug on Bugzilla?

tormath1 (approved):

thesamesam:
Thank you! This sounds like the right idea, but if we can, we should use `ln -r` (although this assumes GNU coreutils, I think) - otherwise I think it can be technically wrong (if unlikely) if EROOT is positioned somewhere odd?

krnowak:
Welp, there was a typo in the patch above - the `usr/` component of the second LIBLTO_PLUGIN should not be there. But otherwise seems to be working for us. About weird EROOT - liblto.so and its symlink are contained entirely withing `${EROOT}/usr`, so the relative symlinks never reaches EROOT itself. Anyway, gentoo bug filed - https://bugs.gentoo.org/880977

krnowak:
@pothos : For the record, I haven't backported this PR to other channels, as they don't have the problem that this PR is fixing.

PR #566, opened 2022-11-09
.github: Modernize mantle update action

krnowak:
This is to avoid warnings about deprecated node 12 actions and deprecated save-state and set-output commands.

### DIFF
--- a/.github/workflows/mantle-releases-main.yml
+++ b/.github/workflows/mantle-releases-main.yml
@@ -42,9 +42,9 @@ jobs:
             major=$(curl -sSL https://${{ matrix.branch }}.release.flatcar-linux.net/amd64-usr/current/version.txt | awk -F= '/FLATCAR_BUILD=/{ print $2 }')
             branch="flatcar-${major}"
           fi
-          echo ::set-output "name=BRANCH::${branch}"
-          echo ::set-output "name=SKIP::${skip}"
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


pothos (approved):

pothos:
@krnowak I don't see any new mantle ref update PRs, did this break anything?

.github/workflows/mantle-releases-main.yml
```diff
@@ -54,7 +54,7 @@ jobs:
         id: fetch-latest-mantle
         run: |
           commit=$(git ls-remote  https://github.com/flatcar/mantle refs/heads/flatcar-master | cut -f1)
-          echo ::set-output name=COMMIT::${commit}
+          echo "COMMIT=${COMMIT}" >>"${GITHUB_OUTPUT}"
```
pothos, line 57: Maybe this should be `echo "COMMIT=${commit}" >>"${GITHUB_OUTPUT}"`

pothos: But it's also good to add a check later for empty `COMMIT`

krnowak: Yeah, that should have been `"${commit}"`.

krnowak: I'll make the change and will test it.

pothos: Thanks

PR #572, opened 2022-11-17
ci-automation/release: Skip release job for Azure ARM in LTS

sayanchowdhury:


ci-automation/release.sh
```diff
@@ -94,6 +94,9 @@ function _inside_mantle() {
 
     for platform in aws azure; do
       for arch in amd64 arm64; do
+        if [[ "${arch}" == "arm64" ]] && [[ "${CHANNEL}" == "lts" ]]; then
+          continue
+        fi
```
krnowak, line 99: This will also skip aws. Not sure if it's desired. Maybe do something like this:

```suggestion
    for platform in aws azure; do
      arches=(amd64 arm64)
      if [[ "${CHANNEL} = 'lts' ]] && [[ "${platform}" = 'azure' ]]; then
        arches=(amd64)
      fi
      for arch in "${arches[@]}"; do
```

ci-automation/release.sh
```diff
@@ -119,6 +122,10 @@ function _inside_mantle() {
       done
     done
     for arch in amd64 arm64; do
+      if [[ "${arch}" == "arm64" ]] && [[ "${CHANNEL}" == "lts" ]]; then
```
krnowak, line 125: With the proposed fix above, this would not be needed.

krnowak: Wait, no. You'd need to change the loop in the line above to `for arch in "${arches[@]}"; do`.

ci-automation/release.sh
```diff
@@ -94,6 +94,9 @@ function _inside_mantle() {
 
     for platform in aws azure; do
       for arch in amd64 arm64; do
+        if [[ "${arch}" == "arm64" ]] && [[ "${CHANNEL}" == "lts" ]]; then
```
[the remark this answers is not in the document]
pothos: This should not skip aws or any future LTS releases with a new major version

ci-automation/release.sh
```diff
@@ -119,6 +122,10 @@ function _inside_mantle() {
       done
     done
     for arch in amd64 arm64; do
+      if [[ "${arch}" == "arm64" ]] && [[ "${CHANNEL}" == "lts" ]]; then
+        continue
+      fi
```
pothos, line 127:
```suggestion
```
Plume release must run to publish the AWS AMIs. I don't think we need to disable the Azure run because it only creates an URL.

ci-automation/release.sh
```diff
@@ -94,6 +94,9 @@ function _inside_mantle() {
 
     for platform in aws azure; do
       for arch in amd64 arm64; do
+        if [[ "${arch}" == "arm64" ]] && [[ "${CHANNEL}" == "lts" ]] && echo "$VERSION" | grep -q "^3033"; then
```
pothos, line 97:
```suggestion
        if [[ "${platform}" = "azure" ]] && [[ "${arch}" == "arm64" ]] && [[ "${CHANNEL}" == "lts" ]] && echo "$VERSION" | grep -q "^3033"; then
```

ci-automation/release.sh
```diff
@@ -93,7 +93,11 @@ function _inside_mantle() {
     secret_to_file google_release_credentials_file "${GOOGLE_RELEASE_CREDENTIALS}"
 
     for platform in aws azure; do
-      for arch in amd64 arm64; do
+      arches=(amd64 arm64)
+      if [[ "${CHANNEL}" == "lts" ]] && [[ "${platform}" == "azure" ]] && echo "$VERSION" | grep -q "^3033"; then
+        arches=(amd64)
+      fi
+      for arch in "${arches[@]}"; do
```
pothos, line 100: Next time someone could want to use `${arches[@]}` in the plume release for loop and this global change prevents the run of plume release for arm64. I would rather call it `arches_prerelease`.

```suggestion
      arches_prerelease=(amd64 arm64)
      if [[ "${CHANNEL}" == "lts" ]] && [[ "${platform}" == "azure" ]] && echo "$VERSION" | grep -q "^3033"; then
        arches_prerelease=(amd64)
      fi
      for arch in "${arches_prerelease[@]}"; do
```

ci-automation/release.sh
```diff
@@ -93,7 +93,11 @@ function _inside_mantle() {
     secret_to_file google_release_credentials_file "${GOOGLE_RELEASE_CREDENTIALS}"
 
     for platform in aws azure; do
-      for arch in amd64 arm64; do
+      arches=(amd64 arm64)
+      if [[ "${CHANNEL}" == "lts" ]] && [[ "${platform}" == "azure" ]] && echo "$VERSION" | grep -q "^3033"; then
```
krnowak, line 97: Forgot about one thing:

```suggestion
      # TODO: drop when LTS-3033 is not supported any more.
      if [[ "${CHANNEL}" == "lts" ]] && [[ "${platform}" == "azure" ]] && echo "$VERSION" | grep -q "^3033"; then
```

krnowak (approved):

ci-automation/release.sh
krnowak: No, nevermind, I thought this is a part of the `main` branch, but it's actually for `flatcar-3033`.

pothos: Ah, didn't see that either - I would suggest to have it on the `main` branch to have more clarity on what code runs where

krnowak (approved):
A lot of confusion from my side in this PR happened, sorry about that.

sayanchowdhury:
Thanks @krnowak @pothos

PR #600, opened 2022-11-28
qemu_update: Add update test from an old release

pothos:
To ensure that we can update from very old releases, add a test with a fixed old release, here the Stable release that introduced arm64 support to have the same test logic for both architectures.

## How to use

backport

## Testing done

in http://jenkins.infra.kinvolk.io:8080/job/container/job/test/4880/console

### DIFF
--- a/ci-automation/vendor-testing/qemu_update.sh
+++ b/ci-automation/vendor-testing/qemu_update.sh
@@ -12,7 +12,8 @@ set -euo pipefail
 
 source ci-automation/vendor_test.sh
 
-if [ "$*" != "" ] && [ "$*" != "*" ] && [ "$*" != "cl.update.payload" ]; then
+# The last check is not perfect (if both tests are rerun, it will only look at the name of the second test) but hopefully still good enough to prevent wrong usage
+if [ "$*" != "" ] && [ "$*" != "*" ] && [[ "$*" != *"cl.update.payload" ]]; then
     echo "1..1" > "${CIA_TAPFILE}"
     echo "not ok - all qemu update tests" >> "${CIA_TAPFILE}"
     echo "  ---" >> "${CIA_TAPFILE}"
@@ -30,8 +31,8 @@ else
     copy_from_buildcache "images/${CIA_ARCH}/${CIA_VERNUM}/flatcar_test_update.gz" tmp/
 fi
 
-if [ -f tmp/flatcar_production_image_previous.bin ] ; then
-    echo "++++ ${CIA_TESTSCRIPT}: Using existing ./tmp/flatcar_production_image_previous.bin for testing update to ${CIA_VERNUM} (${CIA_ARCH}) from previous ${CIA_CHANNEL} ++++"
+if [ -f tmp/flatcar_production_image_previous.bin ] && [ -f tmp/flatcar_production_image_first_dual.bin ] ; then
+    echo "++++ ${CIA_TESTSCRIPT}: Using existing ./tmp/flatcar_production_image_{previous,first_dual}.bin for testing update to ${CIA_VERNUM} (${CIA_ARCH}) from previous ${CIA_CHANNEL} and first dual-arch Stable release ++++"
 else
     echo "++++ ${CIA_TESTSCRIPT}: downloading flatcar_production_image_previous.bin from previous ${CIA_CHANNEL} ++++"
     rm -f tmp/flatcar_production_image_previous.bin.bz2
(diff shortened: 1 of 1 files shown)


pothos:
@krnowak if you want we could make use of your helpers to integrate the reun logic and so on, not sure if it's worth?

ci-automation/vendor-testing/qemu_update.sh
```diff
@@ -73,8 +78,30 @@ kola run \
...
+    --platform=qemu \
+    --qemu-bios="${bios}" \
+    --qemu-image=tmp/flatcar_production_image_first_dual.bin \
+    --tapfile="${CIA_TAPFILE}.first_dual" \
+    --torcx-manifest="${CIA_TORCX_MANIFEST}" \
+    --update-payload=tmp/flatcar_test_update.gz \
+    --qemu-skip-mangle \
+    cl.update.payload
+
+# When one of the above failed, the whole tests gets rerun
```
krnowak, line 99: I wouldn't say it's true. I noticed that `kola run` is executed under `set -e` so any of them failing would result in "${CIA_TAPFILE}" not being created, and thus the test runner not even trying to rerun the failed tests (because the tap analyser expects the file to exist).

pothos: Yes, it would need manual restarts. I will look into your suggestion to make it work with automatic retries.

ci-automation/vendor-testing/qemu_update.sh
```diff
@@ -73,8 +78,30 @@ kola run \
...
     --qemu-bios="${bios}" \
     --qemu-image=tmp/flatcar_production_image_previous.bin \
-    --tapfile="${CIA_TAPFILE}" \
+    --tapfile="${CIA_TAPFILE}.previous" \
     --torcx-manifest="${CIA_TORCX_MANIFEST}" \
     --update-payload=tmp/flatcar_test_update.gz \
     --qemu-skip-mangle \
     cl.update.payload
+
+kola run \
```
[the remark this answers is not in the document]
pothos: Updated it, I went with having the regular execution use the `previous` case instead of doing a dummy run.

ci-automation/vendor-testing/qemu_update.sh
```diff
@@ -12,7 +12,7 @@ set -euo pipefail
 
 source ci-automation/vendor_test.sh
 
-if [ "$*" != "" ] && [ "$*" != "*" ] && [ "$*" != "cl.update.payload" ]; then
+if [ "$*" != "" ] && [ "$*" != "*" ] && [[ "$*" != *"cl.update.payload" ]]; then
```
pothos, line 15: (This is not perfect, if both tests are rerun, it will only look at the second test's name but anyway it's just to skip the 20 retries if someone uses it wrong)

krnowak: Yeah, it's good enough. But maybe you could add what you wrote as a comment for this line?

ci-automation/vendor-testing/qemu_update.sh
```diff
@@ -67,14 +72,40 @@ if [ "${CIA_ARCH}" = "arm64" ]; then
...
-    --qemu-skip-mangle \
-    cl.update.payload
+query_kola_tests() {
+    shift; # ignore the instance type
+    local arg
+    arg="${*}"
+    if [ "${arg}" != "" ]; then
+      arg="cl.update.payload"
+    fi
+    kola list --platform=qemu --filter "${arg}"
```
krnowak, line 82: This could be a bit shorter:

```suggestion
    kola list --platform=qemu --filter "${*:+cl.update.payload}"
```

I always refer to https://wiki.bash-hackers.org/syntax/pe for parameter expansion stuff.

ci-automation/vendor-testing/qemu_update.sh
```diff
@@ -67,14 +72,40 @@ if [ "${CIA_ARCH}" = "arm64" ]; then
...
+      --board="${CIA_ARCH}-usr" \
+      --parallel="${QEMU_PARALLEL}" \
+      --platform=qemu \
+      --qemu-bios="${bios}" \
+      --qemu-image="${image}" \
+      --tapfile="${instance_tapfile}" \
+      --torcx-manifest="${CIA_TORCX_MANIFEST}" \
+      --update-payload=tmp/flatcar_test_update.gz \
+      --qemu-skip-mangle \
+      cl.update.payload
```
krnowak, line 108: Makes no difference here, but…

```suggestion
      "${@}"
```

ci-automation/vendor-testing/qemu_update.sh
pothos: Shorter or readable? :D

pothos: The `shift` would still be needed, or?

ci-automation/vendor-testing/qemu_update.sh
krnowak: For some reason I thought that "${@}" will be "filtered" by `query_kola_tests`, but that happens only for extra instances. Nevermind.

ci-automation/vendor-testing/qemu_update.sh
krnowak: > Shorter or readable? :D

> The `shift` would still be needed, or?

Oh yeah, shift is still needed. Drat, there goes my "Bash Golf". So, with the `shift` left intact, it will be neither short nor readable. :P

krnowak (approved):

PR #608, opened 2022-12-15
common: Use bincache for any developer build

krnowak:
For two-phase SDK build, we would like to use intermediate SDK as a seed. This SDK is only available on bincache, but only nightly builds could use bincache as the source of SDK tarballs. Let's allow any developer build to pull SDKs from bincache.

This won't fix a problem when building a release SDK using a two-phase build, but we rather won't use this feature for releases any time soon.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/468/cldsv


common.sh
```diff
@@ -339,7 +339,7 @@ readonly COREOS_EPOCH=1372636800
 TODAYS_VERSION=$(( (`date +%s` - ${COREOS_EPOCH}) / 86400 ))
 
 # Download URL prefix for SDK and board binary packages
-if [[ "${FLATCAR_BUILD_ID}" =~ ^nightly-.*$ ]] ; then
+if [[ -n "${FLATCAR_BUILD_ID}" ]] ; then
```
pothos, line 342: This also means setting up the binary package URL for the boards, or? It's a tricky question but I don't think we want to use binary packages by default for all dev builds because doing a fix and iterating over it with a rebuild would not result in the fix being picked up because the binary package takes precedence over building from source. The developer would have to bump the ebuild version each time, that's such a big trap.

pothos: (The URL gets used in `get_board_binhost`/`get_sdk_binhost` from `build_library/toolchain_util.sh` to set up the binpkg URLs)

krnowak: I have taken up a different approach, currently running CI in http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/512/cldsv. Let's see.

krnowak:
CI passed.

pothos (approved):

krnowak:
Backporting to 3493 to have working 2 phase builds in beta channel too.

PR #614, opened 2023-01-03
gc: pass OPENSTACK_CREDS to mantle container

tormath1:

<hr>

Otherwise, it fails with:
```
/work/ci-automation/garbage_collect_cloud.sh: line 20: OPENSTACK_CREDS: unbound variable
```


pothos (approved):

pothos:
Doesn't strictly need a backport but I would recommend it to reduce differences

PR #632, opened 2023-01-26
build_library/toolchain_util: Set up symlink for user patches

krnowak:
This was a place I missed where /etc/portage is set up. Because of it, user patches for sys-devel/gcc were not picked up.

Also stop using deprecated PORTDIR and PORTDIR_OVERLAY getters. We still set those variables, but we will drop them eventually.

It's pretty much standalone, but it's needed for https://github.com/flatcar/portage-stable/pull/409 and https://github.com/flatcar/coreos-overlay/pull/2404.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk_container/416/cldsv/


pothos (approved):

krnowak:
Cherry-picked to 3493.

PR #633, opened 2023-01-31
*: Fix user-patches setup

krnowak:
Normally `ln -sf path/to/target at/name` will create a symlink at `at/name` that points to `path/to/target`. But if `at/name` already exists and is a directory or a symlink to some other directory, then this command will create a symlink at `at/name/target` pointing to `path/to/target`. There is an ambiguity between 1st and 3rd form of `ln` (please refer to `man ln` for the available invocation forms). It can be disambiguated by using the `-T` flag to force the 1st form.

In our case, if `/etc/portage/patches` symlink already existed and was pointing to `<coreos-overlay>/coreos/user-patches`, we ended up with a useless symlink at `<coreos-overlay>/coreos/user-patches/user-patches` pointing to `<coreos-overlay>/coreos/user-patches`.

Nothing was really broken, just a useless and harmless symlink was created.

Will backport it to flatcar-3493.

### DIFF
--- a/build_library/catalyst.sh
+++ b/build_library/catalyst.sh
@@ -243,7 +243,7 @@ write_configs() {
     info "    stage4.spec"
     catalyst_stage4 > "$TEMPDIR/stage4.spec"
     info "Putting a symlink to user patches..."
-    ln -sf '/var/gentoo/repos/local/coreos/user-patches' \
+    ln -sfT '/var/gentoo/repos/local/coreos/user-patches' \
         "$TEMPDIR/portage/patches"
 }
 
--- a/setup_board
+++ b/setup_board
@@ -233,7 +233,7 @@ fi
 
 info "Configuring portage in ${BOARD_ROOT}"
 sudo mkdir -p "${BOARD_ETC}/portage/"{profile,repos.conf}
-sudo ln -sf "$(portageq get_repo_path / coreos)/coreos/user-patches" \
+sudo ln -sfT "$(portageq get_repo_path / coreos)/coreos/user-patches" \
    "${BOARD_ETC}/portage/patches"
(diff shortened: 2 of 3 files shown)


krnowak:
Updated #632 with a similar fix and merged.

build_library/catalyst.sh
```diff
@@ -243,7 +243,7 @@ write_configs() {
     info "    stage4.spec"
     catalyst_stage4 > "$TEMPDIR/stage4.spec"
     info "Putting a symlink to user patches..."
-    ln -sf '/var/gentoo/repos/local/coreos/user-patches' \
+    ln -sfT '/var/gentoo/repos/local/coreos/user-patches' \
```
pothos, line 246: Thanks for the explanation, now that you told this, I think we have the problem elsewhere, too, when we set up other symlinks for directories. Would be good to check them and fix it there as well^^ I think the symlink I set up for `/etc/coreos` → `/etc/flatcar`  is one of those examples but I bet there are more…

pothos (approved):

build_library/catalyst.sh
krnowak: True that, but one step at a time. Added an entry to my TODO list. :)

krnowak:
Cherry-picked to 3493.

build_library/catalyst.sh
pothos: I scanned our repos and the only case where we need this is in `init` which is a bug I found earlier but couldn't fix without the new knowlege from today, I'll create a PR.

pothos: https://github.com/flatcar/init/pull/88

krnowak: Thanks for doing it.

PR #634, opened 2023-02-01
Update report generation

krnowak:
This PR makes image job to generate some new report files:
- a "wtd" variant of contents files - these don't have timestamps nor user information, but they have information about device ID and inode, so this will make it easier for the file size accounting scripts to avoid hardlinks as they contribute little to the image size.
- contents files for initrd, both old and wtd variants, so we can use them to show changes inside the initrd. May come in handy when we update toolchain, dracut, bootengine or else.
- disk usage file for production image, so we can replace the size calculation done by package-diff

In order to generate the contents for initrd, I copied the script from flatcar-build-scripts.

Also moved the generation of the report files to the bottom of the `finish_image` function, where I mount the final production image as read only and generate the reports based on it.

Will want to backport it to alpha, beta, stable and LTS.

Tested it locally with https://github.com/flatcar/flatcar-build-scripts/pull/134.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1200/cldsv

This PR does not make any changes to the image changes job yet, I'd like to get this PR and the flatcar-build-scripts PR merged first.

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -244,9 +244,33 @@ systemd_enable() {
 write_contents() {
     info "Writing ${2##*/}"
     pushd "$1" >/dev/null
+    # %M - file permissions
+    # %n - number of hard links to file
+    # %u - file's user name
+    # %g - file's group name
+    # %s - size in bytes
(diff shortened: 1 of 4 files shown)


pothos (approved):

krnowak:

PR #635, opened 2023-02-02
Report file and file size changes in production image and initrd

krnowak:
Tested it locally with #634 and https://github.com/flatcar/flatcar-build-scripts/pull/134.

Will want to backport it to LTS, stable, beta and alpha.

Note that the new functionality is going to work only after we create new releases for the respective branches, so for now, all we will get is some error messages about some files not being found.

### DIFF
--- a/ci-automation/image_changes.sh
+++ b/ci-automation/image_changes.sh
@@ -103,6 +103,9 @@ function _image_changes_impl() {
     echo "Image kernel config changes, compared to ${CHANNEL_A} ${VERSION_A}:"
     FILE=flatcar_production_image_kernel_config.txt flatcar-build-scripts/package-diff "${VERSION_A}" "${VERSION_B}"
     echo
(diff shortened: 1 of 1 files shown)


ci-automation/image_changes.sh
```diff
@@ -86,6 +86,10 @@ function _image_changes_impl() {
     export FROM_B="bincache"
     export VERSION_B="${vernum}"
     export BOARD_B="${arch}-usr"
+    # First parts of the size-changes-report specs, the kind is
+    # appended at call sites.
+    SPEC_A_PART="${FROM_A}:${CHANNEL_A}:${arch}-usr:${VERSION_A}"
+    SPEC_B_PART="${FROM_B}:${arch}:${VERSION_B}"
```
pothos, line 92: One line has `-usr` one not

pothos: Ah, one is release, one it bincache, got it

krnowak:

PR #636, opened 2023-02-03
build_library/toolchain_util.sh: Install matching versions of toolchain packages

krnowak:
When adding a mask or accept keywords entry for some version of a toolchain package (gcc, libc, gdb, binutils or kernel headers), it can't be done by just doing it, for example, for `sys-devel/gcc`. Both `cross-{x86_64,aarch64}-cros-linux-gnu/gcc` needs to be masked/keyworded, otherwise `crossdev` will pick up the latest stable version for `cross-{x86_64,aarch64}-cros-linux-gnu/gcc` and this choice is not affected by masks or accept keywords of `sys-devel/gcc`.

This situation does not happen all the often, but when it happens, it's usually hard to remember to handle also the cross toolchain packages. Forgetting to do so leads to weird issues. So instead of telling crossdev to use the latest stable versions of cross toolchain packages, we will tell it to use specific versions that match the version of plain packages.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/597/cldsv (together with weekly updates)


krnowak:
CI passed.

krnowak:
This can be merged separately from the weekly updates.

krnowak:
To clarify - this can be merged separately from weekly updates, but weekly updates are blocked on this PR.

dongsupark (approved):

PR #647, opened 2023-02-08
ci-automation: Get two files to build vms instead of a whole directory

krnowak:
Getting the contents of the directory in the buildcache involves using rsync with some ssh invocation to log in as a bincache user. It's not a thing that will work locally unless the user gets ahold of the SSH key allowing the user to log in to buildcache as a bincache user.

Replace it by downloading two files that are actually needed for building vms: an image file and the version file. This just uses curl and is accessible for everyone.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1239/cldsv

Will backport it to other channels.

### DIFF
--- a/ci-automation/vms.sh
+++ b/ci-automation/vms.sh
@@ -109,7 +109,8 @@ function _vm_build_impl() {
 
     local images_in="images-in/"
     rm -rf "${images_in}"
-    copy_dir_from_buildcache "images/${arch}/${vernum}/" "${images_in}"
+    copy_from_buildcache "images/${arch}/${vernum}/flatcar_production_image.bin.bz2" "${images_in}"
+    copy_from_buildcache "images/${arch}/${vernum}/version.txt" "${images_in}"
     lbunzip2 "${images_in}/flatcar_production_image.bin.bz2"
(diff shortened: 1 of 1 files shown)


pothos (approved):
Thanks, seeing what was done this change here should also make sure that we don't copy too much anymore when rerunning the job.

krnowak:
> Thanks, seeing what was done this change here should also make sure that we don't copy too much anymore when rerunning the job.

If I understand what you wrote correctly, you mean skipping the download of `flatcar_production_image.bin.bz2` if we already have it locally (or rather, if we have `flatcar_production_image.bin`, since we unpack the archive), right? Possibly we could do that, but please note that we always do `rm -rf images-in` before downloading things to this directory. I suppose we could skip the deletion of the directory, and instead download the DIGESTS file and check if the existing file matches the DIGESTS file, and if it does not, then redownload the whole file.

pothos:
No, I meant that before your change, we would also download the OEM VM images, while not being needed, or?

krnowak:
> No, I meant that before your change, we would also download the OEM VM images, while not being needed, or?

Oh, I get it now. Yeah, that would happen if you decided to regenerate the OEM VM images.

krnowak:

PR #648, opened 2023-02-14
ci-automation: Allow git to work on directory owned by other user

pothos:
The get_git_channel function failed to work which resulted in the Alpha release job skipping the AWS publishing for the Alpha channel because it defaulted to the developer channel as fallback when git rejected to work on the directory owned by the build user while running as root user. A new version of git caused this behavior change and also prints an error message that explains to have to set safe.directory.

Set the git config entry safe.directory for the /work path when entering the mantle container where git runs as root while working on the directory owned by the build user.

## How to use

Either test it by creating a new Alpha release tag that has this change and then rerun the release job from that git tag, or merge, cherry-pick the change, and retag Alpha and rerun the release job from that git tag.

## Testing done

None

### DIFF
--- a/ci-automation/release.sh
+++ b/ci-automation/release.sh
@@ -208,7 +208,7 @@ function _release_build_impl() {
 
     touch sdk_container/.env # This file should already contain the required credentials as env vars
     docker run --pull always --rm --name="${container_name}" --net host \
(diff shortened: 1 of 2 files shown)


pothos:
Oh, right, we need it for all channels, will do.

pothos:
Ah, wait, I already did that but I missed the new beta major version because it wasn't on an active channel at that time.

pothos:
Done, in the future I think we should better only do channel transitions only when we release all channels

PR #666, opened 2023-02-20
Ship /etc from /usr through an overlay mount

pothos:
The existing tmpfile logic took care of folders that the ebuild keepdir directive wanted to exist on the OS. However, files and symlinks were not created, causing them to be missing if we didn't explicitly modify the ebuild files in coreos-overlay to use tmpfiles or patching of paths to be in /usr. We need a logic to provide /etc files from the current /usr partition without getting stale. This can be done best with an overlay mount which requires to keep the original /etc files under /usr.
Move the final /etc folder of the image build to /usr/share/flatcar/etc to serve as lower layer in the overlay. Also remove any state from the rootfs to make sure that we don't rely on it when testing our images before the release. What we get with an overlay mount is essentially a similar behavior to a 3-way merge because as long as the user didn't change the files, the old version is replaced with the new version and as soon as the user did changes, that file is frozen and wins over the provided old (in case of a rollback) or new versions from /usr. It does not work on file lines but on whole file contents, yet that is also what rpm-ostree does to my knowledge. Also, run tmpfiles once and do the SELinux labeling to prevent files being created in the upperdir because they were missing in the lowerdir, or because they had missing SELinux labels.

## How to use

Together with https://github.com/flatcar/bootengine/pull/53

## Testing done

See linked PR

↓ will do the changelog in the coreos-overlay PR

### DIFF
--- a/build_library/build_image_util.sh
+++ b/build_library/build_image_util.sh
@@ -752,9 +752,6 @@ finish_image() {
   sudo "${BUILD_LIBRARY_DIR}/gen_tmpfiles.py" --root="${root_fs_dir}" \
       --output="${root_fs_dir}/usr/lib/tmpfiles.d/base_image_var.conf" \
       ${tmp_ignore} "${root_fs_dir}/var"
-  sudo "${BUILD_LIBRARY_DIR}/gen_tmpfiles.py" --root="${root_fs_dir}" \
-      --output="${root_fs_dir}/usr/lib/tmpfiles.d/base_image_etc.conf" \
-      ${tmp_ignore} "${root_fs_dir}/etc"
 
   # Only configure bootloaders if there is a boot partition
   if mountpoint -q "${root_fs_dir}"/boot; then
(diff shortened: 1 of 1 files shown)


pothos:
This supersedes https://github.com/flatcar/scripts/pull/264

build_library/build_image_util.sh
```diff
@@ -782,18 +779,70 @@ EOF
...
+  sudo sed -i '/^[CLd] *\/etc\//d' "${root_fs_dir}"/usr/lib/tmpfiles.d/*
+
+  # SELinux: Label the root filesystem for using 'file_contexts'.
+  # The labeling has to be done before moving /etc to /usr/share/flatcar/etc to prevent wrong labels for these files and as
+  # the relabeling on boot would cause upcopies in the overlay.
+  # TODO: Breaks the system:
+  # sudo setfiles -Dv -r "${root_fs_dir}" "${root_fs_dir}"/etc/selinux/mcs/contexts/files/file_contexts "${root_fs_dir}"
+  # sudo setfiles -Dv -r "${root_fs_dir}" "${root_fs_dir}"/etc/selinux/mcs/contexts/files/file_contexts "${root_fs_dir}"/usr
+  # For now we only try it with /etc
+  sudo setfiles -Dv -r "${root_fs_dir}" "${root_fs_dir}"/etc/selinux/mcs/contexts/files/file_contexts "${root_fs_dir}"/etc
```
dongsupark, line 814: What about the case where selinux is disabled?
Wouldn't it better to have `if pkg_use_enabled coreos-base/coreos selinux; then` for the selinux-related commands?

dongsupark (approved):
Looks good in general.

build_library/build_image_util.sh
pothos: Thanks, will use it here, too!

PR #672, opened 2023-03-01
bootstrap_sdk: Allow running hooks to modify stage1 repos

krnowak:
Sometimes the modification to stage1 repos is necessary. For example we will need to update baselayout package in stage1, so we get the cleaned up handling of PATH and ROOTPATH, which will allow a newer portage to succeed in calculating dependencies in stage2.

Needed by https://github.com/flatcar/coreos-overlay/pull/2490 and https://github.com/flatcar/portage-stable/pull/423.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/680/console

### DIFF
--- a/bootstrap_sdk
+++ b/bootstrap_sdk
@@ -158,6 +158,20 @@ stage_repo() {
             && git fetch --all \
             && git checkout "$gitref" )
     fi
+    (
+        set -euo pipefail
+        local repo_var hook name
+
+        # FLAGS_coreos_overlay for gitname coreos-overlay
+        repo_var="FLAGS_${gitname//-/_}"
+        shopt -s nullglob
+        for hook in "${FLAGS_coreos_overlay}/coreos/stage1_hooks/"*"-${gitname}.sh"; do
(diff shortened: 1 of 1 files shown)


pothos:
Looks like the bootstrap path is still not perfect if adjustments to the seed's sources are needed… Anyway, good idea to have a framework for these fixups.

krnowak:
CI passed.

jepio:
Don't we already have a way to specify a git reference for stage1 repos instead of using the ones included in the sdk? Does that mechanism not work here?

krnowak:
> Don't we already have a way to specify a git reference for stage1 repos instead of using the ones included in the sdk? Does that mechanism not work here?

That's right. But I think specifying a git ref is an untenable solution. The problem with this mechanism is that you need to remember to pass the proper git refs every time you make a build in CI, otherwise the build will fail. Also, it does not work for nightly builds. I'd could even suggest dropping the options.

pothos (approved):

PR #684, opened 2023-03-13
Test change to trigger a build

pothos:
(Removes a trailing space)

## How to use

PR/branch can be used for fixing any regressions or for refactoring.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,7 +33,7 @@ on:
         required: false
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }} 
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
   cancel-in-progress: true
 
 permissions: {}


pothos:
Not sure why this intel microcode stuff fails, it worked in this PR CI build: https://github.com/flatcar/coreos-overlay/pull/2505

krnowak:
Maybe a too old coreos overlay is being pulled/checked out?

t-lo:
@krnowak is right, the branch [`kai/gha-pr-ci`](https://github.com/flatcar/scripts/tree/kai/gha-pr-ci) was created from nightly 2023-03-09; the coreos-overlay submodule therefore points to a ref before https://github.com/flatcar/coreos-overlay/pull/2505 was merged.

Since I'm planning to look into CI PR builds later this week I'll make a mental note to fail early if a branch was not rebased on latest main. Sorry for the confusion. Long term we likely want a more flexible approach to update submodule refs in scripts main; doing this once per night seems to slow us down too much.

@pothos could I please bug you to rebase your branch on latest main and try again?

pothos:
Good, the build indeed works but the test fails because the developer container is not built and even if, it would not be on bincache but it should be a GH Action build artifact that gets downloaded together with the image and could be downloaded from a local temporary server run or uploaded by kola through `scp`.

t-lo:
> Good, the build indeed works but the test fails because the developer container is not built and even if, it would not be on bincache but it should be a GH Action build artifact that gets downloaded together with the image and could be downloaded from a local temporary server run or uploaded by kola through `scp`.

I think the dev container is [built](https://github.com/flatcar/scripts/blob/main/.github/workflows/ci.yaml#L214) but not [uploaded](https://github.com/flatcar/scripts/blob/main/.github/workflows/ci.yaml#L274). 

If we could inject a custom [`IMAGE_URL` to kola](https://github.com/flatcar/mantle/blob/flatcar-master/kola/tests/devcontainer/devcontainer.go#L102) we could just upload the artifact and use the Github actions artifact cache. Or even better, maybe there is a way to make the test VM use a file in the local file system (the same way the torcx test already works)? Then there's no need to upload the dev container anywhere. Any GO experts who want to look into this? @tormath1 maybe?

> (Current aritfacts: https://github.com/flatcar/scripts/actions/runs/4416268645 - it would be nice to split them up more for easy downloading - as far as I remember the zipping of each file would still be needed?)

I agree, at least splitting VM images and build logs would make sense. We can't have too many artifacts though or we'll be [running into 429s](https://github.com/actions/upload-artifact#too-many-uploads-resulting-in-429-responses), so e.g. having one build log artifact per package would probably cause trouble.

I'm currently also looking into simplifying and speeding up the CI yaml build. For instance, there is no need to `docker commit` after each step since subsequent calls to `run_sdk_container` will simply re-use the previous steps' container image. That should save us quite a bit of time.

t-lo:
I've filed https://github.com/flatcar/Flatcar/issues/976 for the kola feature request.

t-lo:
Started poking the CI gh action in https://github.com/flatcar/scripts/blob/t-lo/speed-up-ci-build/.github/workflows/ci.yaml and am using manually triggered "Run build" actions with my branch for testing.

We can move our conversation to https://github.com/flatcar/Flatcar/issues/794 and close this PR if that's OK.

PR #696, opened 2023-03-23
Github Actions CI: streamline build, export all artifacts required by testing, improve testing

t-lo:
This PR streamlines Github actions CI builds and makes build steps re-use the build container, saving some build time.

It also adds previously missing artifacts required by the tests. It makes artifact upload more granular, so build logs, images, dev container, and torcx packages can be downloaded individually. It exports torcx tarball and binary packages for successive re-use in the kola tests.

The change also updates the github actions kola test runner workflow to use the new, separated artifacts produced by ci.yaml.

Further, it adds a fix for the devcontainer tests. Devcontainer and bin packages used in the devcontainer tests are now served from a local temporary web server.

The change also adds the qemu_update test and provides the respective update payload.

Also, the tests now use a local torcx_manifest.json produced by ci.yaml, which points to a torcx tarball also served by the local temporary web server.

Lastly, the change adds Markdown export capabilities to `tapfile_helper_lib.sh` to create a test run summary suitable for ingestion by Github. Test result reports for each run are posted to PRs as comments.

## How to use

Go to the ["Run build" GH action](https://github.com/flatcar/scripts/actions/workflows/ci.yaml), click on the "Run workflow" drop-down,  select branch "t-lo/speed-up-ci-build", and click on the green "Run workflow" button.

## Testing done

Ran the workflow.

## Cherry-pick and backport

This PR should be cherry-picked into all active maintenance branches (including LTS) as it would allow us to run automated builds and tests for auto-generated PRs (kernel, systemd, package refreshes).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,11 +3,6 @@ on:
   pull_request:
   workflow_dispatch:
     inputs:
-      bincache_server:
-        description: |
-          Bincache server.
-        default: "bincache.flatcar-linux.net"
-        required: true
       image_formats:
         description: |
           Space-separated vendor formats to build.
@@ -79,7 +74,6 @@ jobs:
       - name: Set environment
         shell: bash
         run: |
-          BUILDCACHE_SERVER="bincache.flatcar-linux.net"
           arch="${{ matrix.arch }}"
           COREOS_REMOTE=""
           COREOS_REF=""
@@ -87,20 +81,27 @@ jobs:
           PORTAGE_REF=""
           IMAGE_FORMATS="qemu_uefi"
 
-          [ -z "${{ github.event.inputs.bincache_server }}" ] || BUILDCACHE_SERVER="${{ github.event.inputs.bincache_server }}"
           [ -z "${{ github.event.inputs.coreos_remote }}" ] || COREOS_REMOTE="${{ github.event.inputs.coreos_remote }}"
           [ -z "${{ github.event.inputs.coreos_ref }}" ] || COREOS_REF="${{ github.event.inputs.coreos_ref }}"
           [ -z "${{ github.event.inputs.portage_remote }}" ] || PORTAGE_REMOTE="${{ github.event.inputs.portage_remote }}"
           [ -z "${{ github.event.inputs.portage_ref }}" ] || PORTAGE_REF="${{ github.event.inputs.portage_ref }}"
           [ -z "${{ github.event.inputs.image_formats }}" ] || IMAGE_FORMATS="${{ github.event.inputs.image_formats }}"
 
-          echo "BUILDCACHE_SERVER=${BUILDCACHE_SERVER}" >> $GITHUB_ENV
           echo "arch=${arch}" >> $GITHUB_ENV
           echo "COREOS_REMOTE=${COREOS_REMOTE}" >> $GITHUB_ENV
           echo "COREOS_REF=${COREOS_REF}" >> $GITHUB_ENV
           echo "PORTAGE_REMOTE=${PORTAGE_REMOTE}" >> $GITHUB_ENV
           echo "PORTAGE_REF=${PORTAGE_REF}" >> $GITHUB_ENV
           echo "IMAGE_FORMATS=${IMAGE_FORMATS}" >> $GITHUB_ENV
+
+          # Artifact root for images and torcx tarball as seen from within the container
+          echo "CI_CONTAINER_ARTIFACT_ROOT=/home/sdk/trunk/src/scripts/artifacts" >> $GITHUB_ENV
+          echo "CI_CONTAINER_TORCX_ROOT=/home/sdk/trunk/src/scripts/artifacts/torcx" >> $GITHUB_ENV
+          mkdir -p artifacts/torcx
+
+          # Placeholder URL for run-kola-tests.yaml, "Extract artifacts" step which will replace
+          #  this with its IP address.
+          echo "TORCX_TESTS_PACKAGE_URL=http://localhost:12345" >> $GITHUB_ENV
 
       - name: Checkout submodules
         shell: bash
@@ -129,7 +130,7 @@ jobs:
         shell: bash
         run: |
           exec 2>&1
-          set +x
+          set -x
           set -euo pipefail
 
           source ci-automation/ci_automation_common.sh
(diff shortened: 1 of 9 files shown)


.github/workflows/run-kola-tests.yaml
```diff
@@ -60,53 +60,163 @@ jobs:
...
+          tar -C $[DEVCONTAINER_WEBROOT} -xf torcx.tar
+
+          echo "------------"
+          ls -laR ${DEVCONTAINER_WEBROOT}
+          echo "------------"
+
+          # Extract the generic image we'll use for qemu tests.
+          # Note that the qemu[_uefi] tests use the generic image instead of the
+          #  qemu vendor VM image ("Astronaut: [...] Always have been.").
+          IMG_ARCHIVE=$(readlddink -f images/**/flatcar_production_image.bin.bz2)
```
pothos, line 166:
```suggestion
          IMG_ARCHIVE=$(readlink -f images/**/flatcar_production_image.bin.bz2)
```
Or is this something else?

.github/workflows/run-kola-tests.yaml
```diff
@@ -60,53 +60,163 @@ jobs:
...
+          echo "DEVCONTAINER_WEBROOT=${DEVCONTAINER_WEBROOT}" >> "$GITHUB_ENV"
+
+          # Set up webroot for devcontainer and torcx tests.
+          # The respective tests will download devcontainer and torcx tarball via http.
+          # The devcontainer test will then run a build
+          #   which will download and install binpkgs into the dev container.
+          # For the sake of that test we will serve both via a temporary local web server.
+          mkdir ${DEVCONTAINER_WEBROOT}
+          mv flatcar_developer_container* ${DEVCONTAINER_WEBROOT}
+          tar -C ${DEVCONTAINER_WEBROOT} -xcf binpkgs.tar.gz
```
pothos, line 155:
```suggestion
          tar -C ${DEVCONTAINER_WEBROOT} -xf binpkgs.tar.gz
```
Or `xvf` but maybe we don't need the full output?

.github/workflows/run-kola-tests.yaml
t-lo: That's a leftover which isn't actually needed anymore (no subdirectories anymore for image artifacts). Will remove, please pardon the confusion.

.github/workflows/run-kola-tests.yaml
t-lo: Yeah, that one just bit me in the actions run I triggered before signing off yesterday. Needs to be `-xzvf` (the `v` is indeed a good idea).

pothos:
In https://github.com/flatcar/scripts/pull/697/files I also added a new environment variable to the kola workflow. It will probably be a git conflict.

t-lo:
> In https://github.com/flatcar/scripts/pull/697/files I also added a new environment variable to the kola workflow. It will probably be a git conflict.

Good catch @pothos, thank you! This looks like an easy merge though. I'll rebase on main when your PR was merged.

t-lo:
This should take care of the devcontainer test breakage: https://github.com/flatcar/flatcar-dev-util/pull/14 .

t-lo:
Rebased on latest main.

t-lo:
Included the emerge-gitclone fix from latest coreos-overlay:main.

Let's see if we can trigger a rebuild+retest by re-requesting reviews. This is one of the trigger actions for `ci.yaml` and is supposed to work...

t-lo:
Test results artifacts are 10GB each, this seems excessive. Taking a look.

.github/workflows/ci.yaml
```diff
@@ -76,6 +76,21 @@ jobs:
           fetch-depth: 0
           submodules: true
 
+      # Hack alert: actions/checkout will check out the (disjunct) merge commit of a PR
+      #  instead of its head commit. That commit is not connected to any branch.
+      # This causes breakage downstream e.g. when the devcontainer test wants to check out
+      #  the ref in the scripts repo that corresponds to this build.
+      - name: Check out head commit of PR
```
pothos, line 83: I think we should also do this in the kola workflow for consistency - otherwise it's confusing if one works on the merged state and the other not.

pothos (commented):
Looks good!
The `coreos-overlay: bump ref to include emerge-gitclone fix` commit needs to go before merging, as we would not want to backport this submodule reference update but have to do backports to the `flatcar-MAJOR` branches of coreos-overlay.

.github/workflows/ci.yaml
t-lo: Oh right! Good catch.

t-lo:
> Looks good! The `coreos-overlay: bump ref to include emerge-gitclone fix` commit needs to go before merging, as we would not want to backport this submodule reference update but have to do backports to the `flatcar-MAJOR` branches of coreos-overlay.

Good point. I'll split this out into a separate PR then.

t-lo:
A few things are still open.

**TODO**:
- [ ] Split coreos-overlay bump into a separate PR and create respective PRs for flatcar-3033, flatcar-3549, flatcar-3510, flatcar-3374.
- [ ] Use `${{ github.event.pull_request.head.sha }}` for PR test runs in `run-kola-tests.yaml` in accordance with `ci.yaml`
- [ ] Don't upload 10GB of OS image artifacts generated by the qemu update test.
- [ ] For some reason the merged test runs TAP report is not present after testing concluded. Figure out why and archive the merged tests TAP report.

t-lo:
Manually triggered a new test run here: https://github.com/flatcar/scripts/actions/runs/4541175015

t-lo:
Everything should be in place now. Re-triggering a build/test run from scratch (by re-requesting review from pothos :) )

pothos (approved):

gabriel-samfira (approved):
This all looks reasonable. LGTM.

.github/workflows/run-kola-tests.yaml
```diff
@@ -111,11 +238,24 @@ jobs:
...
-            scripts/__TESTS__
-            scripts/results-.*.tap
+            scripts/__TESTS__/*/_kola_temp/
+            scripts/results-*.tap
+
+      - name: Create Test Summary
+        if: always()
+        uses: test-summary/action@v2
+        with:
+          paths: "scripts/results-*.tap"
```
gabriel-samfira, line 261: These files need to have the first line one of the following patters:

https://github.com/test-summary/action/blob/b266d9a22ea23552e00a4ee4276543880f5f686e/src/test_parser.ts#L299-L303

Otherwise it tries to parse them as XML, and you get the ```Error: Non-whitespace before first tag.``` message.

t-lo: It appears that we need to prepend a magic "TAP version 13" line. There are a [few examples in their unit tests](https://github.com/test-summary/action/tree/b266d9a22ea23552e00a4ee4276543880f5f686e/test/resources/tap) that suggest the format we emit should otherwise work. I'll patch and re-run.

t-lo: Oh wow that TAP parser is brittle. It insists on 2 spaces and who knows what else for embedded YAML. This makes me sad. TAP is a loosely defined format and it is meant to be interpreted loosely, not strictly.

I'll step away from that action and will investigate other options.

t-lo:
I was required to change quite a bit to make test results reporting usable for GitHub. Adding @krnowak as a reviewer because of his awesome bash skills and re-requested review from Kai.

krnowak:
I'll have a look tomorrow.

pothos (approved):

pothos:
One thing to improve is that the amd64 and arm64 status should not be posted to the same comment. Now one has to klick `edited` to see the first one of the status reports.

t-lo:
> One thing to improve is that the amd64 and arm64 status should not be posted to the same comment. Now one has to klick `edited` to see the first one of the status reports.

I'm a little confused it did that - but then I'm still learning the ropes with the add-pr-comment action.

Something we could do on our end to improve this is to merge both amd64 and arm64 into one report, using `tapfile_helper_lib.sh`. We could pass all TAP files to a separate merger action and use `qemu_uefi (amd64)` and `qemu_uefi (arm64)` as "vendors". This would have us end up with a report that includes all runs and would not require further changes to `tapfile_helper_lib.sh`. I'll have a look later.

I'll wait for Krzesimir's review (because I touched a lot of bash) and merge if I get his LGTM.

.github/workflows/run-kola-tests.yaml
```diff
@@ -108,14 +234,127 @@ jobs:
...
+  merge_and_publish_results:
+    name: "Merge TAP reports and post results"
+    needs: tests
+    if: always()
+    runs-on:
+      - self-hosted
+      - debian
+      - kola
+    permissions:
+          pull-requests: write
```
krnowak, line 276: Indentation seems to be off.

t-lo:
@pothos I've added test report merging in a separate job; we will now generate a single test report only, with amd64 and arm64 tests merged in (both qemu_uefi and qemu_update). I've kicked off a test to provide an example of a merged report, please have a look after it concluded: https://github.com/flatcar/scripts/actions/runs/4563204306

ci-automation/tapfile_helper_lib.sh
```diff
@@ -197,27 +197,167 @@ function tap_failed_tests_for_vendor() {
...
+    local name="$2"
+    local succeded_vendors="$3"
+    local failed_vendors="$4"
+    echo "   ..."
+}
+# --
+
+__tap_finish_test_report() {
+    true
+}
```
krnowak, line 255: `finish_test_report` vfunc seems to be unnecessary, as both `tap` and `md` just return `true`. Are there plans for some other format, where `finish_test_report` would be useful?

krnowak: Well, I guess it will be useful when implementing json format. Nevermind.

krnowak (approved):

ci-automation/tapfile_helper_lib.sh
t-lo: Correct! @gabriel-samfira mentioned a desire for JSON at some point.

.github/workflows/run-kola-tests.yaml
t-lo: Good catch; fixed.

t-lo:
Merged and cherry-picked to flatcar-3033, flatcar-3549, flatcar-3510, flatcar-3374.

PR #701, opened 2023-04-04
/usr/share/oem -> /oem

krnowak:
TODO


sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0020-usr-share-oem-oem.patch
```diff
@@ -0,0 +1,53 @@
...
+diff --git a/internal/distro/distro.go b/internal/distro/distro.go
+index 644c42c4..84e575c7 100644
+--- a/internal/distro/distro.go
++++ b/internal/distro/distro.go
+@@ -33,7 +33,7 @@ var (
+ 	// initramfs directory containing distro-provided base config
+ 	systemConfigDir = "/usr/lib/ignition"
+ 	// initramfs directory to check before retrieving file from OEM partition
+-	oemLookasideDir = "/usr/share/oem"
++	oemLookasideDir = "/oem"
```
pothos, line 47: I think this is the initrd directory that users can provide, therefore, it should stay at the same path.

krnowak: How can a user provide a custom directory into initrd? If I understand correctly, we provide initrd inside our kernel binary. Does this mean that we support custom built kernels with a custom initrd?

pothos: One can provide multiple initrds in the boot loader and they get merged. This is actually what happens in the PXE boot with Flatcar where `flatcar_production_pxe_image.cpio.gz` is an additional initrd that brings the sqashfs image.

krnowak: Thanks, I'll revert this change. I must admit that this may be quite confusing having `/usr/share/oem` in initrd, but `/oem` in rootfs. I think this could be addressed with a documentation, but I dunno if we even have docs about this feature.

krnowak: Reverted.

changelog/changes/2023-05-16-oem-mountpoint.md
```diff
@@ -0,0 +1 @@
+- Moved a mountpoint of the OEM partition from `/usr/share/oem` to `/oem`. `/usr/share/oem` became a symlink to `/oem` for backward compatibility. Despite the move, the initrd images providing files through `/usr/share/oem` should keep using `/usr/share/oem`. The move was done to enable activating the OEM sysext images that are placed in the OEM partition.
```
pothos, line 1:
```suggestion
- Moved the mountpoint of the OEM partition from `/usr/share/oem` to `/oem`. `/usr/share/oem` became a symlink to `/oem` for backward compatibility. Despite the move, the initrd images providing files through `/usr/share/oem` should keep using `/usr/share/oem`. The move was done to enable activating the OEM sysext images that are placed in the OEM partition ([scripts#701](https://github.com/flatcar/scripts/pull/701))
```
Not sure about the formatting consistency, maybe a full stop could go in

pothos (approved):

pothos:
I missed this, we should probably add a patch to ignition's mount invocation to use the new path if the old is used. That prevents a breaking change.

sayanchowdhury (approved):

krnowak:
The ignition patch mounting `/oem` when a filesystem for `/usr/share/oem` is specified in Container Linux Config seems to be working. There are still test failures, which are unrelated - `coreos.update.badusr` is veeeeery flaky and EM on arm64 seems to be no better…

pothos (approved):
Thanks!

PR #716, opened 2023-04-18
.github: Move workflow mirroring calico docker packages to mantle

krnowak:
- ~push to flatcar ghcr org by default~
- ~use built-in token for logging in~
- ~make scripts executable~
- ~fixed a typo~

~Test run: https://github.com/flatcar/scripts/actions/runs/4731319782~
~Find the packages in https://github.com/orgs/flatcar/packages~

Moved to mantle, see https://github.com/flatcar/mantle/pull/425


gabriel-samfira:
You might want to limit permissions for that token. See:

https://github.com/containerd/containerd/blob/main/.github/workflows/build-test-images.yml#L33

Default permissions may be too broad.

krnowak:
> You might want to limit permissions for that token. See:
> 
> https://github.com/containerd/containerd/blob/main/.github/workflows/build-test-images.yml#L33
> 
> Default permissions may be too broad.

Good idea to limit them, thanks! I'll test it.

.github/workflows/mirror-calico.yaml
```diff
@@ -4,9 +4,18 @@ on:
     # run every 12h
     - cron:  '0 */12 * * *'
```
krnowak, line 5: One thing I'm wondering about is if these actions should really be called twice a day. Maybe once a week would be enough?

krnowak: I probably miss something, but wouldn't the test just grab whatever is the latest docker package available in flatcar GHCR? Or is there somewhere a code that is sensitive to version mismatches between our calico docker images and actual newest upstream release?

Now I'm wondering if those sync actions should rather reside in mantle repo. We could merge calico docker image sync together with the calico manifest updates.

krnowak: Hm, I did a bit of digging here and I think our issue used to be that we were pulling the tigera operator manifest from a link like `https://docs.projectcalico.org/manifests/tigera-operator.yaml`. By the looks of it, it was always pointing to the latest release of calico. So there was certainly a window where the just released tigera-operator was trying to pull in newer versions of calico images that weren't yet mirrorred. Now, we are using a versioned link (like `https://raw.githubusercontent.com/projectcalico/calico/v3.24.0/manifests/tigera-operator.yaml`), so I think that as long as the version in the link and the version of our mirrored docker images are in sync, we should be safe. And we probably could then just run the action once a week.

Anyway, I think I'll move these actions to mantle. They are not related to anything in the scripts repo.

PR #738, opened 2023-04-24
x11-drivers/nvidia-drivers: Updates to 525.105.17

sayanchowdhury:
# x11-drivers/nvidia-drivers: Updates to 525.105.17

Fixes https://github.com/flatcar/Flatcar/issues/1006

## Testing done

CI passed: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1688/cldsv/


krnowak (commented):
Should the changelog entry have some notes about dropping support for some old nvidia gpus?

sayanchowdhury:
> Should the changelog entry have some notes about dropping support for some old nvidia gpus?

Good Point. I will add a deprecation note when drafting the release note.

PR #740, opened 2023-04-24
.github: Fix changelog generation

krnowak:


pothos (commented):
How about adding `set -euo pipefail` after each `run: |` to get some visibility of these things failing?

krnowak:
There was no command failure, because the changelog generation function checks if `changelog/updates` directory exists and if it does not, the function does nothing. Also, I think that github runs the shell scripts in `run` with `-e` by default - if you click on the triangle in the `Apply patch` step, it will expand and show some information, like `shell: /usr/bin/bash -e {0}`.

krnowak:
Seems to be working now: https://github.com/krnowak/scripts/pull/31/files

dongsupark (approved):

PR #760, opened 2023-04-27
ci-automation: Properly escape parameters passed to bash

krnowak:
Forwarding parameters to another bash invocation through a string interpreted as a bash script is a bit troublesome. It is not enough to wrap a parameter like 'foo bar' in escaped double quotes (\") to avoid it being split into two parameters by bash executing the script string. It mostly works, but there's always a risk of having a path where this breaks. It's rare

Wrapping into escaped quotes, be them double or single, also won't work for passing an array of parameters, so it's even easier here to trigger globbing or bracket expansion or another unwanted splitting of supposedly one parameter into multiple. Globbing can be temporarily disabled with 'set -f' or 'set -o noglob', but this still leaves all the other special bash characters unescaped. So each parameter in the array should be escaped before they are put into the script string.

The escaping can be done with `printf` and its `%q` formatter, so let's do so. For single parameters it is as simple as `foo_escaped=$(printf '%q' "${foo}")`, for arrays a loop needs to be used.

### DIFF
--- a/ci-automation/test.sh
+++ b/ci-automation/test.sh
@@ -114,6 +114,21 @@ function __prepare_torcx() {
 }
 # --
 
+function __escape_multiple() {
+    local out_array_arg_name="${1}"; shift
+    # rest are args to be escape and appended into the array named
+    # after the first arg
+    local -n out_array_arg_ref="${out_array_arg_name}"
+    local arg arg_escaped
+
+    out_array_arg_ref=()
+    for arg; do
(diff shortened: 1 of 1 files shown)


pothos (approved):
Thanks. There are also other ways to do this, like passing environment variables and switching from variable length arguments to a single argument that has a magical separation character. Yet, since cases like this come up often it's also good to have it done once in a robust way for reference.

PR #774, opened 2023-05-02
tapfile_helper: fix foldable details output

t-lo:
This change adds a "<summary>" sub-block to the "<details>" block in markdown test reports.

The test run details output, which is used to print debug and error messages of test runs in case of test failures, was meant to be fold-able, aiding readability of test reports. This is implemented using the "<details>" feature. However, we forgot to mark a "<summary>" line within the "<details>" block, leading to the blocks not being fold-able but instead being visible all the time.

The change should be backported to all active maintenance branches in order to fix markdown test reports there, too.


t-lo:
One more change as the previous version omitted a vital tag from debug output of transient failures.

PR #776, opened 2023-05-02
Move bash to portage stable and update it

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/796/cldsv/

Bash had a customization in bashrc that added aliases for some of the bcc tools - I moved that to a separate package (coreos-base/bashrcs). With that done, bash package could be moved to portage-stable. Since we want to update to 5.2 to address a CVE issue, I also added accept keywords for bash.

--

- app-shells/bash: (PROD) (DEV)
  - from 5.1_p16-r2 to 5.2_p15-r2
  - moved from overlay to portage-stable
  - still unstable, so added accept keywords to overlay
  - license change from gpl3 to gpl3+
  - new IUSE pgo, we don't use it
  - bumped readline dep to 8.2_p1
  - fixes CVE-2022-3715
  - release notes: https://lists.gnu.org/archive/html/bash-announce/2022-09/msg00000.html

- sys-libs/readline: (PROD) (DEV)
  - from 8.1_p2 to 8.2_p1
  - update needed by updated app-shells/bash
  - still unstable, so added accept keywords to overlay
  - release notes: https://lists.gnu.org/archive/html/info-gnu/2022-09/msg00013.html

- sys-devel/gdb: (DEV)
  - from 11.2 to 13.1.90_p20230325
  - EAPI 8
  - moved from overlay to portage-stable
    - the new ebuild has a fix for cross-build
  - updated to fix build with sys-libs/readline 8.2
  - new IUSE flags: sim and zstd, disabled
  - release notes: https://lwn.net/Articles/923819/

--


pothos:
What about the modifications in 55a14648c5a4c3557f2462f2a6e1ab03bb6caa2b? From a quick look it would be that `/usr/share/bash` and `/usr/share/skel` disappear? Do these get referenced by files under `/etc` for old machines meaning we would need a symlink?

pothos:
> What about the modifications in [55a1464](https://github.com/flatcar/scripts/commit/55a14648c5a4c3557f2462f2a6e1ab03bb6caa2b)? From a quick look it would be that `/usr/share/bash` and `/usr/share/skel` disappear? Do these get referenced by files under `/etc` for old machines meaning we would need a symlink?

What we could do is to set up symlinks from `/usr/share/bash/` to `/usr/share/flatcar/etc/bash` and from `/usr/share/skel/` to `/usr/share/flatcar/etc/skel`. This ensure that when `/etc/skel/` exist on update and has the old symlinks to `/usr/share/skel/…`, these won't become dangling links.
Note that there is migration logic for the `/etc` overlay but that's on best-effort and we can't guarantee that this works because the user may have modifications that prevent the migration.

krnowak:
> > What about the modifications in [55a1464](https://github.com/flatcar/scripts/commit/55a14648c5a4c3557f2462f2a6e1ab03bb6caa2b)? From a quick look it would be that `/usr/share/bash` and `/usr/share/skel` disappear? Do these get referenced by files under `/etc` for old machines meaning we would need a symlink?
> 
> What we could do is to set up symlinks from `/usr/share/bash/` to `/usr/share/flatcar/etc/bash` and from `/usr/share/skel/` to `/usr/share/flatcar/etc/skel`. This ensure that when `/etc/skel/` exist on update and has the old symlinks to `/usr/share/skel/…`, these won't become dangling links. Note that there is migration logic for the `/etc` overlay but that's on best-effort and we can't guarantee that this works because the user may have modifications that prevent the migration.

Yeah, installing symlinks is what I was thinking to do. I'll add another package (`coreos-base/misc-files` or something) that will be installing those. I think I'll put the bashrc also in this package. I need to do it anyway, because I just got a build failure:

```
09:15:45  WARNING build_packages: test_image_content: Failed #! check
09:15:48  broken link: b'/home/core/.bash_profile' -> b'../../usr/share/skel/.bash_profile'
09:15:48  broken link: b'/home/core/.bashrc' -> b'../../usr/share/skel/.bashrc'
09:15:48  broken link: b'/home/core/.bash_logout' -> b'../../usr/share/skel/.bash_logout'
```

pothos:
Nice that this was caught, didn't think about home

sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/misc-files-0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+# No source directory.
+S="${WORKDIR}"
+
+# Versions listed below are version of packages that shedded the
+# modifications in their ebuilds.
+RDEPEND="
+	>=app-shells/bash-5.2_p15-r2
+"
+
+# Use absolute paths to be clear about what locations are used. The
```
pothos, line 22: We also have similar things in `coreos-base/coreos-init`. Maybe we could put everything there instead of creating a new package, but as you like.

sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/misc-files-0.ebuild
```diff
@@ -0,0 +1,56 @@
...
+# dosym below will make relative paths out of them.
+#
+# NOTE: The symlinks under /usr that point to entries under /etc will
+# be retargeted to /usr/share/flatcar/etc when building an image
+# later. Do not make them point to /usr/share/flatcar/etc here,
+# because this directory does not exist now and won't exist at all
+# during the whole packages build phase. We don't want to create
+# dangling symlinks because the packages build phase also runs a root
+# check at the end, and it will bail out on dangling symlinks.
+COMPAT_SYMLINKS=(
```
pothos, line 32: Not sure if emerge supports bash 4 features, if yes it could use https://stackoverflow.com/a/3467959

pothos, line 32:
```suggestion
# These links exist because old installations can still have references to `/usr/share/(bash|skel)`
# before these were moved back to the /etc through the overlay.
COMPAT_SYMLINKS=(
```

krnowak: [EAPI 8 supports bash 5](https://devmanual.gentoo.org/ebuild-writing/eapi/index.html#eapi-8-ebuild-format), so yes, we could use a hashmap. Good idea, I'll switch to it.

sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/misc-files-0.ebuild
krnowak: It's a bit of a mess, you know - all this compat symlinks stuff is spread over several places (baselayout Makefile, baselayout ebuild, init Makefile, init ebuild, probably more). I'll have a look where I could add those instead of creating yet another place where we install random files and symlinks.

krnowak: I decided to keep the `coreos-base/misc-files` package, as `coreos-base/coreos-init` should be about systemd stuff like units. Compat symlinks and other extra files could go to the misc-files package. I think it will evolve as we will keep moving more and more packages from overlay back to portage-stable.

sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/misc-files-0.ebuild
krnowak: Added.

krnowak:
Needs https://github.com/flatcar/baselayout/pull/31.

build_library/build_image_util.sh
```diff
@@ -836,9 +836,16 @@ EOF
     sudo setfiles -Dv -r "${root_fs_dir}" "${root_fs_dir}"/etc/selinux/mcs/contexts/files/file_contexts "${root_fs_dir}"/etc
   fi
 
-  # Backup the /etc contents to /usr/share/flatcar/etc to serve as source
-  # for creating missing files
+  # Backup the /etc contents to /usr/share/flatcar/etc to serve as
+  # source for creating missing files. Make sure that the preexisting
+  # /usr/share/flatcar/etc does not have any meaningful (non-empty)
+  # files, so we remove nothing important.
+  if [[ $(sudo find "${root_fs_dir}/usr/share/flatcar/etc" -size -1 2>/dev/null) -gt 0 ]]; then
```
dongsupark: I am not sure how that works.
Should we probably add `| wc -l` at the end of find, to get a number of lines?

krnowak: Meh, there actually used to be `| wc -l`. But I updated it and must have removed that by mistake… Thanks for spotting, will fix it.

krnowak: I kicked off another image job run just to make sure that nothing breaks: http://jenkins.infra.kinvolk.io:8080/job/container/job/image/2698/cldsv/

krnowak: Image job passed.

pothos:
Bringing this up because after booting I don't see bash files in the core home folder.

krnowak:
> `/usr/lib/tmpfiles.d/base_image_var_late.conf` is empty, probably due to the workaround to prevent upcopies - where again was its content defined (a search in this PR diff failed me)?

That was done for one of the weekly updates PRs, and the change was in https://github.com/flatcar/scripts/pull/742. But it shouldn't be empty, it should have one entry. I'll have a look.

krnowak:
> Bringing this up because after booting I don't see bash files in the core home folder. Edit: To make it clear: I've downloaded it from the linked Jenkins build.

I'll check it out.

krnowak:
> Bringing this up because after booting I don't see bash files in the core home folder. Edit: To make it clear: I've downloaded it from the linked Jenkins build.

`/build/amd64-usr/var/tmp/portage/coreos-base/misc-files-0/temp/environment: line 859: dotmpfiles: command not found` and the build succeeded? Ok.

krnowak:
> `/usr/lib/tmpfiles.d/base_image_var_late.conf` is empty, probably due to the workaround to prevent upcopies - where again was its content defined (a search in this PR diff failed me)?

Alright, I got why it's empty now. During one of the weekly updates, I needed to add `dev-lang/lua` to the production image, which pulled in `app-eselect/eselect-lua` as a dependency, which in turn pulled `app-admin/eselect` into the production image. And that package was installing `/var/lib/gentoo/news` that had ownership `root:portage`. Which is why I ended up splitting the `base_image_var.conf` file into two config files, because `portage` group wasn't available in the early stages of booting. But then later I thought that having eselect stuff in production images made no sense, so [I made some changes to avoid having it here](https://github.com/flatcar/scripts/commit/7eec9c7e5c2219bc7d51655e5ddcf302692f0922). And thus the `/var/lib/gentoo/news` (and `/var/lib/gentoo` too) stopped being a part of the production image again, which results in an empty `base_image_var_late.conf` file. Which means that the splitting was probably unnecessary at that point, had I realized sooner that it was eselect packages that were installing the `var/lib/gentoo/news` directory. I'd keep split in case some other package update causes a directory with a different ownership to be installed under `/var`.

krnowak:
Seems to be working - symlinks are there, iovisor aliases work.

pothos:
Where is the `home-core-bash-symlinks.conf` file defined?

pothos:
> Where is the `home-core-bash-symlinks.conf` file defined?

Ah, found it now, it's generated :D

pothos (approved):

PR #784, opened 2023-05-03
.github: check remote branch, add correct changelog

dongsupark:
To avoid noise of touching existing PRs, check first if the remote branch already exists. If that exists, skip creating or updating the PR.

In case of a draft PR created first, no CI gets triggered at the time. So we should trigger CI afterwards when it is being set to a ready_for_review state.

Changelog directory is now located under flatcar scripts, not under coreos-overlay. Fix the location to add correct changelog to the git commit.

## Testing done

Tested via my fork repo

### DIFF
--- a/.github/workflows/cacerts-apply-patch.sh
+++ b/.github/workflows/cacerts-apply-patch.sh
@@ -6,13 +6,18 @@ source "${GHA_SCRIPTS_DIR}/.github/workflows/common.sh"
 
 prepare_git_repo
 
+if ! check_remote_branch "cacerts-${VERSION_NEW}-${TARGET_BRANCH}"; then
+    echo "remote branch already exists, nothing to do"
+    exit 0
+fi
+
 pushd "${SDK_OUTER_OVERLAY}"
 
 # Parse the Manifest file for already present source files and keep the latest version in the current series
 VERSION_OLD=$(sed -n "s/^DIST nss-\([0-9]*\.[0-9]*\).*$/\1/p" app-misc/ca-certificates/Manifest | sort -ruV | head -n1)
 if [[ "${VERSION_NEW}" = "${VERSION_OLD}" ]]; then
-  echo "already the latest ca-certificates, nothing to do"
-  exit 0
+    echo "already the latest ca-certificates, nothing to do"
+    exit 0
 fi
 
 EBUILD_FILENAME=$(get_ebuild_filename app-misc/ca-certificates "${VERSION_OLD}")
--- a/.github/workflows/cacerts-release.yaml
+++ b/.github/workflows/cacerts-release.yaml
@@ -53,6 +53,7 @@ jobs:
           VERSION_NEW: ${{ steps.nss-latest-release.outputs.NSS_VERSION }}
           PACKAGES_CONTAINER: ${{ steps.setup-flatcar-sdk.outputs.PACKAGES_CONTAINER }}
           SDK_NAME: ${{ steps.setup-flatcar-sdk.outputs.SDK_NAME }}
+          TARGET_BRANCH: ${{ steps.figure-out-branch.outputs.BRANCH }}
         run: gha/.github/workflows/cacerts-apply-patch.sh
       - name: Create pull request
         if: (steps.figure-out-branch.outputs.SKIP == 0) && (steps.apply-patch.outputs.UPDATE_NEEDED == 1)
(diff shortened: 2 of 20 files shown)


.github/workflows/cacerts-release.yaml
```diff
@@ -62,6 +63,8 @@ jobs:
           path: work
           branch: "cacerts-${{ steps.nss-latest-release.outputs.NSS_VERSION }}-${{ steps.figure-out-branch.outputs.BRANCH }}"
           base: ${{ steps.figure-out-branch.outputs.BRANCH }}
+          author: "Flatcar Buildbot <buildbot@flatcar-linux.org>"
+          committer: "Flatcar Buildbot <buildbot@flatcar-linux.org>"
```
krnowak, line 67: I'd argue that if git is not clean, we should fail the workflow instead of papering it over by creating a generic commit. If this happens, then it means that the action is not invoking `cleanup_repo` at the end of the script.

.github/workflows/cacerts-apply-patch.sh
```diff
@@ -6,13 +6,19 @@ source "${GHA_SCRIPTS_DIR}/.github/workflows/common.sh"
 
 prepare_git_repo
 
+if ! check_remote_branch "cacerts-${VERSION_NEW}-${TARGET_BRANCH}"; then
+    echo 'UPDATE_NEEDED=0' >>"${GITHUB_OUTPUT}"
```
krnowak, line 10: I'd say that setting UPDATE_NEEDED to 0 is not necessary, since we always check if UPDATE_NEEDED is 1.

.github/workflows/common.sh
```diff
@@ -164,12 +173,17 @@ function commit_changes() {
 
   regenerate_manifest "${pkg}" "${new_version}"
 
-  pushd "${SDK_OUTER_OVERLAY}"
+  pushd "${SDK_OUTER_TOPDIR}"
```
krnowak, line 176: Bah, well spotted. Thanks for fixing it.

.github/workflows/go-apply-patch.sh
```diff
@@ -9,6 +9,11 @@ prepare_git_repo
 # create a mapping between short version and new version, e.g. 1.16 -> 1.16.3
 declare -A VERSIONS
 for version_new in ${VERSIONS_NEW}; do
+  if ! check_remote_branch "go-${version_new}-${TARGET_BRANCH}"; then
+      echo 'UPDATE_NEEDED=0' >>"${GITHUB_OUTPUT}"
+      exit 0
+  fi
```
krnowak, line 15: This won't work. You want to add this check after `branch_name` is set below this loop, so you could do `if check_remote_branch "${branch_name}"; …`.

.github/workflows/common.sh
```diff
@@ -64,6 +64,15 @@ function prepare_git_repo() {
   git -C "${SDK_OUTER_TOPDIR}" config user.email "${BUILDBOT_USEREMAIL}"
 }
 
+function check_remote_branch() {
+  local TARGET_BRANCH="${1}"
+
+  if git -C "${SDK_OUTER_TOPDIR}" show-ref "remotes/origin/${TARGET_BRANCH}"; then
+    echo "Target branch already exists. exit."
```
krnowak, line 71:
```suggestion
    echo "Target branch already exists, exit."
```

.github/workflows/cacerts-release.yaml
dongsupark: Ok, I have dropped the last commit of author/committer.

BTW have just realized we have another git path issue in `cleanup_repo`, so fixed that as well.

.github/workflows/go-apply-patch.sh
dongsupark: Fixed, thanks.

krnowak (approved):
Looks good, one nitpick.

.github/workflows/common.sh
```diff
@@ -64,6 +64,15 @@ function prepare_git_repo() {
   git -C "${SDK_OUTER_TOPDIR}" config user.email "${BUILDBOT_USEREMAIL}"
 }
 
+function check_remote_branch() {
+  local TARGET_BRANCH="${1}"
+
+  if git -C "${SDK_OUTER_TOPDIR}" show-ref "remotes/origin/${TARGET_BRANCH}"; then
+    echo "Target branch already exists, exit."
+    return 1
+  fi
```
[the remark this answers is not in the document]
dongsupark: Done

PR #790, opened 2023-05-04
.github: check for remote branches with git ls-remote

dongsupark:
`git show-ref` in `check_remote_branch` is not working at all, because the GitHub Actions checkout fetches a git repo with only depth 1 by default. That means only main branch is available, not any other remote branches. As a result, `git show-ref` always fails.

Fix that by checking for remote branches by `git ls-remote`.

Fixes bugs resulted from https://github.com/flatcar/scripts/pull/784.

## Testing done

tested via my fork

### DIFF
--- a/.github/workflows/common.sh
+++ b/.github/workflows/common.sh
@@ -66,8 +66,7 @@ function prepare_git_repo() {
 
 function check_remote_branch() {
   local target_branch="${1}"
-
-  if git -C "${SDK_OUTER_TOPDIR}" show-ref "remotes/origin/${target_branch}"; then
+  if git -C "${SDK_OUTER_TOPDIR}" ls-remote --refs --heads --exit-code origin "${target_branch}" >/dev/null; then
     return 1
   fi
   return 0


t-lo:
I think I ran into something similar with `ci.yaml`; if an action is triggered from a PR it checks out the merge commit instead of the tip of the PR branch. In that workflow we used an alternative fix: https://github.com/flatcar/scripts/blob/main/.github/workflows/ci.yaml#L64
(not saying you should do the same but posting this for visibility).

dongsupark:
Thanks for the link. That is interesting.

However, that looks slightly different from the issue this PR tries to fix.
In case of this PR, when `check_remote_branch` runs, the PR's head commit does not exist yet. Before creating the PR, it has to first fetch remote branches, and then do `git show-ref` to reduce the noise.

.github/workflows/common.sh
```diff
@@ -67,6 +67,11 @@ function prepare_git_repo() {
 function check_remote_branch() {
   local target_branch="${1}"
 
+  # It is needed to first fetch remote branches before checking for remote
```
krnowak, line 70: You could use `git -C "${SDK_OUTER_TOPDIR}" ls-remote --refs --heads --exit-code origin "${target_branch}" >/dev/null`, so this function could be:

```bash
function check_remote_branch() {
    local target_branch="${1}"
    # this makes "set +e" below to be effective only within the function scope
    local -
    set +e
    git -C "${SDK_OUTER_TOPDIR}" ls-remote --refs --heads --exit-code origin "${target_branch}" >/dev/null
}
```

dongsupark: Yes, that could work as well.
In that case, however, we have to update all call sites of `check_remote_branch` like `if check_remote_branch branch_name; then error`.
Is that what you want?

krnowak: Oh. Right. Too many reverses just ended up confusing me. I'd avoid the rewrites, so:

```bash
# Returns zero if the passed branch was not found on the remote server, otherwise returns non-zero.
function check_remote_branch() {
    local target_branch="${1}"
    if git -C "${SDK_OUTER_TOPDIR}" ls-remote --refs --heads --exit-code origin "${target_branch}" >/dev/null; then
        return 1
    fi
    return 0
}
```

dongsupark: Done. Thanks

krnowak (approved):

PR #792, opened 2023-05-06
.github Add faux workflow for PR

mkulke:
n/a


pothos:
FYI, as repo setting under "Fork pull request workflows from outside collaborators" we have "Require approval for all outside collaborators", that's why nothing happened

PR #799, opened 2023-05-08
ci-automation: Follow-up fix for 2-phase nightly SDK build tags

krnowak:
We push a commit with the nightly SDK tag to the main branch if the SDK was built from the main branch. Which is what happens when we build the nightly intermediate SDK. The final nightly SDK is not built from the main branch, but rather from the nightly intermediate SDK tag. Both of them point to the exactly same commit, but the difference is in what `git rev-parse --abbrev-ref HEAD` returns for each of those. When the main branch is checked out, the command will return "main". When the nightly intermediate SDK tag is checked out, the command will return "HEAD". So when nightly final SDK is being built, the command returns a string different than "main" and thus decides not to push the commit with the final nightly SDK tag to the main branch. Rework it to assume that if `git rev-parse HEAD` and `git rev-parse main` return the same commit hash (and it's the nightly build and all that) then the commit should be pushed.

### DIFF
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -23,7 +23,7 @@ function check_version_string() {
 
 function update_and_push_version() {
     local version="$1"
-    local push_to_branch="${2:-false}"
+    local target_branch="${2:-}"
 
     # set up author and email so git does not complain when tagging
     if ! git config --get user.name >/dev/null 2>&1 ; then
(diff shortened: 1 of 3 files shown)


dongsupark:
As GitHub Actions cannot run SDK builds yet, can you please start Jenkins SDK builds?

dongsupark:
Moved to backlog

krnowak:
Tested it with my branch krnowak/fake-main (where I actually skip building stuff and focus on the tagging issue) and with some jenkins modifications to avoid running irrelevant jobs.

See: http://jenkins.infra.kinvolk.io:8080/job/container/job/nightly/429/cldsv/

dongsupark (approved):

pothos:
Backported to all branches

PR #800, opened 2023-05-08
Trigger build workflow via PR comment instead of PR state; add a workflow to build an updated SDK container, 

t-lo:
This change adds a github actions workflow to trigger SDK and OS image builds from PR comments. It verifies whether the commenter is a member of the [Flatcar Maintainers](https://github.com/orgs/flatcar/teams/flatcar-maintainers) team before commencing the build.

The PR comment commands are:
- `/build-sdk` - builds SDK and OS image, runs tests
- `/build-image` - builds OS image and runs tests

The previous semantics of triggering builds from PR changes ("ready to review", "synchronised") is removed. **This implies builds now need to be explicitly started using one of the above commands.**

Further, a new workflow is added to create a new SDK container based on an existing SDK container. The new container image is published to bincache. This can be used for CI testing intrusive changes that also affect the SDK without bootstrapping a whole new SDK.

A garbage collector is added to `ci-automation`  for cleaning up SDK containers on bincache.

# Testing done

- Maintainer team membership with hard-coded "commenter" names
  - pass: user `t-lo` - https://github.com/flatcar/scripts/actions/runs/4916044777
  - fail: user `mallory` - https://github.com/flatcar/scripts/actions/runs/4915901180

Further testing is challenging before we merge since the corresponding events (PR comment, PR-has-changed) are only emitted to workflows in the "main" branch.

### DIFF
--- a/.github/workflows/pr-comment-build-dispatcher.yaml
+++ b/.github/workflows/pr-comment-build-dispatcher.yaml
@@ -1,0 +1,57 @@
+name: "Dispatch SDK container rebuilds, OS image builds, and OS image tests from PR comments"
+on:
+  issue_comment:
+    types: [created]
+  pull_request:
+    # This is temporary for testing the workflow.
+    #  Comment events are only processed for workflows in the main branch
+    types: [opened, reopened, synchronize]
+
+permissions:
+  pull-requests: write
+
+jobs:
+
+  check_maintainer_membership:
+    # Only run if this is a PR comment that contains a valid command
+#    if: |
+#      ${{ github.event.issue.pull_request }} &&
+#        (    contains(github.event.comment.body, '/build-sdk')
+#          || contains(github.event.comment.body, '/build-image') )
+    name: Check if commenter is in the Flatcar maintainers team
+    outputs:
+      maintainers: steps.step1.output.maintainers
+    runs-on:
+      - ubuntu-latest
+    steps:
+      - name: Fetch members of the maintainers team
+        env:
+#          - requester: ${{ github.event.comment.user.login }}
+          requester: "t-lo"
+        shell: bash
+        run: |
+          curl -L --silent \
+                  -H "Accept: application/vnd.github+json" \
+                  -H "Authorization: Bearer ${{ secrets.GH_ACTIONS_ORG_READ }}" \
+                  -H "X-GitHub-Api-Version: 2022-11-28" \
+                  https://api.github.com/orgs/flatcar/teams/flatcar-maintainers/members \
+              | jq -r '.[].login' > maintainers.txt
+
+          echo "Current members of the maintainers team:"
+          cat maintainers.txt
+
+          res=false
+          echo "Checking for membership of '${{ env.requester }}'"
+          if grep -qE "^${{ env.requester }}$" maintainers.txt ; then
+            echo "Succeeded."
+            res=true
+          else
+            echo "FAILED: '${{ env.requester }} is not a member of the Flatcar maintainers team."
+          fi
+
+          $res
+
+#    name: "Build the OS image"
+#    uses: ./.github/workflows/ci.yaml
+#    with:
+#      custom_sdk_version: ${{ github.event.inputs.custom_sdk_version }}
--- a/.github/workflows/pr-comment-build-dispatcher.yaml
+++ b/.github/workflows/pr-comment-build-dispatcher.yaml
@@ -1,23 +1,22 @@
-name: "Dispatch SDK container rebuilds, OS image builds, and OS image tests from PR comments"
+name: "PR command build dispatcher"
 on:
   issue_comment:
     types: [created]
-  pull_request:
-    # This is temporary for testing the workflow.
-    #  Comment events are only processed for workflows in the main branch
-    types: [opened, reopened, synchronize]
 
 permissions:
   pull-requests: write
 
-jobs:
+concurrency:
+  group: ${{ github.workflow }}-pr-command-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
 
+jobs:
   check_maintainer_membership:
     # Only run if this is a PR comment that contains a valid command
-#    if: |
-#      ${{ github.event.issue.pull_request }} &&
-#        (    contains(github.event.comment.body, '/build-sdk')
-#          || contains(github.event.comment.body, '/build-image') )
+    if: |
+      ${{ github.event.issue.pull_request }} &&
+        (    contains(github.event.comment.body, '/build-sdk')
+          || contains(github.event.comment.body, '/build-image') )
     name: Check if commenter is in the Flatcar maintainers team
     outputs:
       maintainers: steps.step1.output.maintainers
@@ -26,8 +25,7 @@ jobs:
     steps:
       - name: Fetch members of the maintainers team
         env:
-#          - requester: ${{ github.event.comment.user.login }}
-          requester: "t-lo"
+          requester: ${{ github.event.comment.user.login }}
         shell: bash
         run: |
           curl -L --silent \
@@ -51,7 +49,22 @@ jobs:
 
           $res
 
-#    name: "Build the OS image"
-#    uses: ./.github/workflows/ci.yaml
-#    with:
-#      custom_sdk_version: ${{ github.event.inputs.custom_sdk_version }}
+  build_sdk:
+    needs: check_maintainer_membership
+    if: ( always() && needs.check_maintainer_membership.result == 'success'
+                   && contains(github.event.comment.body, '/build-sdk') )
+    name: "Build the SDK container"
+    # SDK build needs access to bincache ssh secret
+    secrets: inherit
+    uses: ./.github/workflows/update-sdk.yaml
+
+  build_image:
+    needs: [ check_maintainer_membership, build_sdk ]
+    if: ( always() && needs.check_maintainer_membership.result == 'success'
+                   && (    contains(github.event.comment.body, '/build-image')
+                        || needs.build_sdk.result == 'success' ) )
+    name: "Build the OS image"
+    uses: ./.github/workflows/ci.yaml
+    with:
+      custom_sdk_version: ${{ needs.update_sdk.outputs.sdk_version }}
+      image_formats: qemu_uefi
--- a/.github/workflows/pr-comment-build-dispatcher.yaml
+++ b/.github/workflows/pr-comment-build-dispatcher.yaml
@@ -15,7 +15,7 @@ jobs:
     # Only run if this is a PR comment that contains a valid command
     if: |
       ${{ github.event.issue.pull_request }} &&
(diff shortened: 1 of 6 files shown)


t-lo:
Rebased on latest main.

.github/workflows/pr-comment-build-dispatcher.yaml
```diff
@@ -0,0 +1,70 @@
...
+    name: "Build the SDK container"
+    # SDK build needs access to bincache ssh secret
+    secrets: inherit
+    uses: ./.github/workflows/update-sdk.yaml
+
+  build_image:
+    needs: [ check_maintainer_membership, build_sdk ]
+    if: ( always() && needs.check_maintainer_membership.result == 'success'
+                   && (    contains(github.event.comment.body, '/build-image') 
+                        || needs.build_sdk.result == 'success' )
```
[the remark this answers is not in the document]
t-lo: Good catch! Will fix.

t-lo:
Note: Test results can be ignored because some parts of the workflows will only work after merge to main. This is caused by Github only emitting certain action events to workflows in the main repo.

pothos:
I don't understand why we would need to publish a container image. The SDK container could be built as part of `ci.yaml` and doesn't need to be uploaded and downloaded.
Since it doesn't bootstrap a new SDK but only updates packages, this is fast - and I wonder if we even need to build this new container image, I think we should rather call `./update_chroot --toolchain_boards="amd64-usr arm64-usr"` and `./setup_board --board="arm64-usr" --regen_configs && ./setup_board --board="amd64-usr" --regen_configs` as optional preparation step in `ci.yaml`.

pothos:
> We actually perform the above steps [every time the SDK container is entered](https://github.com/flatcar/scripts/blob/main/sdk_lib/sdk_entry.sh#L40-L43). So in that case, one would simply go with `/build-image`.

Maybe `update_chroot` does the difference or that some directory is not there in the container image creation? We could also do these steps in `ci.yaml` (optional if they slow the build down). I don't understand the use case of building a new SDK this way; if it would be fully bootstrapped, then yes.

t-lo:
> > We actually perform the above steps [every time the SDK container is entered](https://github.com/flatcar/scripts/blob/main/sdk_lib/sdk_entry.sh#L40-L43). So in that case, one would simply go with `/build-image`.
> 
> Maybe `update_chroot` does the difference or that some directory is not there in the container image creation? We could also do these steps in `ci.yaml` (optional if they slow the build down). I don't understand the use case of building a new SDK this way; if it would be fully bootstrapped, then yes.

The only reason why we would do this step is to explicitly produce a new SDK container image for later re-use (manual testing, bugfixing, reproduction of issues) with the changes proposed by the PR. As stated, we're doing the same *transiently* in the package build. The only benefit of calling `/build-sdk` instead of `/build-image` is to generate a separate artifact. Which might be useful for debugging downstream build issues (think auto-cut toolchain update PRs etc.).

Re: SDK bootstrapping, that's a good point - we can consider adding this at a later point.

pothos:
For me the benefit and use case is a bit too narrow compared to running this command on the nightly SDK when needed. As high level action, the full bootstrap would be a better fit. This doesn't have to be done now but later I would like to see that `/build-sdk` does a full bootstrap because it's the more common thing to do and covers this special case, too, with a bit more waiting. As optimization this reusing of an existing SDK makes sense and could be offered through a special parameter or new command, similar to how we could offer a two-phase SDK build.

t-lo:
Makes sense - I'll change the command to `/update-sdk` for the time being so we can add `/build-sdk` later.

t-lo:
Done. Coincidentally the actual workflow file's name already was `update-sdk.yaml`.

.github/workflows/pr-comment-build-dispatcher.yaml
```diff
@@ -0,0 +1,70 @@
...
+      maintainers: steps.step1.output.maintainers
+    runs-on:
+      - ubuntu-latest
+    steps:
+      - name: Fetch members of the maintainers team
+        env:
+          requester: ${{ github.event.comment.user.login }}
+        shell: bash
+        run: |
+          curl -L --silent \
```
pothos, line 31:
```suggestion
          set -euo pipefail
          curl --fail --show-error -L --silent \
```
When the API fails, we can exit with an error instead of having an empty maintainers file

.github/workflows/update-sdk.yaml
```diff
@@ -0,0 +1,145 @@
...
+      - name: Upload the SDK container and binary packages to bincache
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          source ci-automation/ci_automation_common.sh
+
+          mkdir -p ~/.ssh
+          trap 'rm -f ~/.ssh/bincache' EXIT
+          echo "${{ secrets.BINCACHESSH }}" > ~/.ssh/bincache
```
pothos, line 130: There's some risk attached as GitHub Actions could be misconfigured to allow extraction when it runs without manual approval required. Another option would be to save this as regular CI artifact. What do you think?

pothos: (Plus we don't need to think about extra garbage collection)

t-lo: I understand the concern and I realise it's a trade-off.
My concerns are that using github artifacts will make the overall process a lot slower - a lot of time is already spent on copying back and forth image build artifacts. Bincache access on the other hand is  significantly faster - often by a factor of 10 or higher.
Also, it's not unlikely that - even though officially Artifacts storage is unlimited for public repos - we'd eventually create friction with Github since we'd potentially produce hundreds of gigabytes of artifacts per week.

pothos: Ok, maybe I don't know all plans, I thought that building an SDK is a rare even and while bigger than the regular images, it's still not that much of a difference.

t-lo: At the end of the day this is just one building block for fully automating merging of auto-generated PRs.
I think quite a few of the PRs that we're generating currently - kernel, rust, bulk Gentoo packages refreshes - would mandate an SDK update if we really want to test these fully automatedly.

But you're right with your concern, and one way around it would be to run the SDK container update as an optional step in the main `ci.yaml`. I'll think through it.

pothos: Adding support for the SDK tar ball as GitHub Action artifact would be interesting security-wise but I think we can merge it as is and maybe try this out in a follow-up (I don't know how easy that would be).

pothos (approved):

dongsupark (approved):

ci-automation/garbage_collect_github_ci_sdk.sh
```diff
@@ -0,0 +1,99 @@
...
+    # 3. remove the "/"
+   local versions_sorted="$(echo "${versions_detected}" \
+                        | sed 's/\(-github\(-pr-[0-9]*\)*-\)/\1\//' \
+                        | sort -k 2 -t / \
+                        | sed 's:/::')"
+
+    echo "######## Full list of version(s) found ########"
+    echo "${versions_sorted}" | awk '{printf "%5d %s\n", NR, $0}'
+
+    local keep="$((keep + 1))" # for tail -n+...
```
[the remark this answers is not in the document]
krnowak: Although safer would probably be define a new variable (`keep_tail`) as we seem to be using the `keep` variable also for `awk` invocation.

ci-automation/garbage_collect_github_ci_sdk.sh
```diff
@@ -0,0 +1,99 @@
...
+    local sshcmd="$(gen_sshcmd)"
+
+    echo
+    echo "######## The following version(s) will be purged ########"
+    if [ "$dry_run" = "y" ] ; then
+        echo
+        echo "(NOTE this is just a dry run since DRY_RUN=y)"
+        echo
+    fi
+    echo "${purge_versions}" | awk -v keep="${keep}" '{if ($0 == "") next; printf "%5d %s\n", NR + keep - 1, $0}'
```
krnowak, line 72: If you decide to have a variable `keep_tail` above, thus keeping `keep` unmodified, then I think that the ` - 1` could be removed.

ci-automation/garbage_collect_github_ci_sdk.sh
```diff
@@ -0,0 +1,99 @@
...
+                        | sed 's/\(-github\(-pr-[0-9]*\)*-\)/\1\//' \
+                        | sort -k 2 -t / \
+                        | sed 's:/::')"
+
+    echo "######## Full list of version(s) found ########"
+    echo "${versions_sorted}" | awk '{printf "%5d %s\n", NR, $0}'
+
+    local keep="$((keep + 1))" # for tail -n+...
+    local purge_versions="$(echo "${versions_sorted}" \
+                            | tail -n+"${keep}")"
```
krnowak, line 60: I'd say it usually better to keep list of values as arrays, especially when you want to iterate it with a loop later. Unfortunately bash has a quite obtuse way of creating arrays from multiline strings:

```suggestion
    local purge_versions
    mapfile -t purge_versions < <(tail -n+"${keep}" <<<"${versions_sorted}")
```

krnowak: Or, if you'll go with `tail_keep` variable:

```suggestion
    local purge_versions
    mapfile -t purge_versions < <(tail -n+"${tail_keep}" <<<"${versions_sorted}")
```

ci-automation/garbage_collect_github_ci_sdk.sh
krnowak: If `purge_versions` is an array then this would need to be something like:
```suggestion
    printf '%s\n' "${purge_versions[@]}" | awk -v keep="${keep}" '{if ($0 == "") next; printf "%5d %s\n", NR + keep - 1, $0}'
```

ci-automation/garbage_collect_github_ci_sdk.sh
```diff
@@ -0,0 +1,99 @@
...
+        echo
+        echo "(NOTE this is just a dry run since DRY_RUN=y)"
+        echo
+    fi
+    echo "${purge_versions}" | awk -v keep="${keep}" '{if ($0 == "") next; printf "%5d %s\n", NR + keep - 1, $0}'
+    echo
+    echo
+
+    local version=""
+    for version in ${purge_versions}; do
```
krnowak, line 77: If `purge_versions` is an array then this would need to be:

```suggestion
    for version in "${purge_versions[@]}"; do
```

krnowak (commented):
Some nitpicks.

t-lo:
I've updated `garbage_collect_github_ci_sdk.sh` as suggested by @krnowak in https://github.com/flatcar/scripts/pull/800#discussion_r1196572965; we're now using a separate variable `tail_keep` for the `tail` invocation.

krnowak (approved):

PR #806, opened 2023-05-09
Enable elfutils flag for systemd to get coredump stacktraces in journal

pothos:
When encountering https://github.com/systemd/systemd/issues/27491 no debug info is available when the coredump file is not found (e.g., because it's cleaned up). Normally a stracktrace is included in the journal but the ebuild feature is disabled.
Enable the elfutils ebuild feature for systemd to get coredump stacktraces in the system journal.

## How to use

todo: insert link in changelog

## Testing done

todo: check that the feature gets enabled
[build running](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1786/cldsv/)

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/changelog/changes/2023-05-09-systemd-journal-stacktrace.md
+++ b/sdk_container/src/third_party/coreos-overlay/changelog/changes/2023-05-09-systemd-journal-stacktrace.md
@@ -1,0 +1,1 @@
+- Enabled support for coredump stacktraces in the system journal ([]())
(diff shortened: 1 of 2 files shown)


jepio:
https://github.com/flatcar/coreos-overlay/pull/2489 ?

jepio:
It looks like this isn't in stable yet, but is in beta-3549.

pothos:
Yes, thanks, I've checked and Beta has it enabled (`/usr/lib/systemd/systemd --version`).

PR #808, opened 2023-05-10
Upgrade Docker in main from 20.10.24 to 23.0.6

flatcar-infra:
Subject says it all.


dongsupark:
Added `on-hold`, as we would need to wait for a while until migration to Docker 23.x could be done.

t-lo:
Closing since we merged a docker24 update recently.

PR #809, opened 2023-05-10
sys-apps/acl: Sync with Gentoo to fix crash

jepio:
# sys-apps/acl: Sync with Gentoo to fix crash

getacl is currently (main/alpha/beta) failing with a buffer overflow:

```
  $ getfacl /
  getfacl: Removing leading '/' from absolute path names
  # file: .
  # owner: root
  # group: root
  *** buffer overflow detected ***: terminated
  Aborted (core dumped)
```

This has been addressed upstream in acl-2.3.1-r1 by lowering FORTIFY_SOURCE level.

I can't reproduce with stable even though stable has the same acl version, so this most likely depends on GCC versions.

Gentoo bug: https://bugs.gentoo.org/847280.
Upstream commit 6273f7c0c37bc0fae7071ec6237654796bdcca55.

## How to use

```
emerge -av sys-apps/acl
getfacl /
```

## Testing done

```
emerge -av sys-apps/acl
getfacl /
```


jepio:
> Could you please add sys-apps/acl to `.github/workflows/portage-stable-packages-list` too?

Yes, done.

krnowak (approved):
Thanks.

PR #810, opened 2023-05-10
sys-kernel/*: Bump to kernel 6.1.27

krnowak:
Bump the kernel to 6.1.27, sync docker package to avoid checking some kernel config stuff that was dropped in 6.1. Drop a bunch of kernel patches that are already upstreamed. Add a new one that makes the kernel build reproducible, which is very important as BTF format seems to be quite fragile to mismatches that may come from just rebuilding vmlinux from the same sources.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1807/cldsv


krnowak:
Pff, github CI did not kick off, will run one in jenkins.

sdk_container/src/third_party/coreos-overlay/app-emulation/docker/docker-20.10.24-r1.ebuild
```diff
@@ -152,6 +153,12 @@ pkg_setup() {
 		"
 	fi
 
+	if kernel_is lt 6 1; then
+		CONFIG_CHECK+="
+			~MEMCG_SWAP
+		"
+	fi
+
```
dongsupark, line 161: Nice catch.

Ideally, however, we should first create a commit for syncing with Gentoo, (like ee672bcdb9cd414480b48389f5fd07a6148a0125) and then create another commit for Flatcar changes (like 742c507270c1d80629c1a4d4297a5cd93c37d765), so that all other Kernel-related changes could be updated as well.

krnowak: Right, I should have followed that. Will do.

krnowak: Synced properly. Needed to pull in a new version of app-emulation/docker-proxy.

krnowak:
Kicked off another CI build. The previous one looks promising, the tests passed.

krnowak:
CI passed, ready for review.

sdk_container/src/third_party/coreos-overlay/app-emulation/docker/docker-20.10.24-r1.ebuild
dongsupark: The Docker part looks good

dongsupark (approved):

pothos:
In the CI result I see that the config now sets `CONFIG_BPF_UNPRIV_DEFAULT_OFF=y`. Before `/proc/sys/kernel/unprivileged_bpf_disabled` was 0 but now it would 1 one, I'm not sure if that is intended, I think it should stay 1, i.e., the config should be `n`?

pothos:
Another config change is `CONFIG_USELIB` which is now on but I think we don't need it. Given that the kernel grew 2 MB and is now over the 60 MB watermark, I think we need to review the config and trim things down again or systems won't be able to update in the future.

PR #817, opened 2023-05-12
Upgrade Linux Kernel for flatcar-3510 from 5.15.108 to 5.15.111

flatcar-infra:
Subject says it all.


dongsupark:
We need to first release the next Stable with Kernel 5.15.108, https://github.com/flatcar/scripts/pull/729, not this one.
Added `on-hold`.

dongsupark:
See instead https://github.com/flatcar/scripts/pull/826

PR #819, opened 2023-05-12
Upgrade Containerd in main from 1.6.20 to 1.7.1

flatcar-infra:
Subject says it all.


dongsupark:
Added `on-hold`, as we would need to wait for a while until migration to containerd 1.7 could be done.

dongsupark:
See instead https://github.com/flatcar/scripts/pull/899

PR #823, opened 2023-05-17
Detect multipart mime and ignore it

gabriel-samfira:
# Detect multipart mime user-data

This change makes ignition ignore user-data that is multipart mime.

## How to use

Deploy a VM and add multipart mime userdata. If the VM boots, it works.

## Testing done

Deployed a VM on OpenStack and set its userdata to a multipart mime userdata. The VM succeeded to boot.


sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0019-Detect-multipart-mime-userdata.patch
```diff
@@ -0,0 +1,291 @@
...
++	return strings.Contains(header, "Content-Type: multipart/mixed;")
++}
++
+ func isCloudConfig(userdata []byte) bool {
+ 	header := strings.SplitN(string(decompressIfGzipped(userdata)), "\n", 2)[0]
+ 
+diff --git a/vendor/github.com/flatcar/ignition/config/v2_4/config.go b/vendor/github.com/flatcar/ignition/config/v2_4/config.go
+index e6120bda..b76deeec 100644
+--- a/vendor/github.com/flatcar/ignition/config/v2_4/config.go
++++ b/vendor/github.com/flatcar/ignition/config/v2_4/config.go
```
pothos, line 279: Do we really need to patch the vendor folder? Maybe we could move these checks to our `0006-config-v3_5-convert-ignition-2.x-to-3.x.patch` and to return early when a cloud config or script is found.

sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0019-Detect-multipart-mime-userdata.patch
```diff
@@ -0,0 +1,291 @@
...
++
+ func isCloudConfig(userdata []byte) bool {
+ 	header := strings.SplitN(string(decompressIfGzipped(userdata)), "\n", 2)[0]
+ 
+@@ -62,7 +71,7 @@ func decompressIfGzipped(data []byte) []byte {
+ func HandleParseErrors(rawConfig []byte, to interface{}) (report.Report, error) {
+ 	r := report.Report{}
+ 
+-	if isCloudConfig(rawConfig) || isScript(rawConfig) {
++	if isCloudConfig(rawConfig) || isScript(rawConfig) || isMultipartMime(rawConfig) {
```
pothos, line 50: Or maybe is it enough to do it only here and the changes in bendor are not needed?

sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0019-Detect-multipart-mime-userdata.patch
gabriel-samfira: no. I was on auto pilot when I made these changes. Searched for places where we added cloud-config and script handling and added multipart mime. I'll refrain from making changes at night :smile:

sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0019-Detect-multipart-mime-userdata.patch
gabriel-samfira: This is indeed the only needed change.

gabriel-samfira: updated. PTAL.

sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0019-detect-multipart-mime-and-ignore-it.patch
```diff
@@ -0,0 +1,42 @@
...
++
+ func isCloudConfig(userdata []byte) bool {
+ 	header := strings.SplitN(string(decompressIfGzipped(userdata)), "\n", 2)[0]
+ 
+@@ -62,7 +71,7 @@ func decompressIfGzipped(data []byte) []byte {
+ func HandleParseErrors(rawConfig []byte, to interface{}) (report.Report, error) {
+ 	r := report.Report{}
+ 
+-	if isCloudConfig(rawConfig) || isScript(rawConfig) {
++	if isCloudConfig(rawConfig) || isScript(rawConfig) || isMultipartMime(rawConfig) {
```
gabriel-samfira, line 36: I wonder if it wouldn't be easier to test for a valid ignition config and just ignore foreign formats, instead of failing and dropping to ```emergency.target```. The assumption being that some other tooling might do something meaningful with that data.

jepio: i don't know how much `isIgnitionConfig` would check, but we wouldn't want ignition to ignore invalid ignition configs and continue booting without error.
generally ignition assumes that a distro that uses ignition relies on it for correctness, and there are only a handful special-cases where it should yield to a different tool (like here). there was also a discussion here: https://github.com/coreos/ignition/issues/849.

gabriel-samfira: Apropos, this PR:

  * https://github.com/flatcar/coreos-cloudinit/pull/21

adds multipart mime support to coreos-cloudinit.

gabriel-samfira: > I think there is _some_ value in ignoring userdata that is not a valid ignition config.

To clarify, I don't have particularly strong feelings about this. I am okay with just ignoring the extra case proposed in this patch (included in patch 13, as @pothos mentioned).

sdk_container/src/third_party/coreos-overlay/sys-apps/ignition/files/0019-Detect-multipart-mime-userdata.patch
gabriel-samfira: I added the changes to patch 13.

pothos:
Last time I checked `0019-detect-multipart-mime-and-ignore-it.patch` didn't apply during the build. Is this solved already?

PR #824, opened 2023-05-18
Upgrade Linux Kernel for flatcar-3602 from 5.15.111 to 5.15.112

flatcar-infra:
Subject says it all.


dongsupark:
See instead https://github.com/flatcar/scripts/pull/841

PR #829, opened 2023-05-18
Upgrade Linux Kernel for flatcar-3033 from 5.10.178 to 5.10.180

flatcar-infra:
Subject says it all.


dongsupark (approved):
CI passed.

Note: In tests like http://jenkins.infra.kinvolk.io:8080/job/container/job/test/10645/, kubeadm tests still fail, but that is expected, because the test ran without rebase on top of the bug fix https://github.com/flatcar/scripts/pull/836.

PR #838, opened 2023-05-24
Upgrade Docker in main from 20.10.24 to 24.0.1

flatcar-infra:
Subject says it all.


dongsupark:
We first need to update Docker 23, before trying out 24.
Disabled the Docker workflow temporarily.

PR #839, opened 2023-05-24
dev-lang/go: update Go to 1.20.4

dongsupark:
1. Add `dev-lang/go` 1.20.4 and drop the unnecessary 1.18.10.
1. Update the default Go version from 1.19 to the latest 1.20.

1. Since Go 1.20 [stopped shipping](https://go.dev/doc/go1.20) pre-compiled package archives for the standard library in `$GOROOT/pkg`, it became impossible to install `pkg/linux_{amd64,arm64}` for every version. To avoid build failure, install the library if the Go slot number is 1.19 or older.

1. Update `dev-lang/go-bootstrap` to 1.18.6, by syncing with Gentoo commit [1b3154a4473d](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=1b3154a4473deedaf3d1deb978f9a19739ac03af). This is necessary because Go >= 1.20 requires go-bootstrap to be at least 1.17.13.
  - Drop unnecessary Flatcar patches from `go-bootstrap`, and move the ebuild from coreos-overlay to portage-stable.

1. Go 1.18 is already EOL, so no security update at all. Since upstream docker projects already have Go [1.19](https://github.com/moby/moby/blob/5d6db842238e3c4f5f9fb9ad70ea46b35227d084/Dockerfile#L6), Flatcar should also have that.
  - Note, we still keep setting `COREOS_GO_VERSION="go1.19"` for `containerd`, `cri-tools`, `docker`, `docker-cli`, `docker-proxy`, `docker-runc`, following the default version of upstream projects.
  -  Delete Go 1.12 which does not work with go-bootstrap 1.18.

1. For `amazon-ssm-agent`, use Go 1.19 as Go 1.12 does not exist. Update `amazon-ssm-agent` to 3.2.985.0, because 2.3.1319.0 does not work with Go 1.19.

Fixes https://github.com/flatcar/Flatcar/issues/1035.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/825/cldsv/


sayanchowdhury (approved):

dongsupark:
Just a moment. I am rerunning the CI on Jenkins http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/822/cldsv/ just in case.

sayanchowdhury (approved):

PR #840, opened 2023-05-25
Upgrade Linux Kernel for flatcar-3510 from 5.15.111 to 5.15.113

flatcar-infra:
Subject says it all.


dongsupark:
We need to first release the next Stable with Kernel 5.15.111, https://github.com/flatcar/scripts/pull/817, not this one.
Added `on-hold`.

PR #844, opened 2023-05-25
Test systemd selinux fix

pothos:
This is a test PR to see if we can still reproduce the issue. If we can't, this backport should be upstreamed and then we integrate it by updating the 252 stable tag (on all of our channels).

## Testing done


pothos:
Worked, will file this to main once it lands upstream

PR #845, opened 2023-05-25
app-emulation: Upgrade Docker 20.10.24 to 20.10.25

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1872/cldsv


krnowak:
`systemd.sysext.custom-docker` failed for 20.10.25, because 20.10.25 binaries are still not available in neither https://download.docker.com/linux/static/stable/x86_64/ nor https://download.docker.com/linux/static/stable/aarch64/

dongsupark:
Ok, let's wait for now. Added `on-hold`.

dongsupark:
As it is outdated, and the missing tarballs will not be available, I would like to suggest closing the PR.

krnowak:
Yeah, updating to 24.x would be a better way forward. Closing then.

PR #846, opened 2023-05-26
workflows/pr-comment,run-kola: fix concurrency + results comment

t-lo:
This change includes 2 unrelated small fixes to the "PR comment command" build+test feature:

1. The concurrency group now contains the PR number, ensuring that builds for multiple PRs can run in parallel. This addresses an issue in which starting a build on a PR would cancel a running build of another PR.
2. Post test results to the PR where the build command was issued on. The add-pr-comment step's condition was not updated when switching to PR comment commands for starting builds, so the step would only run if the action was triggered by a PR change event. Since we now trigger on issue_comment, the step never ran.

Additionally, the PR from which a run was triggered will now receive a comment with a link to the workflow run.

Fixes [#1046](https://github.com/flatcar/Flatcar/issues/1046).

### DIFF
--- a/.github/workflows/pr-comment-build-dispatcher.yaml
+++ b/.github/workflows/pr-comment-build-dispatcher.yaml
@@ -7,7 +7,7 @@ permissions:
   pull-requests: write
 
 concurrency:
-  group: ${{ github.workflow }}-pr-command-${{ github.head_ref || github.ref_name }}
+  group: ${{ github.workflow }}-pr-command-${{ github.event.issue.pull_request.number }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/pr-comment-build-dispatcher.yaml
+++ b/.github/workflows/pr-comment-build-dispatcher.yaml
@@ -57,8 +57,8 @@ jobs:
 
   update_sdk:
     needs: check_maintainer_membership
-    if: ( always() && needs.check_maintainer_membership.result == 'success'
-                   && contains(github.event.comment.body, '/update-sdk') )
+    if: ( needs.check_maintainer_membership.result == 'success'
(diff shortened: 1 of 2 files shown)


pothos (approved):

gabriel-samfira:
We should remove the

```yaml
if: always()
```

condition.

t-lo:
> We should remove the
> 
> ```yaml
> if: always()
> ```
> 
> condition.

I had trouble getting the condition to work properly at some point, but this might have been before I refactored the PR commands action. Might be cargo cult at this point, so I've removed it.

gabriel-samfira (approved):

.github/workflows/pr-comment-build-dispatcher.yaml
```diff
@@ -49,19 +49,25 @@ jobs:
...
     name: "Build an updated SDK container"
     # SDK build needs access to bincache ssh secret
     secrets: inherit
     uses: ./.github/workflows/update-sdk.yaml
 
   build_image:
     needs: [ check_maintainer_membership, update_sdk ]
-    if: ( always() && needs.check_maintainer_membership.result == 'success'
-                   && (    contains(github.event.comment.body, '/build-image') || needs.update_sdk.result == 'success' ) )
+    if: ( needs.check_maintainer_membership.result == 'success'
```
gabriel-samfira, line 69: Apparently doing something like:

```yaml
if: (always() && !canceled()) && other_conditions
```

takes care of the "uncancel-able" job issue we saw when using ```always()```.

t-lo: OOOF 🙃
Let's give it a try without `always()` and see if it works. I'll keep the above in mind in case it doesn't.

PR #856, opened 2023-05-31
Upgrade Linux Kernel for main from 6.1.30 to 6.1.31

flatcar-infra:
Subject says it all.


dongsupark:
See instead https://github.com/flatcar/scripts/pull/886

PR #683, opened 2023-03-10
export portage db, oem sysext utils, sysext for azure

krnowak:
Portage database is put into the tarball and uploaded to bincache before it's dropped from the generic image. This is done in order to reinject it into the mounted generic image when building sysext images, so we have retained all the portage information and we don't need to strip RDEPENDS from the oem packages. That way they can become bog-standard ebuilds.

Adds some scripts for building oem sysexts and then ports azure to use them.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1472/cldsv

Related coreos overlay PR: https://github.com/flatcar/coreos-overlay/pull/2506

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

PR #858, opened 2023-05-31
export portage db, oem sysext utils, sysext for azure

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1905/cldsv

Portage database is put into the tarball and uploaded to bincache before it's dropped from the generic image. This is done in order to reinject it into the mounted generic image when building sysext images, so we have retained all the portage information and we don't need to strip RDEPENDS from the oem packages. That way they can become bog-standard ebuilds.

Adds some scripts for building oem sysexts and then ports azure to use them.

overlay app-emulation/wa-linux-agent: Start changing the ebuild to resemble a normal ebuild, without the OEM hacks. This replaces dependency on dev-lang/python-oem and dev-python/distro-oem with dev-lang/python and dev-python/distro, respectively.

overlay dev-lang/python-oem, dev-python/distro-oem: Drop the unnecessary packages.

overlay profiles, coreos-devel/board-packages: Drop things related to the dropped python packages. Maybe we should start pulling in dev-lang/python instead?

This PR replaces https://github.com/flatcar/scripts/pull/683 and https://github.com/flatcar/coreos-overlay/pull/2506.


build_library/oem_sysext_util.sh
```diff
@@ -0,0 +1,201 @@
...
+        info "  Removing ${entry##*/}"
+        rm -rf "${entry}"
+    done
+
+    local metadata metadata_file
+    info "Adding sysext metadata"
+    mkdir -p "${sysext_rootfs}/usr/lib/extension-release.d"
+    metadata=(
+        'ID=flatcar'
+        "VERSION_ID=${version_id}"
```
pothos, line 165:
```suggestion
        "SYSEXT_LEVEL=1.0"
```
Same here, this will be used until update-engine can update extensions.

build_library/prod_image_util.sh
```diff
@@ -100,6 +101,8 @@ create_prod_image() {
           || die_notrace "coreos-au-key is missing the 'official' use flag"
   fi
 
+  tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
+
```
t-lo, line 105: Eventually, we should also capture `"${BUILD_DIR}"/configroot` and pass it as `PORTAGE_CONFIGROOT` later when generating sysexts, but I guess the above works for now.

I've created a follow-up issue https://github.com/flatcar/Flatcar/issues/1052 to keep track of this.

krnowak: I think configroot is considered as deprecated and should always be the same as sysroot. Not sure if it's relevant, though, See https://dev.gentoo.org/~zmedico/portage/doc/man/emerge.1.html (search for PORTAGE_CONFIGROOT there).

t-lo: That's a good point (and also, TIL! thanks for sharing) but it's not quite what I meant. *Currently*, `build_image` uses a config root external to the OS image rootfs, via `prod_image_util.sh::create_prod_image()` -> `build_image_util.sh::emerge_to_image()`, here: https://github.com/flatcar/scripts/blob/main/build_library/build_image_util.sh#L183

All I'm saying is that at some point (not necessarily in your PR though) we should capture the contents from there, too, to accurately recreate the base OS image including all portage metadata when we install sysext packages on top.

But doing this cleanly is for https://github.com/flatcar/Flatcar/issues/1052 to look into. No reason to feature-creep this into your work here, Krzesimir.

krnowak:
Needs https://github.com/flatcar/bootengine/pull/62 to be merged first.

pothos (approved):
Thanks a lot!

changelog/changes/2023-06-06-sysext-for-azure-and-qemu-oem.md
```diff
@@ -0,0 +1 @@
+- Azure and QEMU images currently use sysext images for additional platform-specific software. For Azure images this also means that the image will have a normal python installation available through the sysext image.
```
pothos, line 1:
```suggestion
- Azure and QEMU OEM images now use systemd-sysext images for layering additional platform-specific software on top of `/usr`. For Azure images this also means that the image has a normal Python installation available through the sysext image. The OEM software is still not updated but this will be added soon.
```

krnowak: Thanks!

PR #862, opened 2023-06-01
Add on pr workflow

gabriel-samfira:
This replaces #860 

# Add a workflow that runs on every PR

This change adds a workflow that is triggered on every PR. This workflow is started in the ```development``` environment and required manual approval from the flatcar-maintainers team before it runs.

Once approved, the workflow runs as usual.

## Triggering an SDK update

When opening a new PR, add the following string anywhere in the **PR summary**:

/update-sdk

This will be parsed by the workflow and signal the ```update_sdk``` job to run. 

Note, the workflows will probably be active for this PR and will try to update the SDK and then build the image. If we don't want to do that, we can simply not approve the run.

The command based PR runs have also been updated and *should* work the same way. If we want to trigger just the image build, it should be enough to specify:

/build-image

anywhere in the comment body. We can optionally use:

/update-sdk
/build-image

to also update the SDK. 

Note: any workflow that requires secrets will not run if the PR is opened from a fork.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,6 @@
 name: "Run build"
 on:
+  pull_request:
   workflow_dispatch:
     inputs:
       image_formats:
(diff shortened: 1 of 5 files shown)


t-lo (approved):
LGTM, thanks!

PR #863, opened 2023-06-01
Move sys-process/tini from overlay to portage-stable

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1911/cldsv

This drops one customization that we had in the docker ebuild.


krnowak:
CI passed (other than the coreos.update.badusr flake).

PR #865, opened 2023-06-01
Update commonconfig-6.1 to build kernel with TLS module

krishjainx:
This pull request aims to modify commonconfig-6.1 in order to incorporate a TLS module for kernel building. By integrating the TLS module, applications and containers will have the ability to offload TLS processing to the kernel.

Fixes https://github.com/flatcar/Flatcar/issues/995

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1992/cldsv/

PR #866, opened 2023-06-01
Migrate to Type=notify

krishjainx:
#  [containerd] Migrate to Type=notify in containerd.service
    
Race condition arises when the containerd service unit assumes services are ready as soon as they start running, rather than when they can actually accept socket requests. To rectify this, changing the unit to Type=notify is required, utilizing the existing containerd support for sd_notify call after socket setup.
    
 References:
- https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=
- https://github.com/flatcar/Flatcar/issues/203

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1993/cldsv/

### DIFF
--- a/changelog/changes/2023-06-15-containerd-service
+++ b/changelog/changes/2023-06-15-containerd-service
@@ -1,0 +1,1 @@
+- Migrate to Type=notify in containerd.service. Changed the unit to Type=notify, utilizing the existing containerd support for sd_notify call after socket setup.
--- a/sdk_container/src/third_party/coreos-overlay/app-containers/containerd/files/containerd.service
+++ b/sdk_container/src/third_party/coreos-overlay/app-containers/containerd/files/containerd.service
@@ -1,20 +1,24 @@
 [Unit]
-Description=Containerd Container Daemon
-Documentation=http://github.com/docker/containerd
+Description=containerd container runtime
+Documentation=https://containerd.io
+After=network.target local-fs.target
 
 [Service]
-Type=simple
+ExecStartPre=-/sbin/modprobe overlay
 ExecStart=/usr/bin/containerd --listen unix:///var/run/docker/libcontainerd/docker-containerd.sock --shim /usr/bin/containerd-shim --state-dir /var/run/docker/libcontainerd/containerd --start-timeout 2m
-Restart=always
 
-# (lack of) limits from the upstream docker service unit
-LimitNOFILE=1048576
+Type=notify
+Delegate=yes
+KillMode=process
+Restart=always
+RestartSec=5
+# Having non-zero Limit*s causes performance problems due to accounting overhead
+# in the kernel. We recommend using cgroups to do container-local accounting.
 LimitNPROC=infinity
 LimitCORE=infinity
+LimitNOFILE=infinity
 TasksMax=infinity
-
-# set delegate yes so that systemd does not reset the cgroups of containers
-Delegate=yes
+OOMScoreAdjust=-999
 
 [Install]
 WantedBy=multi-user.target early-docker.target


jepio:
Can you check `git log sdk_container/src/third_party/coreos-overlay/app-emulation/containerd` and then update the commit message to specify the package name in the first line and have a bit more information in the commit body about what is being done and why. Thanks

krishjainx:
@jepio Does this work?

jepio:
It's better, but if you look at the git log you'll see that most historical commit messages have a summary line starting with:
`app-emulation/containerd` and we wrap the text in the body at <80 columns.

krishjainx:
This resolves it @jepio

sdk_container/src/third_party/coreos-overlay/app-emulation/containerd/files/containerd.service
```diff
@@ -3,7 +3,7 @@ Description=Containerd Container Daemon
 Documentation=http://github.com/docker/containerd
 
 [Service]
-Type=simple
+Type=notify
```
pothos, line 6: I think we should take this opportunity to align it more with upstream:
https://github.com/containerd/containerd/blob/main/containerd.service
```suggestion
Type=notify
KillMode=process
RestartSec=5
OOMScoreAdjust=-999
LimitNOFILE=infinity
ExecStartPre=-/sbin/modprobe overlay
```
(different order and without the LimitNOFILE duplicate below)

sdk_container/src/third_party/coreos-overlay/app-emulation/containerd/files/containerd.service
```diff
@@ -3,7 +3,7 @@ Description=Containerd Container Daemon
 Documentation=http://github.com/docker/containerd
```
pothos, line 3:
```suggestion
Documentation=https://containerd.io
After=network.target local-fs.target
```
and `Description=containerd container runtime`

dongsupark:
> @dongsupark Unsure why Kola tests fail

From a quick look, CI failed with 17 different failures. Those consist of 3 different types:

* coreos.update.badusr
  - It is expected, an already known issue, please ignore it. That has been always a source of flaky tests, so the kola tests are already trying to retry 5 times at max to get that done. That should be enough mostly in case of (our internal) Jenkins CI runs. With the Github Actions CI, however, somehow it is not enough.

* kubeadm.v1.24.14.calico.cgroupv1.base, etc
  - Looks like flaky tests, That seems to sometimes happen in other PRs as well. Have not looked into its detail.

* devcontainer.systemd-nspawn, devcontainer.docker
  - `error: pathspec 'c326eab136' did not match any file(s) known to git`
  - It is interesting, because that failure seems to happen only when you ran the tests. (Same as https://github.com/flatcar/scripts/pull/865) I am wondering if it has something to do with the dev container trying to clone the default flatcar/scripts git repo, and trying to find a wrong commit c326eab136, which exists only in your cloned repo.

sdk_container/src/third_party/coreos-overlay/app-emulation/containerd/files/containerd.service
pothos: Yes, maybe follow the upstream order of settings for an easier to understand `diff` output

pothos: Thanks, looks good

pothos:
This needs a rebase. The tests looked good, the failed devcontainer test just has problems because this PR is from another repo.

jepio:
You still need to rebase c462c411eb49740f9fb89b578e53bfcb82a80c03 onto main

jepio:
The commit message still says "app-emulation", that should be "app-containers" too

dongsupark:
Running Jenkins CI http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1993/cldsv/ for cross-checking.

dongsupark:
Jenkins CI passed.

A changelog file is missing here as well, like the other [one](https://github.com/flatcar/scripts/pull/865#issuecomment-1593261454).

dongsupark (approved):
Thanks!

PR #867, opened 2023-06-01
Add fstrim conf

krishjainx:
# Add fstrim conf

This pull request aims to include an fstrim configuration to util-linux and install it in pkg_postinst. By doing so, the default configuration shipped with upstream will be overwritten. This change is necessary to prevent the failure of fstrim.service on Flatcar Linux due to the absence of /etc/fstab.

### DIFF
--- a/sdk_container/src/third_party/portage-stable/sys-apps/util-linux/files/fstrim.conf
+++ b/sdk_container/src/third_party/portage-stable/sys-apps/util-linux/files/fstrim.conf
@@ -1,0 +1,5 @@
+[Service]
+# Use `-a` flag - Flatcar does not have /etc/fstab file - so run against all
+# mounted filesystems
+ExecStart=
+ExecStart=/sbin/fstrim -av
--- a/sdk_container/src/third_party/portage-stable/sys-apps/util-linux/util-linux-2.37.4.ebuild
+++ b/sdk_container/src/third_party/portage-stable/sys-apps/util-linux/util-linux-2.37.4.ebuild
@@ -330,4 +330,5 @@ pkg_postinst() {
 		elog "The agetty util now clears the terminal by default. You"
 		elog "might want to add --noclear to your /etc/inittab lines."
 	fi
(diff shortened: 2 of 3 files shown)


jepio:
> By doing so, the default configuration shipped with upstream will be overwritten.

I don't see an upstream configuration that is being overwritten.

> This change is necessary to prevent the failure of fstrim.service on Flatcar Linux due to the absence of /etc/fstab.

It doesn't look like it's failing to me:
```
core@localhost ~ $ sudo systemctl status fstrim.service
○ fstrim.service - Discard unused blocks on filesystems from /etc/fstab
     Loaded: loaded (/usr/lib/systemd/system/fstrim.service; static)
     Active: inactive (dead)
       Docs: man:fstrim(8)

Jun 02 06:58:08 localhost systemd[1]: Starting fstrim.service...
Jun 02 06:58:08 localhost fstrim[1428]: /boot: 69.9 MiB (73270272 bytes) trimmed on /dev/vda1
Jun 02 06:58:08 localhost fstrim[1428]: /usr/share/oem: 127 MiB (133132288 bytes) trimmed on /dev/vda6
Jun 02 06:58:08 localhost fstrim[1428]: /: 5.9 GiB (6299480064 bytes) trimmed on /dev/vda9
Jun 02 06:58:08 localhost systemd[1]: fstrim.service: Deactivated successfully.
Jun 02 06:58:08 localhost systemd[1]: Finished fstrim.service.
core@localhost ~ $ ls -la /etc/fstab
ls: cannot access '/etc/fstab': No such file or directory
```

jepio:
Can you follow the PR template that appears when you open a PR:
- what is the error currently (show output)
- how do you validate that it works after this PR
- the commit message should have more detail (at least a couple of sentences), and start with the package name being touched.

krishjainx:
My bad @jepio . It appears that this was fixed in this commit https://github.com/flatcar/scripts/commit/faf115acdec955cc31d528ada266d9053c13bc71 and in flatcar release stable-3510.2.2.

PR #868, opened 2023-06-02
Update from "flatcar_production_packet.ipxe" to "flatcar_production_equinix_metal.ipxe"

krishjainx:
# Update from "flatcar_production_packet.ipxe" to "flatcar_production_equinix_metal.ipxe"

In the current release, the "flatcar_production_packet.ipxe" file still mentions "Packet" instead of "Equinix Metal". To address this, let's rename the file from "flatcar_production_packet.ipxe" to "flatcar_production_equinix_metal.ipxe" for accurate representation.

Reference: https://github.com/flatcar/Flatcar/issues/483

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -780,7 +780,7 @@ You can pass extra kernel parameters with -append, for example:
 When using -nographic or -serial you must also enable the serial console:
   ./$(basename "${script}") -nographic -append 'console=ttyS0,115200n8'
 EOF
-    local packetipxe="$(_dst_dir)/flatcar_production_packet.ipxe"
+    local packetipxe="$(_dst_dir)/flatcar_production_equinix_metal.ipxe"
     cat > "$packetipxe" <<EOF
 #!ipxe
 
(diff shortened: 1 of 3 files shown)


build_library/vm_image_util.sh
```diff
@@ -780,7 +780,7 @@ You can pass extra kernel parameters with -append, for example:
 When using -nographic or -serial you must also enable the serial console:
   ./$(basename "${script}") -nographic -append 'console=ttyS0,115200n8'
 EOF
-    local packetipxe="$(_dst_dir)/flatcar_production_packet.ipxe"
+    local packetipxe="$(_dst_dir)/flatcar_production_equinix_metal.ipxe"
```
jepio, line 783: How does this impact users who rely on the current filename?

jepio (commented):
The fstrim changes do not belong in this PR

build_library/vm_image_util.sh
krishjainx: > The fstrim changes do not belong in this PR

I fixed that @jepio 

> How does this impact users who rely on the current filename?

I don't think it does.

dongsupark: +1 on what @jepio said. We cannot simply rename like that.
It is one of the cases where we should have a long-term deprecation plan.
For example, support both names for a while, announce the deprecation in public, and finally deprecate packet.

dongsupark: Let's put it this way. That rename has never been a priority. ;-)
Sure, would be nice to do rename that.

pothos: Assuming that rsync preserves the symlinks and that caddy/nginx handle them well, I think we can do the rename and generate the compat symlinks here as well.

PR #892, opened 2023-06-07
sys-apps/systemd: Update to 252.11

pothos:
This pulls in
https://github.com/systemd/systemd-stable/releases/tag/v252.11 mainly for fixing https://github.com/flatcar/Flatcar/issues/1048

Added another commit:
- sys-kernel/bootengine: Prevent a bootloop on setup-root service failure
    
    This pulls in
    https://github.com/flatcar/bootengine/pull/63
    to prevent the bootloop that sometimes happened and now happens more
    often, e.g., as seen with the coreos.update.badusr kola test.

## How to use

This should eventually land in Stable.
There are of course other backports which could have regressions, I suggest we first update it in Alpha and Beta and after two weeks in-the-wild time we also do it for Stable, like we do for kernel point releases.

## Testing done

Ongoing, workflow plus [Jenkins for clouds](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1941/cldsv/)


pothos:
The badusr test passed when executed manually, I wonder if this is a timeout issue

krnowak (approved):

pothos:
Backported for Beta/Alpha. If we do it later for Stable, first cherry-pick the bootengine repo commit to `flatcar-3510-backport`, then use `flatcar-build-scripts/cherry-pick-for flatcar-3510 https://github.com/flatcar/scripts/pull/892` and on the conflict edit the bootengine ebuild to use the commit ID from `flatcar-3510-backport`, run `git add` and `git am --continue`, then copy-paste the printed instructions (`git checkout … git merge --no-ff … git branch -d xyz… push`).

pothos:
End of the June we can backport to Stable

pothos:
Did the backport just now which is after the Stable release build, thus it will land in the next one.

PR #907, opened 2023-06-13
overlay: Profile cleanups

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/851/cldsv

- Clean the accept keywords a bit.
  - There were duplicates, unnecessary or obsolete entries.
  - Also, added comments to the keywords where possible.
    - I have no idea why we have accept keywords for sys-power/acpid, thus the `# ?` comment there.
- Merge all accept keywords files into one in the coreos/base profile.
  - This will be easier for the future package automation to update.
  - Placement of some accept keywords was inconsistent (keywords for SDK were in base profile, SDK target profile and in arm64 profile)
    - I'd say that the base profile should be the one deciding which package version should be installed for any other profile, while the other profiles may decide if they will pull the package into the rootfs, or not.
 - Put one security changelog in proper place.
 - Drop unmasking of openssl 3 - looks like profiles in portage-stable are not masking it any more.


krnowak:
CI passed.

dongsupark (approved):
Thanks for moving `2023-06-02-sudo-1.9.13p3.md` into a proper place. Yesterday I found out myself when touching other things. ;-)

Sorting things in packages.accept_keywords, I would have to rebase my PR https://github.com/flatcar/scripts/pull/908, but that's ok. Let's get this PR merge first.

PR #908, opened 2023-06-13
dev-lang/rust: Update to 1.70.0

dongsupark:
Update `dev-lang/rust` to 1.70.0, syncing with Gentoo commit [7aba087b0bd4](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=7aba087b0bd466d34adb02dba157f5a7ba5bbbe9).

Apply Flatcar modifications like below based on commit [9ac87490bd4a](https://github.com/flatcar/scripts/commit/9ac87490bd4a4306496b4dfaa0aa8e85fb10e8d7):
 - Apply cross-compiling changes.
 - Remove python3_12, which is not supported by Flatcar.
 - Remove dependency on `sys-apps/lsb-release`, which conflicts with `sys-apps/baselayout` of Flatcar.
 - Remove rust-src to keep the SDK size minimal.
 - Add a comment about keeping the patchlevel 0 of stage0.
 - Update Rust version from 1.69 to 1.70 in profiles.

Fix build issues in afterburn and update-ssh-keys by setting LTO to false in `Cargo.toml`.
Pulls in https://github.com/flatcar/update-ssh-keys/pull/8.

Note: we have to create a manual PR like this, because recently Github Actions for getting the latest Rust version stopped working, due to a wrong stage0 patchlevel bumped by upstream Gentoo.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/861/cldsv/


dongsupark:
Please ignore the Github Actions test result. It does not support the full SDK build yet.
See instead http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/852/cldsv/.

dongsupark:
CI failed when building afterburn and update-ssh-keys.

dongsupark:
I think I know what is going on.

Gentoo ebuild of Rust 1.70 started turning on [`USE=lto`](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=06c05b95a6ccc8456ba1a829d89a3834e781e96b) by default.
That makes both builds of afterburn and update-ssh-keys fail.
Changing `lto = true` to `lto = "off"` (or similar) in `Cargo.toml` makes the build failure disappear.

To avoid creating another 3rd-party patch in afterburn, I will disable `USE=lto` in Rust for now.

dongsupark:
CI passed http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/861/cldsv/.

First we need to get https://github.com/flatcar/update-ssh-keys/pull/8 merged.

sdk_container/src/third_party/coreos-overlay/dev-lang/rust/rust-1.70.0.ebuild
```diff
@@ -3,6 +3,7 @@
 
 EAPI=8
 
+# Flatcar: skip python3_12 which is not supported
```
krnowak, line 6: Was this change necessary? Python 12 is masked, so we will use python 3.10 anyway. I mean - the change itself is not a problem, just checking if we can minimize the diff with Gentoo.

krnowak (approved):

sdk_container/src/third_party/coreos-overlay/dev-lang/rust/rust-1.70.0.ebuild
dongsupark: It was necessary.
I have just tested again with adding back python3_12. Then it immediately failed.

```
Calculating dependencies / * ERROR: dev-lang/rust-1.70.0::coreos failed (depend phase):
 *   Invalid implementation in PYTHON_COMPAT: python3_12
```

krnowak: Interesting, probably a recent update in python eclass is needed. Anyway, this will come with weekly updates, so let's keep this modification for now.

PR #910, opened 2023-06-13
build_library: For now don't use a version for the OEM sysext image

pothos:
The initial MVP of the OEM sysext usage we release won't have updates for the sysext image and, therefore, it is not bound to the OS version. The special name suffix instead of the version hints bootengine at using it if no matching version is found. The name will also be used at hint for update-engine to clean it up when versioned sysext images arrive.

## Testing done

Tested with [Jenkins](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/1968/cldsv/) and checked the filename on the qemu OEM image.

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -548,19 +548,22 @@ install_oem_sysext() {
     local built_sysext_dir="${FLAGS_to}/${oem_sysext}-sysext"
     local built_sysext_filename="${oem_sysext}.raw"
     local built_sysext_path="${built_sysext_dir}/${built_sysext_filename}"
+    # TODO: Set 'version' to with "${FLATCAR_VERSION}" and
+    # 'version_id' to "${FLATCAR_VERSION_ID}" when we implement updating OEM sysexts
+    local version='initial'
+    local version_id='initial'
     local build_oem_sysext_flags=(
         --board="${BOARD}"
         --build_dir="${built_sysext_dir}"
         --prod_image_path="${VM_SRC_IMG}"
         --prod_pkgdb_path="${VM_SRC_PKGDB}"
-        # TODO: Drop this when we implement updating OEM sysexts.
-        --version_id=initial
+        --version_id="${version_id}"
     )
 
     "${SCRIPT_ROOT}/build_oem_sysext" "${build_oem_sysext_flags[@]}" "${oem_sysext}"
 
     local installed_sysext_oem_dir='/oem/sysext'
-    local installed_sysext_file_prefix="${oem_sysext}-${FLATCAR_VERSION}"
+    local installed_sysext_file_prefix="${oem_sysext}-${version}"
     local installed_sysext_filename="${installed_sysext_file_prefix}.raw"
     local installed_sysext_abspath="${installed_sysext_oem_dir}/${installed_sysext_filename}"
     info "Installing ${oem_sysext} sysext"


krnowak:
Oh, meh, one more place that I missed. How about this diff to keep the version in one place? Though I'm not sure if it's correct - usage of version vs version ID begins to confuse me.

```diff
diff --git a/build_library/vm_image_util.sh b/build_library/vm_image_util.sh
index 802f0aa137..4c2d046697 100644
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -548,19 +548,21 @@ install_oem_sysext() {
     local built_sysext_dir="${FLAGS_to}/${oem_sysext}-sysext"
     local built_sysext_filename="${oem_sysext}.raw"
     local built_sysext_path="${built_sysext_dir}/${built_sysext_filename}"
+    # TODO: Replace 'initial' with "${FLATCAR_VERSION}" when we
+    # implement updating OEM sysexts.
+    local version_id='initial'
     local build_oem_sysext_flags=(
         --board="${BOARD}"
         --build_dir="${built_sysext_dir}"
         --prod_image_path="${VM_SRC_IMG}"
         --prod_pkgdb_path="${VM_SRC_PKGDB}"
-        # TODO: Drop this when we implement updating OEM sysexts.
-        --version_id=initial
+        --version_id="${version_id}"
     )
 
     "${SCRIPT_ROOT}/build_oem_sysext" "${build_oem_sysext_flags[@]}" "${oem_sysext}"
 
     local installed_sysext_oem_dir='/oem/sysext'
-    local installed_sysext_file_prefix="${oem_sysext}-${FLATCAR_VERSION}"
+    local installed_sysext_file_prefix="${oem_sysext}-${version_id}"
     local installed_sysext_filename="${installed_sysext_file_prefix}.raw"
     local installed_sysext_abspath="${installed_sysext_oem_dir}/${installed_sysext_filename}"
     info "Installing ${oem_sysext} sysext"
```

pothos:
Ok, done and restarted the tests

krnowak:
Alright, I have read a bit more. I find the mixing of version and version ID annoying as they are easy to confuse, but well.

We need VERSION_ID (so `1234.0.0`) to write it to the extension file inside the squashfs image. We also need VERSION (so `1234.0.0+stuff`) for naming the image. So this unfortunately makes my diff "neat, plausible, and wrong", for which I'm sorry.

I suppose we could use your old diff or the following one:
```diff
diff --git a/build_library/vm_image_util.sh b/build_library/vm_image_util.sh
index 802f0aa137..6bd754c201 100644
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -548,19 +548,22 @@ install_oem_sysext() {
     local built_sysext_dir="${FLAGS_to}/${oem_sysext}-sysext"
     local built_sysext_filename="${oem_sysext}.raw"
     local built_sysext_path="${built_sysext_dir}/${built_sysext_filename}"
+    # TODO: Replace 'initial' with "${FLATCAR_VERSION}" and
+    # "${FLATCAR_VERSION_ID}" when we implement updating OEM sysexts.
+    local version='initial'
+    local version_id='initial'
     local build_oem_sysext_flags=(
         --board="${BOARD}"
         --build_dir="${built_sysext_dir}"
         --prod_image_path="${VM_SRC_IMG}"
         --prod_pkgdb_path="${VM_SRC_PKGDB}"
-        # TODO: Drop this when we implement updating OEM sysexts.
-        --version_id=initial
+        --version_id="${version_id}"
     )
 
     "${SCRIPT_ROOT}/build_oem_sysext" "${build_oem_sysext_flags[@]}" "${oem_sysext}"
 
     local installed_sysext_oem_dir='/oem/sysext'
-    local installed_sysext_file_prefix="${oem_sysext}-${FLATCAR_VERSION}"
+    local installed_sysext_file_prefix="${oem_sysext}-${version}"
     local installed_sysext_filename="${installed_sysext_file_prefix}.raw"
     local installed_sysext_abspath="${installed_sysext_oem_dir}/${installed_sysext_filename}"
     info "Installing ${oem_sysext} sysext"
```

Another thing that comes to my mind - why not use VERSION_ID for the sysext image name too?

pothos:
Thanks, updated.

The version ID is ambiguous, we can have a dev build with the same ID as a release or as some other dev build.

pothos:
The A/B logic would overwrite one of them and could end up removing it - the update test of a nightly release maintenance branch has the same version ID, and this could happen, too, when one builds a dev build of a release and gives it to a user to reproduce a bug.

krnowak (approved):

PR #917, opened 2023-06-14
selinux: update

tormath1:
This PR is a follow-up of https://github.com/flatcar/portage-stable/pull/339 and https://github.com/flatcar/coreos-overlay/pull/1993 - the idea is to "only" upgrade SELinux related software and to pull the new "selinux-container" policy, relabeling investigation could be done in a second time: at least, let's try to get the foundation.

The main change is a transition from `virt` to `container` policy (which solves some issues) while I did not want to fix _all_ the SELinux AVC in this PR, I just patched the container policy to get Cilium working out of the box (because now, even in permissive mode, it was failing) 

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2521/cldsv/

While it closes a certain number of issues, there is still some investigation to do on the others but from a first look it's just a matter of upstreaming some policies (example: https://github.com/SELinuxProject/refpolicy/pull/621) 

Closes: https://github.com/flatcar/Flatcar/issues/479, https://github.com/flatcar/Flatcar/issues/891, https://github.com/flatcar/Flatcar/issues/696

Tested in the CI with: https://github.com/flatcar/mantle/pull/344

NOTE: All the `kernel_t` related patches have to be DROPPED once we support a fully labelled system.

### DIFF
--- a/.github/workflows/portage-stable-packages-list
+++ b/.github/workflows/portage-stable-packages-list
@@ -295,6 +295,7 @@ eclass/python-single-r1.eclass
 eclass/python-utils-r1.eclass
 eclass/readme.gentoo-r1.eclass
 eclass/savedconfig.eclass
+eclass/selinux-policy-2.eclass
 eclass/strip-linguas.eclass
 eclass/systemd.eclass
 eclass/tmpfiles.eclass
@@ -363,8 +364,17 @@ profiles
 #
 # scripts
 
+sec-policy/selinux-base
+sec-policy/selinux-base-policy
+sec-policy/selinux-container
+sec-policy/selinux-dbus
+sec-policy/selinux-sssd
+sec-policy/selinux-unconfined
+
+
 sys-apps/acl
 sys-apps/attr
+sys-apps/checkpolicy
 sys-apps/config-site
 sys-apps/coreutils
 sys-apps/debianutils
@@ -400,6 +410,7 @@ sys-apps/portage
 sys-apps/pv
 sys-apps/sandbox
 sys-apps/sed
+sys-apps/semodule-utils
 sys-apps/smartmontools
 sys-apps/texinfo
 sys-apps/usbutils
@@ -454,6 +465,8 @@ sys-libs/libcap
 sys-libs/libcap-ng
 sys-libs/libnvme
 sys-libs/libseccomp
+sys-libs/libselinux
+sys-libs/libsepol
 sys-libs/ncurses
 sys-libs/readline
 sys-libs/talloc


sdk_container/src/third_party/coreos-overlay/sys-libs/libsemanage/libsemanage-3.5.ebuild
krnowak, line 1: Same here as in policycoreutils - I wonder if Gentoo would accept `+python` USE flag.

sdk_container/src/third_party/coreos-overlay/sys-libs/libselinux/libselinux-3.5.ebuild
```diff
@@ -145,12 +145,12 @@ pkg_postinst() {
...
-		mkdir -p /etc/selinux/${POLTYPE}/contexts/files || die
-		touch /etc/selinux/${POLTYPE}/contexts/files/file_contexts.local || die
+		mkdir -p "${ROOT}/etc/selinux/${POLTYPE}/contexts/files" || die
+		touch "${ROOT}/etc/selinux/${POLTYPE}/contexts/files/file_contexts.local" || die
 		# Fix bug 516608
 		for EXPRFILE in file_contexts file_contexts.homedirs file_contexts.local ; do
-			if [[ -f "/etc/selinux/${POLTYPE}/contexts/files/${EXPRFILE}" ]]; then
-				sefcontext_compile /etc/selinux/${POLTYPE}/contexts/files/${EXPRFILE} \
+			if [[ -f "${ROOT}/etc/selinux/${POLTYPE}/contexts/files/${EXPRFILE}" ]]; then
+				sefcontext_compile "${ROOT}/etc/selinux/${POLTYPE}/contexts/files/${EXPRFILE}" \
```
krnowak, line 153: Do you have time to upstream those? Or want me to do it?

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
```diff
@@ -113,12 +123,22 @@ src_compile() {
...
 }
 
+# ugly but somehow necessary: these files does not seem detected
+# by portage -> they are orphans. When portage remove the previous package
+# these files remain and prevent the symlink to be done after -> we keep the old stuff
+# and the symlinks are not correctly done (they are done on backup folders preventing collision)
+pkg_preinst() {
+	rm --recursive ${ROOT}/etc/selinux/{mcs,mls,targeted}
+}
+
```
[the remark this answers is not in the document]
krnowak: Or maybe this is something for upstream too?

sdk_container/src/third_party/coreos-overlay/sys-libs/libselinux/libselinux-3.5.ebuild
```diff
@@ -1,9 +1,9 @@
-# Copyright 1999-2021 Gentoo Authors
+# Copyright 1999-2023 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI="7"
-PYTHON_COMPAT=( python3_{6..10} )
-USE_RUBY="ruby25 ruby26 ruby27"
+PYTHON_COMPAT=( python3_{10..11} )
```
krnowak, line 5: I think we have already updated python eclasses to include python 3.12 as a known implementation, so there should be no need for dropping python 3.12 from PYTHON_COMPAT (one Flatcar modification less).

sdk_container/src/third_party/coreos-overlay/sys-libs/libsemanage/libsemanage-3.5.ebuild
```diff
@@ -79,38 +78,18 @@ multilib_src_compile() {
...
-}
+		SHLIBDIR="/usr/$(get_libdir)" \
+		DEFAULT_SEMANAGE_CONF_LOCATION="/usr/lib/selinux/semanage.conf" \
+		DESTDIR="${ED}" \
+		install
 
-multiib_src_install_all() {
-	python_setup
-	python_fix_shebang "${ED}"/usr/libexec/selinux/semanage_migrate_store
+	dotmpfiles "${FILESDIR}/tmpfiles.d/libsemanage.conf"
```
krnowak, line 92: I think we don't need this any more - I think we can just install semanage.conf in `/etc`. But maybe we will need to install a compatibility symlink (`/usr/lib/selinux/semanage.conf` -> `/usr/share/flatcar/etc/selinux/semanage.conf`) with coreos-base/misc-files…

sdk_container/src/third_party/coreos-overlay/sys-libs/libsemanage/libsemanage-3.5.ebuild
```diff
@@ -79,38 +78,18 @@ multilib_src_compile() {
...
-			emake \
-				AR="$(tc-getAR)" \
-				CC="$(tc-getCC)" \
-				LIBDIR="${EPREFIX}/usr/$(get_libdir)" \
-				"$@"
-		}
-		python_foreach_impl building_py swigify
-		python_foreach_impl building_py pywrap
-	fi
+	dotmpfiles "${FILESDIR}/tmpfiles.d/libsemanage.conf"
```
krnowak, line 81: We already install this in `multilib_src_install`. Is this necessary do also do it in `multlib_src_compile`?

sdk_container/src/third_party/coreos-overlay/sys-apps/checkpolicy/checkpolicy-3.5.ebuild
krnowak, line 1: The modifications we do in `src_compile` and `src_install` look wrong to me - they are using ROOT variable, which is only allowed in `pkg_*` functions. Are these modifications still needed?

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
krnowak, line 1: This is about our modifications in `${FILESDIR}/config`:

Instead of modifying the Gentoo-provided config, we probably could provide our own in coreos-base/misc-files, and and add the Gentoo-provided config to PKG_INSTALL_MASK and INSTALL_MASK inside `<coreos-overlay>/coreos/config/env/sec-policy/selinux-base`.

Writing it here, because Github "helpfully" hides the config modifications that we make…

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
```diff
@@ -29,34 +28,27 @@ HOMEPAGE="https://wiki.gentoo.org/wiki/Project:SELinux"
 LICENSE="GPL-2"
 SLOT="0"
 
-# flatcar changes
 RDEPEND=">=sys-apps/policycoreutils-2.8
-	>=sys-apps/checkpolicy-2.8
-"
+	>=sys-apps/checkpolicy-2.8"
 DEPEND="${RDEPEND}"
```
krnowak, line 33: The sys-apps/checkpolicy is originally placed in BDEPEND, why moving it back to RDEPEND?

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
```diff
@@ -29,34 +28,27 @@ HOMEPAGE="https://wiki.gentoo.org/wiki/Project:SELinux"
...
 
+S=${WORKDIR}/
 
-# flatcar changes
 PATCHES=(
 	"${FILESDIR}"/0001-policy-modules-kernel-all-more-actions-for-kernel.patch
-	"${FILESDIR}"/0001-policy-ms-MCS-restricts-relabelfrom.patch
 	"${FILESDIR}"/icmp-bind.patch
+	"${FILESDIR}"/files-relabel.patch
 )
```
krnowak, line 44: I think we probably could have this patches as user patches in `<coreos-overlay>/coreos/user-patches/sec-policy/selinux-base` directory.

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
```diff
@@ -132,7 +122,7 @@ src_compile() {
 
 	for i in ${POLICY_TYPES}; do
 		cd "${S}/${i}" || die
-		# flatcar changes
+		emake base
 		emake base BINDIR="${ROOT}/usr/bin" NAME=$i SHAREDIR="${ROOT%/}"/usr/share/selinux \
 			LD_LIBRARY_PATH="${ROOT}/usr/lib64:${LD_LIBRARY_PATH}" -C "${S}"/${i}
```
krnowak, line 127: I think this is just wrong. The ROOT variable can only be used in `pkg_*` functions (see https://devmanual.gentoo.org/ebuild-writing/variables/index.html).

Do you know why we have this change? Could you add an inline comment like `# Flatcar: yadda yadda`?

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
```diff
@@ -167,23 +165,20 @@ src_install() {
...
 	insinto /usr/lib/selinux
 	doins "${FILESDIR}/config"
 
 	insinto /etc/selinux/mcs/contexts
 	doins "${FILESDIR}/lxc_contexts"
 
-	# flatcar changes
 	mkdir -p "${D}/usr/lib/selinux"
 	for i in ${POLICY_TYPES}; do
 		mv "${D}/etc/selinux/${i}" "${D}/usr/lib/selinux"
```
krnowak, line 184: Probably something for coreos-base/misc-files too.

krnowak (commented):
Thanks for working on it. My review was mostly focused on finding a way to avoid modifications in the ebuilds, so we could put them under the package automation. We probably don't need to solve it right now in this PR, but I still had some questions.

There are issues though, mostly with the use of the ROOT variable outside of the `pkg_*` functions.

.github/workflows/portage-stable-packages-list
```diff
@@ -454,6 +458,7 @@ sys-libs/libnvme
 sys-libs/libseccomp
 sys-libs/ncurses
 sys-libs/readline
+sys-libs/libselinux
```
krnowak, line 461:
```suggestion
sys-libs/libselinux
sys-libs/libsepol
```

.github/workflows/portage-stable-packages-list
```diff
@@ -362,8 +362,12 @@ profiles
 #
 # scripts
 
+sec-policy/selinux-base-policy
+sec-policy/selinux-container
```
krnowak, line 366: How about these additions too?

```suggestion
sec-policy/selinux-base
sec-policy/selinux-base-policy
sec-policy/selinux-container
sec-policy/selinux-dbus
sec-policy/selinux-sssd
sec-policy/selinux-unconfined
```

.github/workflows/portage-stable-packages-list
krnowak, line 1: I think we could also add `eclass/selinux-policy-2.eclass` to the list.

krnowak (changes requested):
Thanks for the update. I had some more questions though. Also:

- The `sys-apps/policycoreutils: apply flatcar changes` commit could be reworded to mention moving selinux policy directory from `/etc/selinux/policy` to `/usr/lib/selinux/policy`.
- The `sec-policy/selinux-base: apply Flatcar patches` commit could be reworded to `coreos/user-patches: yadda yadda` header, since it does not modify the `sec-policy/selinux-base` (other than dropping unused ebuilds).
  - Which also means that maybe we could move `sec-policy/selinux-base` to portage-stable too?
- The `sys-libs/libsemanage: apply flatcar patches` commit could be reworded to describe the actual changes made (dropping python deps, setting SHLIBDIR for installation).
- Where is ssh selinux module defined? Should we add a dependency on `sec-policy/selinux-dbus` to the package that provides it?

sdk_container/src/third_party/coreos-overlay/sec-policy/selinux-base/selinux-base-2.20221101-r4.ebuild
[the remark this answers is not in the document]
krnowak: Maybe not related, but I remember getting weird issues when I was testing some packages with `ebuild`. Basically I was doing `ebuild <path to ebuild> install`, check the temporary installation directory, fix something in the ebuild, removed the ebuild installed flag file (`rm -f <whatever weird path>/.installed`) and did `ebuild <path to ebuild> install` again. I actually needed to do `ebuild <path to ebuild> clean` every time instead of removing the flag file.

Anyway, water under the bridge, things seem to be working for you now.

krnowak:
> @krnowak thanks for the review - addressed your comments except the [nitpick](https://github.com/flatcar/scripts/pull/917#discussion_r1326971493) as it seems we still need to pull anyway `setools` into the overlay? I don't think it's worth as we still need to maintain another downstream patch for `policycoreutils`.
> 

Fair, we can tackle it when we will be able to move policycoreutils to portage-stable.

> A build has been kicked off.
> 
> Finally, to answer your question:
> 
> > Where is ssh selinux module defined? Should we add a dependency on sec-policy/selinux-dbus to the package that provides it?
> 
> I already asked myself, `ssh` module is part of the `selinux-base-policy` (https://gitweb.gentoo.org/repo/gentoo.git/tree/sec-policy/selinux-base-policy/selinux-base-policy-2.20221101-r4.ebuild#n30) so we can't really add a dependency for `ssh` in particular but we could add a dependency to `selinux-base-policy` to `selinux-dbus` but I fear of falling into a cyclic dependency.

Meh. selinux-dbus should be deprecated and dbus module moved to selinux-base-policy. Anyway, won't be blocking this PR on this kind of stuff, obviously.

.github/workflows/portage-stable-packages-list
```diff
@@ -363,8 +364,17 @@ profiles
...
+sec-policy/selinux-dbus
+sec-policy/selinux-sssd
+sec-policy/selinux-unconfined
+
+
 sys-apps/acl
 sys-apps/attr
+sys-apps/checkpolicy
```
krnowak, line 377: One last thing - could you add `sys-apps/semodule-utils` back to the list too? I remember that it used to be in the automation list, but I dropped it as the update broke with the older version of selinux packages. Now it should be fine to bring it back.

This change should not require the CI rerun as it does not affect the generated images at all.

krnowak (approved):
Thanks for your work on selinux, I really appreciate it. There was one small thing that I missed earlier. With it fixed, this can go in.

PR #919, opened 2023-06-14
Upload dev-key-signed OEM sysext update payload

pothos:
For release tests and updating a machine to a dev build we already have the dev-key-signed generic update payload but not yet the OEM sysext update payload.
Generate the dev-key-signed OEM sysext update payload during build and upload it.

## How to use

From ue-rs (or with `protoc` when starting from the right offset)

## Testing done

TODO: check if it gets uploaded to bincache with a Jenkins build

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -579,6 +579,11 @@ install_oem_sysext() {
     for to_move in "${built_sysext_dir}/${oem_sysext}"*; do
         mv "${to_move}" "${upload_dir}/${to_move##*/}"
     done
+    # Generate dev-key-signed update payload for testing
+    delta_generator \
+      -private_key "/usr/share/update_engine/update-payload-key.key.pem" \
+      -new_image "${upload_dir}/${built_sysext_filename}" \
+      -out_file "${upload_dir}/flatcar_test_update-${oem_sysext}.gz"
     # Remove sysext_dir if building sysext and installing it
     # succeeded.
     rm -rf "${built_sysext_dir}"


pothos:
Here the decoded DeltaArchiveManifest:
```
partition_operations {
  type: REPLACE_BZ
  data_offset: 0
  data_length: 1417
  dst_extents {
    start_block: 0
    num_blocks: 256
  }
  data_sha256_hash: "o\207\243\220\363\320\372Y\2602\3208\360\"!p\216[\351h\37562\356\013k{OU\252\235a"
}
noop_operations {
  type: REPLACE
  data_offset: 1417
  data_length: 264
  dst_extents {
    start_block: 18446744073709551615
    num_blocks: 1
  }
}
block_size: 4096
signatures_offset: 1417
signatures_size: 264
new_partition_info {
  size: 4096
  hash: "\327\033\217#\004\305\252\207L\036!\360\345\"\215_\'\251\351\301\000U\214!`\024{a1q\312\320"
}
```
The `REPLACE_BZ` vs `REPLACE` type gets decided by delta_generator based on the resulting size of the compressed data: If compression doesn't help, it gets stored uncompressed. In this case, the qemu squashfs was reduced by compression because of the zero padding to fill the block.

We could extend the protocol with a new InstallProcedure type where we place the sysext instead of reusing the InstallOperation but I think it's ok to use the file name in Omaha to know what to do with the payload.

build_library/vm_image_util.sh
```diff
@@ -579,6 +579,11 @@ install_oem_sysext() {
     for to_move in "${built_sysext_dir}/${oem_sysext}"*; do
         mv "${to_move}" "${upload_dir}/${to_move##*/}"
     done
+    # Generate dev-key-signed update payload for testing
+    delta_generator \
+      -private_key "/usr/share/update_engine/update-payload-key.key.pem" \
+      -new_image "${upload_dir}/${built_sysext_filename}" \
+      -out_file "${upload_dir}/flatcar_test_update-${oem_sysext}.gz"
```
krnowak, line 586: One question that comes to my mind is how this delta_generator figures out what to base the delta on? There's `-new_image`, but for deltas (diffs) you usually need two images, right?

dongsupark: delta_generator has 2 different modes of generation, full update and partial delta update.

If an option `-old_image` is given, a partial delta update as you describe. If not given, it means implicitly a full update, so falls into the following code block.
https://github.com/flatcar/update_engine/blob/c6f566d47d8949632f7f43871eb8d5c625af3209/src/update_engine/delta_diff_generator.cc#L1436
I know, the code is so tricky. ;-)

krnowak: Thanks!

krnowak (approved):
I'd say it's ok, but eventually we will probably want to limit that only to some platforms instead of all of them, as our tests probably won't cover them all. I suppose that this is for kola to consume, right?

pothos:
Kola and for users that switch a machine to a dev build with `flatcar-update` to reproduce a bug.

PR #920, opened 2023-06-14
Build_sysext script to enable sysext builds in the OS image build process

krishjainx:
# Build_sysext script to enable sysext builds in the OS image build process

Included a script to enable generating systemd-sysexts. Successfully tested sysext generation (manually; script refinement in progress) with a fresh Flatcar image (e.g., Python and Neofetch system extension). Part of my internship work.

Link to the related issue: [Flatcar Issue #1052](https://github.com/flatcar/Flatcar/issues/1052).

# How to use 

The current OS images we provide are not OK as base for flatcar specific sysext images: it lacks the package metadata and portage configuration, in order to keep end user OS image clean. This script retains this information and allows you to produce systemd-sysexts to extend the system. This script can be used to build a Flatcar sysext image. Recommended to run from image build folder.

```
sudo build_sysext --board=amd64-usr command-extras package_1 package_2 package_3 - builds sysext named python-sysext with package_1, package_2 and package_3 within it  for amd64 architecture.
sudo build_sysext --board=arm64-usr command-extras package_1 package_2 package_3 - builds sysext named python-sysext with package_1, package_2 and package_3 within it  for arm64 architecture.

```

# Testing done

I have built multiple sysext images using the script and have been able to successfully extend a fresh upstream Flatcar image with packages such as Python, Neofetch, etc. I provided a Butane configuration similar to the one below, transpiled it to Ignition, and booted with it. I received a review and advice on how to make the script more robust and meet the demands of the Flatcar team.

```
variant: flatcar
version: 1.0.0
storage:
  files:
    - path: /etc/extensions/python.raw
      mode: 0644
      contents:
        source: https://0x0.st/HQpJ.raw

```

Notes:

> A problem with installing regular packages is that in many cases the software won't correctly work because only /usr is added. Files on /etc will be missing and the approach in the Azure OEM was that we ship the needed files in the sysext and then set up symlinks before, e.g., a service starts (We could build a confext in the future but this doesn't work now). Also, files in /var will be missing. For some of these files there might be tmpfile directives but they won't get executed on load. Finally, the services that use .wants symlinks to be enabled would currently work because we have deployed a workaround but we plan to drop this because the recommended way is to use Upholds= instead of Wants= (currently through drop-ins for the multi-user target but in systemd 254 it should work to use .upholds symlinks). In summary, we need some postprocessing for some scenarios and call out what won't work for the rest.

"I think this should be addressed to some degree by the --manglefs_script option" - @krnowak  

Edit: A great use case for this would be to build a Docker and containerd sysext to replace torcx.

### DIFF
--- a/build_library/prod_image_util.sh
+++ b/build_library/prod_image_util.sh
@@ -78,6 +78,7 @@ create_prod_image() {
   local image_initrd_contents_wtd="${image_name%.bin}_initrd_contents_wtd.txt"
   local image_disk_usage="${image_name%.bin}_disk_usage.txt"
   local image_pkgdb="${image_name%.bin}_pkgdb.tar.xz"
+  local image_sysext_base="${image_name%.bin}_sysext.squashfs"
 
   start_image "${image_name}" "${disk_layout}" "${root_fs_dir}" "${update_group}"
 
@@ -102,7 +103,11 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
+  sudo cp -a "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2"
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc"
+  sudo mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/${image_sysext_base}" -noappend
+  sudo rm -rf "${BUILD_DIR}/root_fs_dir2"
+  
   # clean-ups of things we do not need
   sudo rm ${root_fs_dir}/etc/csh.env
   sudo rm -rf ${root_fs_dir}/etc/env.d
@@ -162,6 +167,7 @@ EOF
     "${BUILD_DIR}/${image_initrd_contents}"
     "${BUILD_DIR}/${image_initrd_contents_wtd}"
     "${BUILD_DIR}/${image_disk_usage}"
+    "${BUILD_DIR}/${image_sysext_base}"
   )
 
   local files_to_evaluate=( "${BUILD_DIR}/${image_name}" )
--- a/build_sysext
+++ b/build_sysext
@@ -1,0 +1,171 @@
+#!/bin/bash
+
+# Copyright (c) 2023 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can
+# be found in the LICENSE file.
+#
+# Script to generate sysext. See systemd-sysext(8). Prerequisite is
+# that you've run build_packages and build_image.
+
+SCRIPT_ROOT=$(dirname "$(readlink -f "$0")")
+. "${SCRIPT_ROOT}/common.sh" || exit 1
+
+# Script must run inside the chroot
+assert_inside_chroot
+assert_root_user
+
+# All these are used to set up the 'BUILD_DIR' variable
+DEFINE_string board "${DEFAULT_BOARD}" \
+  "The board to build a sysext for."
+DEFINE_string build_dir '' \
+  "Directory used for building the sysext image, must exist, but should be empty."
+DEFINE_string metapkgs '' \
+  "Comma-separated list of meta-packages to build from source and install into sysext image."
+DEFINE_string manglefs_script '' \
+  "A path to executable that will customize the rootfs of the sysext image."
+DEFINE_string squashfs_base '' \
+  "The path to the squashfs base image."
+
+FLAGS_HELP="USAGE: build_sysext [flags] [sysext name] [binary packages to install into image].
+
+This script is used to build a Flatcar sysext image. 
+
+Examples:
+
+Builds a sysext image named interpreters in the images
+directory with dev-lang/python and dev-lang/perl packages for amd64:
+
+sudo build_sysext \
+     --board=amd64-usr \
+     --build_dir=images \
+     interpreters dev-lang/python dev-lang/perl
+
+Builds a sysext image named oem-azure in the oem-images directory with
+metapackage coreos-base/oem-azure for arm64:
+
+sudo build_sysext \
+     --board=arm64-usr \
+     --build_dir=oem-images \
+     --metapkgs=coreos-base/oem-azure \
+     --mangle_fs=…/coreos-base/oem-azure/files/manglefs.sh \
+     oem-azure
+
+"
+
+show_help_if_requested "$@"
+
+# Parse command line
+FLAGS "$@" || exit 1
+
+eval set -- "${FLAGS_ARGV}"
+
+. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
+
+if [[ -z "${FLAGS_build_dir}" ]]; then
+  die "Need a build directory to be specified with a --build_dir option"
+fi
+
+BUILD_DIR=$(realpath "${FLAGS_build_dir}")
+
+# Architecture values are taken from systemd.unit(5).
+declare -A SYSEXT_ARCHES
+SYSEXT_ARCHES['amd64-usr']='x86-64'
+SYSEXT_ARCHES['arm64-usr']='arm64'
+
+declare -r SYSEXT_ARCHES
+
+# Usage: _get_sysext_arch board [board...]
+_get_sysext_arch() {
+    local board=${1}
+    if [[ ${#SYSEXT_ARCHES["${board}"]} -ne 0 ]]; then
+        echo "${SYSEXT_ARCHES["${board}"]}"
+    else
+        die "Unknown board '${board}'"
+    fi
+}
+
+
+set -euo pipefail
+
+cleanup() {
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" 2>/dev/null || true
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
+}
+
+if [[ ${#} -lt 1 ]]; then
+   show_help_if_requested -h
+  die 'Expected at least one parameter for sysext image name'
+fi
+
+SYSEXTNAME="${1}"
+shift
+
+# Set up trap to execute cleanup() on script exit
+trap cleanup EXIT
+
+ARCH=$(_get_sysext_arch "${FLAGS_board}")
+if [[ -z "${FLAGS_squashfs_base}" ]]; then
+  FLAGS_squashfs_base="${BUILD_DIR}/flatcar_production_image_sysext.squashfs"
+fi
+cleanup
+
+mkdir "${BUILD_DIR}/fs-root"
+mount -rt squashfs -o loop,nodev "${FLAGS_squashfs_base}" "${BUILD_DIR}/fs-root"
+mkdir "${BUILD_DIR}/install-root"
+mkdir "${BUILD_DIR}/workdir"
+mount -t overlay overlay -o lowerdir="${BUILD_DIR}/fs-root",upperdir="${BUILD_DIR}/install-root",workdir="${BUILD_DIR}/workdir" "${BUILD_DIR}/install-root"
+VERSION_BOARD=$(grep "^VERSION=" ${BUILD_DIR}/fs-root/usr/lib/os-release | cut -d = -f 2-)
+if [ "$VERSION_BOARD" != "$FLATCAR_VERSION" ]; then
+  echo "$VERSION_BOARD"
+  echo "$FLATCAR_VERSION"
+  echo "Version mismatch between board flatcar release and SDK container flatcar release"
+  exit 1
+fi
+
+if [[ -n "${FLAGS_metapkgs}" ]]; then
+  mapfile -t metapkgs < <(tr ',' '\n' <<<"${FLAGS_metapkgs}")
+  "emerge-${FLAGS_board}" --nodeps --buildpkgonly --usepkg n --verbose "${metapkgs[@]}"
+  set -- "${metapkgs[@]}" "${@}"
+fi
+
+for package; do
+  echo "Installing package into sysext image: $package"
+  FEATURES="-ebuild-locks" emerge \
+    --root="${BUILD_DIR}/install-root" \
+    --config-root="/build/${FLAGS_board}"  \
+    --sysroot="/build/${FLAGS_board}"  \
+    --root-deps=rdeps \
+    --usepkgonly \
+    --verbose \
+    "${package}"
+done
+
+# Unmount in order to get rid of the overlay
+umount "${BUILD_DIR}/install-root"
+umount "${BUILD_DIR}/fs-root"
+
+if [[ -n "${FLAGS_manglefs_script}" ]]; then
+  if [[ ! -x "${FLAGS_manglefs_script}" ]]; then
+    die "${FLAGS_manglefs_script} is not executable"
+  fi
+  "${FLAGS_manglefs_script}" "${BUILD_DIR}/install-root"
+fi
+
+info "Removing non-/usr directories from sysext image"
+for entry in "${BUILD_DIR}/install-root"/*; do
+  if [[ "${entry}" = */usr ]]; then
+    continue
+  fi
+  info "  Removing ${entry##*/}"
+  rm -rf "${entry}"
+done
+mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+version_field="${VERSION_FIELD_OVERRIDE:-VERSION_ID=${FLATCAR_VERSION_ID}}"
+all_fields=(
+  'ID=flatcar'
+  "${version_field}"
+  "ARCHITECTURE=${ARCH}"
+)
+printf '%s\n' "${all_fields[@]}" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+mksquashfs "${BUILD_DIR}/install-root" "${BUILD_DIR}/${SYSEXTNAME}.raw"
+rm -rf "${BUILD_DIR}"/{fs-root,install-root,workdir}


build_library/prod_image_util.sh
```diff
@@ -101,7 +102,8 @@ create_prod_image() {
           || die_notrace "coreos-au-key is missing the 'official' use flag"
   fi
 
-  tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
```
t-lo, line 104: I suggest we keep this line for the time being - it's used for the current Azure OEM sysext build.
While keeping the line adds redundancy, removing it would require us us to also update the Azure OEM sysext build script to use `build_sysext` instead of the system it currently uses.
Keeping the line would enable us to update the Azure OEM sysext build in a follow-up PR instead.

build_library/prod_image_util.sh
```diff
@@ -101,7 +102,8 @@ create_prod_image() {
...
+   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
+  mksquashfs "${root_fs_dir}" "${BUILD_DIR}/configroot" "${BUILD_DIR}"/flatcar_production_image_sysext.sqsh
```
pothos, line 106: The difference of this image and the final image is that the final image has some folders removed and other postprocessing. Instead of archiving one more big artifact, could we rather mount the built `/usr` partition as first lowerdir and as second lowerdir the slim squashfs with the `var/cache/edb` and `var/db/pkg` folders which we could create here? The main difference would be to have either the tar or the squashfs. (You could maybe even not create the squashfs and rather use the existing tar by extracting to a temporary lowerdir).

t-lo: While we would save a few bytes with an incremental approach it would make the whole system more feeble - particularly if we consider making the squashfs a new build artifact to be published for general use (e.g. enabling downstream users to build custom sysexts). I actually like the idea of having a single self-contained squashfs; it makes the approach more generally usable.

With an incremental approach I see 2 problems we wouldn't have with a self-contained image:
1. the squashfs would require the "right" base image to work. We would need to introduce a versioning mechanism to ensure `build_sysext`, e,g, when used on a published release instead of an intermediate step at build time, uses the correct base. A self-contained image on the other hand would ship everything in the right versions - and it would even include a version file itself which `build_sysext` could optionally use to pin a sysext to a Flatcar release.
2. we would need to keep this `tar` line up to date to capture the full state of the OS image before clean-up, introducing additional (manual) maintenance burden. Even today that line doesn't capture all of the state - see the [rm lines right after tar](https://github.com/flatcar/scripts/blob/main/build_library/prod_image_util.sh#L106-L111). This may have subtle side effects in the future depending on where dependencies keep meta-information required for packages at installation time.

pothos: The usage would be from the SDK and there it already needs to match the right version, or?
If possible, we could reuse the `/usr` partition artifact as first lowerdir: https://stable.release.flatcar-linux.net/amd64-usr/current/flatcar_production_update.bin.bz2 but when it's too complicated, I'm also fine.

build_library/prod_image_util.sh
```diff
@@ -101,7 +102,8 @@ create_prod_image() {
           || die_notrace "coreos-au-key is missing the 'official' use flag"
   fi
 
-  tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
+   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
```
pothos, line 105: Is the spacing off?

build_library/prod_image_util.sh
```diff
@@ -146,7 +148,8 @@ EOF
       "${image_kconfig}" \
       "${image_initrd_contents}" \
       "${image_initrd_contents_wtd}" \
-      "${image_disk_usage}"
+      "${image_disk_usage}" \
+      "${image_sysext_base}"
```
pothos, line 152: Why is this added here?

build_sysext
```diff
@@ -0,0 +1,76 @@
...
+. "${BUILD_LIBRARY_DIR}/test_image_content.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/torcx_manifest.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/vm_image_util.sh" || exit 1
+
+BUILD_DIR="${DEFAULT_BUILD_ROOT}/images/${BOARD}/${IMAGE_SUBDIR}"
+
+# export SYSEXTNAME=$2
+
+# # with my 1 line patch before the rms in build_library/prod_image_util.sh (mksquashfs "${root_fs_dir}" "${BUILD_DIR}/configroot" "${BUILD_DIR}"/flatcar_production_image_sysext.sqsh)
+sudo mkdir ${BUILD_DIR}/fs-root
```
pothos, line 61: Here and in the lines below we should quote variables.
We should also support rerunning and handle the case that this folder exists or that the mount is still set up.
The best would be to run `umount` first to be safe and also set up a `trap` to `umount` when the script exits.

build_library/prod_image_util.sh
pothos: We can also check that the Flatcar version matches, it has to be the same anyway and this will prevent wrong usage.

pothos:
A problem with installing regular packages is that in many cases the software won't correctly work because only `/usr` is added.
Files on `/etc` will be missing and the approach in the Azure OEM was that we ship the needed files in the sysext and then set up symlinks before, e.g., a service starts (We could build a confext in the future but this doesn't work now).
Also, files in `/var` will be missing. For some of these files there might be tmpfile directives but they won't get executed on load.
Finally, the services that use `.wants` symlinks to be enabled would currently work because we have deployed a workaround but we plan to drop this because the recommended way is to use `Upholds=` instead of `Wants=` (currently through drop-ins for the multi-user target but in systemd 254 it should work to use `.upholds` symlinks).
In summary, we need some postprocessing for some scenarios and call out what won't work.

build_library/prod_image_util.sh
pothos: It's gone again

build_sysext
```diff
@@ -0,0 +1,103 @@
...
+    sudo rm -rf "${BUILD_DIR}/{fs-root, install-root, workdir}"
+
+upload_image "${BUILD_DIR}/${SYSEXTNAME}.raw"
+
+else
+    echo "Version mismatch"
+fi
+
+# Set up trap to execute cleanup() on script exit
+trap cleanup EXIT
```
pothos, line 103: The trap should be set up before anything is done

build_sysext
```diff
@@ -0,0 +1,103 @@
...
+
+
+# export SYSEXTNAME=$2
+
+ARCH=[[ "$BUILD_DIR" == *arm64-usr* ]] && ARCH=arm64 || [[ "$BUILD_DIR" == *amd64-usr* ]] && ARCH=x86-64
+
+source /etc/os-release
+
+
+unmount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
```
pothos, line 72: First you would need to create this folder and then I think you meant `|| true`

build_sysext
```diff
@@ -0,0 +1,103 @@
...
+
+
+# # with my 1 line patch before the rms in build_library/prod_image_util.sh (mksquashfs "${root_fs_dir}" "${BUILD_DIR}/configroot" "${BUILD_DIR}"/flatcar_production_image_sysext.sqsh)
+
+VERSION_BOARD=$(awk -F= '$1=="FLATCAR_VERSION_ID" { print $2 ;}' "${BUILD_DIR}/version.txt")
+
+if [ "$VERSION_BOARD" == "$VERSION_ID" ]; then
+    sudo mkdir "${BUILD_DIR}/fs-root"
+    sudo mount -rt squashfs -o loop,nodev,noexec "${BUILD_DIR}/flatcar_production_image_sysext.sqsh" "${BUILD_DIR}/fs-root"
```
pothos, line 81: At this point we should check the `/usr/lib/os-release` file for the correct version.

build_sysext
```diff
@@ -0,0 +1,103 @@
...
+    sudo rm -rf install-root/{dev, etc, var}
+    sudo mkdir -p install-root/usr/lib/extension-release.d
+    sudo printf "ID=flatcar-releaseinfo\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}" > "${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+    sudo mksquashfs install-root "${BUILD_DIR}/${SYSEXTNAME}.raw" --all-root
+
+    sudo rm -rf "${BUILD_DIR}/{fs-root, install-root, workdir}"
+
+upload_image "${BUILD_DIR}/${SYSEXTNAME}.raw"
+
+else
```
pothos, line 98: Small suggestion, if you invert the match you can exit early and don't need to have the whole block above in the if clause

build_sysext
t-lo: Oh, I get Kai's point - we could mount the squashfs and extract `FLATCAR_VERSION_ID` from there (`...fs-root/usr/lib/os-release`) instead of using `version.txt` from `${BUILDDIR}`. 
`version.txt` is not guaranteed to be in the build directory - for instance if a user downloads the sysext squasfs of a release to build a custom sysext instead of running a full build themselves.

build_sysext
pothos: I would put this action after `mkdir -p` and before the corresponding `mount`, but you could also leave it here and suppress the error messages.

build_sysext
pothos: The field to extract from in the `os-release` file is `VERSION` and this should match the version the SDK has built/would build.
The field `VERSION_ID` for the extension release file is the `VERSION_ID` field from the `os-release` file (that's missing the build ID but that's Flatcar-specific and not supported for matching - means there are corner cases where the sysext isn't rejected but we can't prevent this).

jepio:
@krishjainx
can you paste a log showing the output you see when you execute `./build_sysext` (with whatever package you're testing with)?

Please try running the build two times in a row to make sure things don't break when this is done.

build_sysext
```diff
@@ -0,0 +1,103 @@
...
+  sudo umount "${BUILD_DIR}/install-root"
+  sudo umount "${BUILD_DIR}/fs-root"
+  sudo rm -rf install-root/{dev, etc, var}
+  sudo mkdir -p install-root/usr/lib/extension-release.d
+  sudo printf "ID=flatcar-releaseinfo\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}" > "${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+  sudo mksquashfs install-root "${BUILD_DIR}/${SYSEXTNAME}.raw" --all-root
+
+  sudo rm -rf "${BUILD_DIR}/{fs-root, install-root, workdir}"
+
+  upload_image "${BUILD_DIR}/${SYSEXTNAME}.raw"
```
pothos, line 101: If this is also for users, we could skip the upload step and let the build scripts handle it for the release case only?

build_sysext
pothos: As Jeremi said below, we should move the `else` code out and exit in the `if` case with failure.

build_library/prod_image_util.sh
pothos:
```suggestion

```

jepio:
Could you also make the script executable with `chmod`?

build_library/prod_image_util.sh
```diff
@@ -102,8 +103,11 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
-  # clean-ups of things we do not need
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2" 
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc" 
+  mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" -noappend
+  
+# clean-ups of things we do not need
```
pothos, line 110:
```suggestion
  # clean-ups of things we do not need
```
This whitespace change is not needed

build_library/prod_image_util.sh
```diff
@@ -146,7 +150,7 @@ EOF
       "${image_kconfig}" \
       "${image_initrd_contents}" \
       "${image_initrd_contents_wtd}" \
-      "${image_disk_usage}"
+      "${image_disk_usage}" \
```
pothos, line 153:
```suggestion
      "${image_disk_usage}"
```
Not needed

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+
+cleanup() {
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
+  rm "${BUILD_DIR}/${SYSEXTNAME}.raw"
+}
+
+# Set up trap to execute cleanup() on script exit
+
+# trap cleanup EXIT
```
pothos, line 58: When setting up the trap you need to make sure that `SYSEXTNAME` is defined first.

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+. "${BUILD_LIBRARY_DIR}/cros_vm_constants.sh" || exit 1
+
+BUILD_DIR="/mnt/host/source/src/build/images/amd64-usr/developer-3621.0.0+nightly-20230531-2100-66-gf786da53f9-a1"
+
+echo "${BUILD_DIR}"
+
+cleanup() {
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
+  rm "${BUILD_DIR}/${SYSEXTNAME}.raw"
```
pothos, line 53:
```suggestion
  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
  rm "${BUILD_DIR}/${SYSEXTNAME}.raw" || true
```
We should ignore failure of a command and continue

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+# Set up trap to execute cleanup() on script exit
+
+# trap cleanup EXIT
+
+export SYSEXTNAME=$1
+
+case $BOARD in
+*arm64-usr*) ARCH=arm64 ;;
+*amd64-usr*) ARCH=x86-64 ;;
+*) echo "Unknown board" ;;
```
pothos, line 65: The block should be moved in.
But it would be great to rather reuse the `_get_sysext_arch` function from `oem_sysext_util.sh`

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+esac
+
+source /usr/lib/os-release
```
pothos, line 68: A commend would be good, what variables do you want to be set?
And if it's only one or two, better only use them, e.g.:
```suggestion
source <(grep -P '^(VERSION|VERSION_ID)=' /etc/os-release)
```

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+
+umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
```
pothos, line 70:
```suggestion
cleanup
```
We should run the full cleanup function here in advance.

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+
+mkdir "${BUILD_DIR}/fs-root"
+mount -rt squashfs -o loop,nodev,noexec "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" "${BUILD_DIR}/fs-root"
```
pothos, line 73: Why `noexec`?

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+mkdir "${BUILD_DIR}/workdir"
+mount -t overlay overlay -o lowerdir="${BUILD_DIR}/fs-root",upperdir="${BUILD_DIR}/install-root",workdir="${BUILD_DIR}/workdir" "${BUILD_DIR}/install-root"
+VERSION_BOARD=$(grep "^VERSION=" ${BUILD_DIR}/fs-root/usr/lib/os-release | cut -d = -f 2-)
+if [ "$VERSION_BOARD" != "$VERSION" ]; then
+  echo "$VERSION_BOARD"
+  echo "$VERSION"
+  echo "Version mismatch between board flatcar release and SDK container flatcar release"
+  exit 1
+else
+  /usr/local/bin/emerge-${BOARD} --jobs=${NUM_JOBS} --getbinpkg "${SYSEXTNAME}"
```
pothos, line 84: Is `emerge-$BOARD` not in the PATH?

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+  ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="/build/${BOARD}" FEATURES="-ebuild-locks" emerge --root-deps=rdeps --usepkgonly --verbose "${SYSEXTNAME}"
+  umount "${BUILD_DIR}/install-root" || true
+  umount "${BUILD_DIR}/fs-root" || true
```
pothos, line 87: Wouldn't be needed if `cleanup` is used at the start

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || exit 1
+  rm "${BUILD_DIR}/${SYSEXTNAME}.raw"
+}
+
+# Set up trap to execute cleanup() on script exit
+
+# trap cleanup EXIT
+
+export SYSEXTNAME=$1
```
pothos, line 60: I would distinguish between package name and sysext name. Also this needs to be quoted. I recommend to take the first argument as sysext name and the following ones as package names.

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+mount -t overlay overlay -o lowerdir="${BUILD_DIR}/fs-root",upperdir="${BUILD_DIR}/install-root",workdir="${BUILD_DIR}/workdir" "${BUILD_DIR}/install-root"
+VERSION_BOARD=$(grep "^VERSION=" ${BUILD_DIR}/fs-root/usr/lib/os-release | cut -d = -f 2-)
+if [ "$VERSION_BOARD" != "$VERSION" ]; then
+  echo "$VERSION_BOARD"
+  echo "$VERSION"
+  echo "Version mismatch between board flatcar release and SDK container flatcar release"
+  exit 1
+else
+  /usr/local/bin/emerge-${BOARD} --jobs=${NUM_JOBS} --getbinpkg "${SYSEXTNAME}"
+  ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="/build/${BOARD}" FEATURES="-ebuild-locks" emerge --root-deps=rdeps --usepkgonly --verbose "${SYSEXTNAME}"
```
pothos, line 85: Can you explain which command does what and why two are needed?

build_sysext
pothos: Ah, no, you umount because you want to get rid of the overlay, please add a comment.

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+  umount "${BUILD_DIR}/install-root" || true
+  umount "${BUILD_DIR}/fs-root" || true
+  rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var"
```
pothos, line 88: Maybe also remove `/opt` or anything that's not `/usr`?

build_sysext
pothos: Is it commented for testing purposes? You could guard this through a debug env var.

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+
+DEFINE_string board "${DEFAULT_BOARD}" \
+  "The board to build an image for."
+
+FLAGS_HELP="USAGE: build_sysext [flags] [package to build].
+This script is used to build a Flatcar sysext. 
+
+Examples:
+
+build_sysext --board=amd64 python - builds python sysext for amd64 architecture.
```
pothos, line 23:
```suggestion
sudo build_sysext --board=amd64 python - builds python sysext for amd64 architecture.
```
Can you add a few more and also describe how the resulting file is named and where it will be stored?

build_sysext
```diff
@@ -0,0 +1,94 @@
...
+*arm64-usr*) ARCH=arm64 ;;
+*amd64-usr*) ARCH=x86-64 ;;
+*) echo "Unknown board" ;;
+esac
+
+source <(grep -P '^(VERSION|VERSION_ID)=' /etc/os-release)
+
+cleanup
+
+mkdir "${BUILD_DIR}/fs-root"
```
pothos, line 72: In what directory is this script supposed to be run? Can we add checks and docs for that (and maybe let the script switch to the right folder by itself)?

pothos:
Normally we have the _How to use_ and _Testing done_ sections in the PR description. Please readd them and document the testing done and how to use it with a few more examples (best would be to align it with the usage section to keep it for the future).
As mentioned in the review we should decouple the sysext name from the package names to be able to add more packages and define an own naming for the resulting file.

build_sysext
pothos: Ok, if it has no assumptions that's good. Maybe it can't be run under the mounted `/usr` image during image build because of space reasons but I would say that the recommended folder should be the image build folder and document that.

build_sysext
```diff
@@ -0,0 +1,100 @@
...
+DEFINE_integer build_attempt 1 \
+  "The build attempt for this build."
+
+FLAGS_HELP="USAGE: build_sysext [flags] [package to build].
+This script is used to build a Flatcar sysext. 
+
+Examples:
+
+sudo build_sysext --board=amd64-usr command-extras package_1 package_2 package_3 - builds sysext named python-sysext with package_1, package_2 and package_3 within it  for amd64 architecture.
+sudo build_sysext --board=arm64-usr command-extras package_1 package_2 package_3 - builds sysext named python-sysext with package_1, package_2 and package_3 within it  for arm64 architecture.
```
pothos, line 30:
```suggestion
Build a systemd-sysext image named command-extras with package_1, package_2 and package_3 for amd64:
sudo build_sysext --board=amd64-usr command-extras package_1 package_2 package_3

Builds a systemd-sysext image named command-extras with package_1, package_2 and package_3 for arm64:
sudo build_sysext --board=arm64-usr command-extras package_1 package_2 package_3
```

build_sysext
```diff
@@ -0,0 +1,100 @@
...
+. "${BUILD_LIBRARY_DIR}/torcx_manifest.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/vm_image_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/cros_vm_constants.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/oem_sysext_util.sh" || exit 1
+
+BUILD_DIR="/mnt/host/source/src/build/images/amd64-usr/developer-3621.0.0+nightly-20230531-2100-66-gf786da53f9-a1" # TO FIX! Hardcoded currently
+
+cleanup() {
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
```
pothos, line 58: The order is wrong, first you need to unmount, then remove.

build_sysext
pothos: Please remove the `else`, it's not needed because the rest can run as normal block.
```suggestion
fi
```

build_sysext
```diff
@@ -0,0 +1,106 @@
...
+. "${BUILD_LIBRARY_DIR}/vm_image_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/cros_vm_constants.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/oem_sysext_util.sh" || exit 1
+
+cleanup() {
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
+}
+
+export SYSEXTNAME="$1"
```
pothos, line 62:
```suggestion
export SYSEXTNAME="${1-}"
```
To handle the case where no argument was passed we can default to the empty string and it will be handled by your check below.

pothos: By the way, why is it exported?

build_sysext
```diff
@@ -0,0 +1,106 @@
...
+  
+  # Unmount in order to get rid of the overlay
+  umount "${BUILD_DIR}/install-root"
+  umount "${BUILD_DIR}/fs-root"
+  rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var" "${BUILD_DIR}/install-root/opt"
+  mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+  printf "ID=flatcar\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+  mksquashfs "${BUILD_DIR}/install-root" "${BUILD_DIR}/${SYSEXTNAME}.raw" -all-root
+  rm -rf "${BUILD_DIR}/{fs-root, install-root, workdir}"
+  upload_image "${BUILD_DIR}/${SYSEXTNAME}.raw"
```
pothos, line 106: The leading whitespace in this block can go

pothos, line 106: Let's make the upload not the default but opt-in through a command line flag

build_sysext
```diff
@@ -0,0 +1,106 @@
...
+assert_root_user
+
+DEFINE_string board "${DEFAULT_BOARD}" \
+  "The board to build a sysext for."
+DEFINE_string output_root "${DEFAULT_BUILD_ROOT}/images" \
+  "Directory in which to place sysext images and intermediate files"
+DEFINE_string group "developer" \
+  "The normal image build group"
+DEFINE_integer build_attempt 1 \
+  "The build attempt for this build."
```
pothos, line 22: These aren't used
```suggestion

```

build_sysext
```diff
@@ -0,0 +1,106 @@
...
+  echo "$VERSION_BOARD"
+  echo "$VERSION"
+  echo "Version mismatch between board flatcar release and SDK container flatcar release"
+  exit 1
+fi
+  shift
+  
+  for package in "$@"; do
+  	echo "Installing package into sysext image: $package"
+  	ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="/build/${BOARD}" FEATURES="-ebuild-locks" emerge --root-deps=rdeps --usepkgonly --verbose "${package}"
```
pothos, line 95: Where is `BOARD` defined? I think you want to use `${FLAGS_board}`?

build_sysext
```diff
@@ -0,0 +1,106 @@
...
+if [ -z "$SYSEXTNAME" ]; then
+    echo "Error: SYSEXTNAME environment variable is not set."
+    echo "Try build_sysext --help for usage notes"
+    exit 1
+fi
+
+# Set up trap to execute cleanup() on script exit
+trap cleanup EXIT
+
+ARCH=$(_get_sysext_arch "$board")
```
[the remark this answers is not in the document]
pothos: Or since you source `board_options.sh` you don't need to redefine the flag and can use `BOARD`

build_sysext
pothos:
```suggestion
DEFINE_string board "${DEFAULT_BOARD}" \
  "The board to build a sysext for."
DEFINE_string output_root "${DEFAULT_BUILD_ROOT}/images" \
  "Directory in which to place sysext images and intermediate files"
DEFINE_string group "developer" \
  "The normal image build group"
DEFINE_integer build_attempt 1 \
  "The build attempt for this build."
```
We don't need all of these, they are either unused or already defined by `board_options.sh`

build_sysext
pothos: Not resolved

build_sysext
```diff
@@ -0,0 +1,100 @@
...
+sudo build_sysext --board=arm64-usr command-extras package_1 package_2 package_3
+
+"
+
+show_help_if_requested "$@"
+
+# Parse command line
+FLAGS "$@" || exit 1
+
+eval set -- "${FLAGS_ARGV:-prod}"
```
pothos, line 36: What is the purpose of this line?

build_sysext
```diff
@@ -0,0 +1,100 @@
...
+
+. "${BUILD_LIBRARY_DIR}/release_util.sh" || exit 1
+
+. "${BUILD_LIBRARY_DIR}/toolchain_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/build_image_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/prod_image_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/dev_container_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/test_image_content.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/torcx_manifest.sh" || exit 1
```
pothos, line 46: Why is this imported? Please only import what you use.
Also note that if you want to rely on a flag that is defined here, the import would need to be moved before the flag parsing. For example, if the user should be able to set `BUILD_DIR`.

build_sysext
```diff
@@ -0,0 +1,98 @@
...
+# Parse command line
+FLAGS "$@" || exit 1
+
+eval set -- "${FLAGS_ARGV:-prod}"
+
+. "${BUILD_LIBRARY_DIR}/release_util.sh" || exit 1
+
+. "${BUILD_LIBRARY_DIR}/toolchain_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/build_image_util.sh" || exit 1
```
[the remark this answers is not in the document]
pothos: Yes, it was there before, sorry for my wrong suggestion to remove it - please add it back with a comment on why it's there/what this does ;)

build_sysext
pothos: What we need for sure is first `build_library/board_options.sh` and then `build_library/build_image_util.sh` to set up `BUILD_DIR`. And `build_library/oem_sysext_util.sh` for the helper.
What about the rest, where is it used?

build_sysext
```diff
@@ -0,0 +1,108 @@
...
+
+# Unmount in order to get rid of the overlay
+umount "${BUILD_DIR}/install-root"
+umount "${BUILD_DIR}/fs-root"
+rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var" "${BUILD_DIR}/install-root/opt"
+mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+printf "ID=flatcar\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+mksquashfs "${BUILD_DIR}/install-root" "${BUILD_DIR}/${SYSEXTNAME}.raw" -all-root
+rm -rf "${BUILD_DIR}/{fs-root, install-root, workdir}"
+upload_image "${BUILD_DIR}/${SYSEXTNAME}.raw"
```
pothos, line 108:
```suggestion
```
Let's remove it for now because it won't run as `build_library/release_util.sh` would have to be sourced before flag parsing and anyway we can't test it in this PR.

build_sysext
pothos:
```suggestion
eval set -- "${FLAGS_ARGV}"
```
That's what other scripts have

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,10 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2" 
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc" 
+  mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" -noappend
+
```
pothos, line 109: As test case, how about running the script here to create a python or docker sysext? Then we can also see it pass from the GitHub Actions CI workflow.

pothos: (Storing a docker sysext on the image is out of scope for this PR but I think it would be good to have the script used somewhere to see that it works and whether it breaks)

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,13 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2" 
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc" 
+  mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" -noappend
 
+  # Run build_sysext script to create python sysext so we can see it from GitHub actions CI workflow
+  sudo ../build_sysext python-sysext python
```
pothos, line 111:
```suggestion
  sudo "$(dirname ${BASH_SOURCE[0]})/../build_sysext" python-sysext python
```
I think that should resolve the path correctly

build_library/prod_image_util.sh
pothos: Yes, thanks

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,13 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2" 
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc" 
+  mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" -noappend
```
pothos, line 109:
```suggestion
  sudo rm -rf "${BUILD_DIR}/root_fs_dir2"
```
We also need to clean it up

build_sysext
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
pothos, line 1: Please run `chmod +x build_sysext` and push the changes

build_sysext
pothos: `prod` has no meaning for this script, or?

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+done
+
+# Unmount in order to get rid of the overlay
+umount "${BUILD_DIR}/install-root"
+umount "${BUILD_DIR}/fs-root"
+rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var" "${BUILD_DIR}/install-root/opt"
+mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+printf "ID=flatcar\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+mksquashfs "${BUILD_DIR}/install-root" "${BUILD_DIR}/${SYSEXTNAME}.raw" -all-root
+rm -rf "${BUILD_DIR}/{fs-root, install-root, workdir}"
```
pothos, line 107:
```suggestion
rm -rf "${BUILD_DIR}"/{fs-root,install-root,workdir}
```
Otherwise it doesn't work

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+. "${BUILD_LIBRARY_DIR}/test_image_content.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/torcx_manifest.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/vm_image_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/cros_vm_constants.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/oem_sysext_util.sh" || exit 1
+
+set -euo pipefail
+
+cleanup() {
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
```
pothos, line 61:
```suggestion
  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" 2>/dev/null || true
```
Suppress error output to have clearer logs

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+  ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="/build/${FLAGS_board}" FEATURES="-ebuild-locks" emerge --root-deps=rdeps --usepkgonly --verbose "${package}"
+done
+
+# Unmount in order to get rid of the overlay
+umount "${BUILD_DIR}/install-root"
+umount "${BUILD_DIR}/fs-root"
+rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var" "${BUILD_DIR}/install-root/opt"
+mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+printf "ID=flatcar\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+mksquashfs "${BUILD_DIR}/install-root" "${BUILD_DIR}/${SYSEXTNAME}.raw" -all-root
```
pothos, line 106:
```suggestion
mksquashfs "${BUILD_DIR}/install-root" "${BUILD_DIR}/${SYSEXTNAME}.raw"
```
We can use the original file permissions here.

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,13 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2" 
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc" 
+  mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" -noappend
+  sudo rm -rf "${BUILD_DIR}/root_fs_dir2"
+  # Run build_sysext script to create python sysext so we can see it from GitHub actions CI workflow
+  sudo "$(dirname ${BASH_SOURCE[0]})/../build_sysext" --board="${BOARD}" python-sysext python
```
pothos, line 111:
```suggestion
  sudo "$(dirname ${BASH_SOURCE[0]})/../build_sysext" --board="${BOARD}" --group="${FLAGS_group}" --output_root="${FLAGS_output_root}" --build_attempt="${FLAGS_build_attempt}" python-sysext python
```
I think we need to align these values

build_sysext
pothos: Please try to reduce this list to only those that get used

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,13 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2" 
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc"
```
pothos, line 107: These lines have trailing whitespace, can you check the others, too?

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+    echo "Error: No sysext name passed"
+    show_help_if_requested -h
+fi
+
+# Set up trap to execute cleanup() on script exit
+trap cleanup EXIT
+
+ARCH=$(_get_sysext_arch "${FLAGS_board}") 
+
+source <(grep -P '^(VERSION|VERSION_ID)=' /etc/os-release)
```
pothos, line 77: I think we rather have to use `FLATCAR_VERSION` and `FLATCAR_VERSION_ID` from `sdk_container/.repo/manifests/version.txt`

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+. "${BUILD_LIBRARY_DIR}/oem_sysext_util.sh" || exit 1
+
+set -euo pipefail
+
+cleanup() {
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" 2>/dev/null || true
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
+}
+
+export SYSEXTNAME="${1-}"
```
pothos, line 65:
```suggestion
SYSEXTNAME="${1-}"
```

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+
+cleanup
+
+mkdir "${BUILD_DIR}/fs-root"
+mount -rt squashfs -o loop,nodev "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" "${BUILD_DIR}/fs-root"
+mkdir "${BUILD_DIR}/install-root"
+mkdir "${BUILD_DIR}/workdir"
+mount -t overlay overlay -o lowerdir="${BUILD_DIR}/fs-root",upperdir="${BUILD_DIR}/install-root",workdir="${BUILD_DIR}/workdir" "${BUILD_DIR}/install-root"
+VERSION_BOARD=$(grep "^VERSION=" ${BUILD_DIR}/fs-root/usr/lib/os-release | cut -d = -f 2-)
+if [ "$VERSION_BOARD" != "$VERSION" ]; then
```
pothos, line 87:
```suggestion
if [ "$VERSION_BOARD" != "$FLATCAR_VERSION" ]; then
```

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+  echo "$VERSION_BOARD"
+  echo "$VERSION"
```
pothos, line 89:
```suggestion
  echo "$FLATCAR_VERSION"
```

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+  echo "Installing package into sysext image: $package"
+  ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="/build/${FLAGS_board}" FEATURES="-ebuild-locks" emerge --root-deps=rdeps --usepkgonly --verbose "${package}"
+done
+
+# Unmount in order to get rid of the overlay
+umount "${BUILD_DIR}/install-root"
+umount "${BUILD_DIR}/fs-root"
+rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var" "${BUILD_DIR}/install-root/opt"
+mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+printf "ID=flatcar\nVERSION_ID=${VERSION_ID}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
```
pothos, line 105:
```suggestion
printf "ID=flatcar\nVERSION_ID=${FLATCAR_VERSION_ID}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
```

build_sysext
```diff
@@ -0,0 +1,107 @@
...
+    echo "Error: No sysext name passed"
+    show_help_if_requested -h
+fi
+
+# Set up trap to execute cleanup() on script exit
+trap cleanup EXIT
+
+ARCH=$(_get_sysext_arch "${FLAGS_board}") 
+
+source <(grep -P '^(FLATCAR_VERSION|FLATCAR_VERSION_ID)=' /etc/os-release)
```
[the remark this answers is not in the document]
jepio: These variables (`FLATCAR_VERSION` and `FLATCAR_VERSION_ID` are set in `common.sh` so they could be used directly.

or let's use `${REPO_MANIFESTS_DIR}` instead of `${BASH_SOURCE[0]}`

t-lo:
The build looks good, tests are successful with a single failure that can be explained.
The devcontainer tests are failing for systemd-nspawn and docker on both archs; the tests break in `emerge_gitclone` with:
```
Cloning into _/var/lib/portage/scripts_..."
error: pathspec _285cff6c98_ did not match any file(s) known to git"
```
Pathspec cannot be found because the upstream `scripts` repo was cloned but this PR uses a fork.

pothos (changes requested):
Looks good except for unused imports: https://github.com/flatcar/scripts/pull/920/files#r1242488294

krnowak:
I'd like to make a review before merging this, please.

build_sysext
```diff
@@ -0,0 +1,97 @@
...
+  
+for package in "$@"; do
+  echo "Installing package into sysext image: $package"
+  ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="/build/${FLAGS_board}" FEATURES="-ebuild-locks" emerge --root-deps=rdeps --usepkgonly --verbose "${package}"
+done
+
+# Unmount in order to get rid of the overlay
+umount "${BUILD_DIR}/install-root"
+umount "${BUILD_DIR}/fs-root"
+rm -rf "${BUILD_DIR}/install-root/dev" "${BUILD_DIR}/install-root/etc" "${BUILD_DIR}/install-root/var" "${BUILD_DIR}/install-root/opt"
```
krnowak, line 93: Two things here:

- Before these files/directories are removed, I'd like to be able to run a script that may do some more cleanups or modifications to the rootfs before ossifying it inside the squashfs image. It would be something like `build_sysext --manglefs_script=path/to/manglefs.sh my-image dev-lang/python`. So, after doing `umount "${BUILD_DIR}/fs-root"`, we could do something like:

```bash
# somewhere at the top:
…

# before removing the extra directories from the sysext rootfs
if [[ -n "${FLAGS_manglefs_script}" ]]; then
  if [[ ! -x "${FLAGS_manglefs_script}" ]]; then
    die "${FLAGS_manglefs_script} is not executable"
  fi
  "${FLAGS_manglefs_script}" "${BUILD_DIR}/install-root"
fi
```

- I think we want to be quite loud about the stuff that we remove, so there's no surprise about missing stuff in the sysext image (borrowed from `oem_sysext_util.sh`):

```suggestion
info "Removing non-/usr directories from sysext image"
for entry in "${BUILD_DIR}/install-root"/*; do
  if [[ "${entry}" = */usr ]]; then
    continue
  fi
  info "  Removing ${entry##*/}"
  rm -rf "${entry}"
done
```

build_sysext
```diff
@@ -0,0 +1,97 @@
...
+VERSION_BOARD=$(grep "^VERSION=" ${BUILD_DIR}/fs-root/usr/lib/os-release | cut -d = -f 2-)
+if [ "$VERSION_BOARD" != "$FLATCAR_VERSION" ]; then
+  echo "$VERSION_BOARD"
+  echo "$FLATCAR_VERSION"
+  echo "Version mismatch between board flatcar release and SDK container flatcar release"
+  exit 1
+fi
+shift
+  
+for package in "$@"; do
```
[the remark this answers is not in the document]
krnowak: I'm also wondering how to make this script to satisfy the needs of the OEM sysext image building. The way OEM sysext image are built is that we tell emerge to use just one package (like `coreos-base/oem-azure`). This package is a metapackage of sorts - has no contents, just pulls in some more deps. The catch is that these metapackages are not built during `build_packages`, so there are no available binary packages for them. So we were doing a two step build of those: first we just built a binary package of the metapackage (ignoring its deps, not installing it) and then install the metapackage and its deps using only binary packages.

I suppose we could have a flag like `metapkgs` and we could do something like this:

```bash
# somewhere at the top:
DEFINE_string metapkgs '' \
  "Comma-separated list of meta-packages to build and install into sysext image"
…

# Before calling emerge on list of packages:
if [[ -n "${FLAGS_metapkgs}" ]]; then
  mapfile -t metapkgs < <(tr ',' '\n' <<<"${FLAGS_metapkgs}")
  "emerge-${FLAGS_board}" --nodeps --buildpkgonly --usepkg n --verbose "${metapkgs[@]}"
  set -- "${metapkgs[@]}" "${@}"
fi
```

See the suggestion below:

```suggestion
if [[ -n "${FLAGS_metapkgs}" ]]; then
  mapfile -t metapkgs < <(tr ',' '\n' <<<"${FLAGS_metapkgs}")
  "emerge-${FLAGS_board}" --nodeps --buildpkgonly --usepkg n --verbose "${metapkgs[@]}"
  set -- "${metapkgs[@]}" "${@}"
fi

for package; do
```

build_sysext
```diff
@@ -0,0 +1,97 @@
...
+
+# Parse command line
+FLAGS "$@" || exit 1
+
+eval set -- "${FLAGS_ARGV}"
+
+. "${BUILD_LIBRARY_DIR}/release_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/build_image_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/oem_sysext_util.sh" || exit 1
```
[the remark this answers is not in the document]
krnowak: Yes, just drop `oem_sysext_util.sh` from includes, the rest stays as is.

build_sysext
```diff
@@ -0,0 +1,97 @@
...
+
+# Set up trap to execute cleanup() on script exit
+trap cleanup EXIT
+
+ARCH=$(_get_sysext_arch "${FLAGS_board}") 
+
+cleanup
+
+mkdir "${BUILD_DIR}/fs-root"
+mount -rt squashfs -o loop,nodev "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" "${BUILD_DIR}/fs-root"
```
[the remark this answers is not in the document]
krnowak: We may have it elsewhere during CI build. The `flatcar_production_image_sysext.squashfs` is built during the `image` step, but we will be using this script in the next step, `vms`. Which means that `image` step will upload the squashfs image to the bincache, and `vms` step will download it. So in `vms` step the location of this file may be different.

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,13 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
+  sudo cp -r "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2"
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc"
+  sudo mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/flatcar_production_image_sysext.squashfs" -noappend
+  sudo rm -rf "${BUILD_DIR}/root_fs_dir2"
+  # Run build_sysext script to create python sysext so we can see it from GitHub actions CI workflow
+  sudo "$(dirname ${BASH_SOURCE[0]})/../build_sysext" --board="${BOARD}" --group="${FLAGS_group}" --output_root="${FLAGS_output_root}" --build_attempt="${FLAGS_build_attempt}" python-sysext python
```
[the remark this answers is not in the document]
krnowak: Cool, thanks.

krnowak:
> A problem with installing regular packages is that in many cases the software won't correctly work because only `/usr` is added. Files on `/etc` will be missing and the approach in the Azure OEM was that we ship the needed files in the sysext and then set up symlinks before, e.g., a service starts (We could build a confext in the future but this doesn't work now). Also, files in `/var` will be missing. For some of these files there might be tmpfile directives but they won't get executed on load. Finally, the services that use `.wants` symlinks to be enabled would currently work because we have deployed a workaround but we plan to drop this because the recommended way is to use `Upholds=` instead of `Wants=` (currently through drop-ins for the multi-user target but in systemd 254 it should work to use `.upholds` symlinks). In summary, we need some postprocessing for some scenarios and call out what won't work for the rest.
> 
> Edit: A great use case for this would be to build a Docker and containerd sysext to replace torcx.

I think this should be addressed to some degree by the `--manglefs_script` option.

krnowak:
I'm currently trying to use this script to build OEM sysexts, just to see if we can do it. I think I'll want more changes, mostly related to the BUILD_DIR stuff. You can see my branch here: https://github.com/flatcar/scripts/commits/krnowak/test-new-oem-script (ignore the fact that the commits are made by Flatcar Buildbot, apparently I messed up my git config :) ). The changes in the script that I have made are here: https://github.com/flatcar/scripts/commit/72d2b62855dc434b2e118fd65a363c63eee18d35

krnowak:
> Ok cool. Just these changes?

I think I'll let you know on Monday - the build is currently running on our Jenkins instance.

build_library/prod_image_util.sh
```diff
@@ -102,7 +103,13 @@ create_prod_image() {
   fi
 
   tar -cf "${BUILD_DIR}/${image_pkgdb}" -C "${root_fs_dir}" var/cache/edb var/db/pkg
-
+  sudo cp -a "${root_fs_dir}" "${BUILD_DIR}/root_fs_dir2"
+  sudo rsync -a --delete  "${BUILD_DIR}/configroot/etc/portage" "${BUILD_DIR}/root_fs_dir2/etc"
+  sudo mksquashfs "${BUILD_DIR}/root_fs_dir2"  "${BUILD_DIR}/${image_sysext_base}" -noappend
+  sudo rm -rf "${BUILD_DIR}/root_fs_dir2"
+  # Run build_sysext script to create python sysext so we can see it from GitHub actions CI workflow; TO REMOVE
+  sudo "$(dirname ${BASH_SOURCE[0]})/../build_sysext" --board="${BOARD}" --group="${FLAGS_group}" --output_root="${FLAGS_output_root}" --build_attempt="${FLAGS_build_attempt}" --squashfs_base="${BUILD_DIR}/${image_sysext_base}" python-sysext python
```
pothos, line 111: Before merge we should delete this and also squash the commits and add a longer commit message (the body should have two paragraphs, one describing the problem/goal, one describing what the changes do)

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+# All these are used to set up the 'BUILD_DIR' variable
+DEFINE_string board "${DEFAULT_BOARD}" \
+  "The board to build a sysext for."
+DEFINE_string output_root "${DEFAULT_BUILD_ROOT}/images" \
+  "Directory in which to place sysext images and intermediate files"
+DEFINE_string group "developer" \
+  "The normal image build group"
+DEFINE_integer build_attempt 1 \
+  "The build attempt for this build."
+
```
krnowak, line 26: These flags are needed only because this script includes `build_image_util`, which doesn't make much sense here - we are not using it for anything else but the `BUILD_DIR` variable. I'd say that this is something that we could get directly from a flag like `--build_dir`.

```suggestion
DEFINE_string build_dir '' \
  "Directory used for building the sysext image, must exist, but should be empty."
```

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+DEFINE_string metapkgs '' \
+  "Comma-separated list of meta-packages to build and install into sysext image"
```
krnowak, line 28: Let's clarify that it's about building from source:
```suggestion
  "Comma-separated list of meta-packages to build from source and install into sysext image."
```

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+DEFINE_string manglefs_script '' \
+  "A path to executable that will customize the rootfs of the sysext image."
+DEFINE_string squashfs_base '' \
+  "The path to the squashfs base image."
+
+FLAGS_HELP="USAGE: build_sysext [flags] [sysext name] [packages to build].
+This script is used to build a Flatcar sysext.
```
krnowak, line 35: Let's make it clear it's about using prebuilt binary packages here.

```suggestion
FLAGS_HELP="USAGE: build_sysext [flags] [sysext name] [binary packages to install into image].

This script is used to build a Flatcar sysext image. 
```

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+FLAGS_HELP="USAGE: build_sysext [flags] [sysext name] [packages to build].
+This script is used to build a Flatcar sysext. 
+
+Examples:
+
+Build a systemd-sysext image named command-extras with package_1, package_2 and package_3 for amd64:
+sudo build_sysext --board=amd64-usr command-extras package_1 package_2 package_3
+
+Builds a systemd-sysext image named command-extras with package_1, package_2 and package_3 for arm64:
+sudo build_sysext --board=arm64-usr command-extras package_1 package_2 package_3
```
krnowak, line 43: How about these examples?
```suggestion
Builds a sysext image named interpreters in the images
directory with dev-lang/python and dev-lang/perl packages for amd64:

sudo build_sysext \
     --board=amd64-usr \
     --build_dir=images \
     interpreters dev-lang/python dev-lang/perl

Builds a sysext image named oem-azure in the oem-images directory with
metapackage coreos-base/oem-azure for arm64:

sudo build_sysext \
     --board=arm64-usr \
     --build_dir=oem-images \
     --metapkgs=coreos-base/oem-azure \
     --mangle_fs=…/coreos-base/oem-azure/files/manglefs.sh \
     oem-azure
```

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+show_help_if_requested "$@"
+
+# Parse command line
+FLAGS "$@" || exit 1
+
+eval set -- "${FLAGS_ARGV}"
+
+. "${BUILD_LIBRARY_DIR}/release_util.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
+. "${BUILD_LIBRARY_DIR}/build_image_util.sh" || exit 1
```
krnowak, line 56: With the build dir coming directly from flags, including build_image_util is not necessary, so could be removed. Also, I'm not sure why are we including release_utils.sh, so it can be dropped too. And we can define the BUILD_DIR ourselves now:

```suggestion
. "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1

if [[ -z "${FLAGS_build_dir}" ]]; then
  die "Need a build directory to be specified with a --build_dir option"
fi

BUILD_DIR=$(realpath "${FLAGS_build_dir}")
```

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+}
+
+
+set -euo pipefail
+
+cleanup() {
+  umount "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" 2>/dev/null || true
+  rm -rf "${BUILD_DIR}/fs-root" "${BUILD_DIR}/install-root" "${BUILD_DIR}/workdir" || true
+}
+
```
krnowak, line 82: I'd add a check for a required parameter for the image name here, so we won't get errors like `$1: unbound variable` on wrong invocations:
```suggestion

if [[ ${#} -lt 1 ]]; then
  die 'Expected at least one parameter for sysext image name'
fi

```

build_sysext
```diff
@@ -0,0 +1,153 @@
...
+for entry in "${BUILD_DIR}/install-root"/*; do
+  if [[ "${entry}" = */usr ]]; then
+    continue
+  fi
+  info "  Removing ${entry##*/}"
+  rm -rf "${entry}"
+done
+mkdir -p "${BUILD_DIR}/install-root/usr/lib/extension-release.d"
+version_field="${VERSION_FIELD_OVERRIDE:-VERSION_ID=${FLATCAR_VERSION_ID}}"
+printf "ID=flatcar\n${version_field}\nARCHITECTURE=${ARCH}\n" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
```
krnowak, line 151: Since this printf is printing stuff coming from the caller (possibly the VERSION_FIELD_OVERRIDE or stuff in FLATCAR_VERSION_ID), which may be unsafe to be put inside the format string (the first parameter of printf command) because there may be `%` signs lurking, I'd rewrite it as follows:

```suggestion
all_fields=(
  'ID=flatcar'
  "${version_field}"
  "ARCHITECTURE=${ARCH}"
)
printf '%s\n' "${all_fields[@]}" >"${BUILD_DIR}/install-root/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
```

krnowak (changes requested):

krnowak (approved):

build_sysext
```diff
@@ -0,0 +1,174 @@
...
+if [[ ${#} -lt 1 ]]; then
+  die 'Expected at least one parameter for sysext image name'
+fi
+
+SYSEXTNAME="${1-}"
+shift
+if [ -z "$SYSEXTNAME" ]; then
+    echo "Error: No sysext name passed"
+    show_help_if_requested -h
+fi
```
pothos, line 104: This and the below doesn't really fit now anymore. I would move the `show_help_if_requested` up into the previous check and then remove this check here.

pothos: Thanks, looks good!

pothos (approved):

PR #933, opened 2023-06-16
[sys-auth/polkit] Harden systemd service configuration

krishjainx:
# [sys-auth/polkit] Harden systemd service configuration

This pull request introduces hardened systemd service configuration for the polkit service, resulting in a significant reduction in the exposure level from 9.6 to 1.9, as determined by the [``systemd-analyze security``](https://www.freedesktop.org/software/systemd/man/systemd-analyze.html#systemd-analyze%20security%20%5BUNIT...%5D) command.

## Changes Made

- Added hardening measures utilizing systemd features to enhance the security of the polkit service.

## How to Use

I have already modified the ebuild to include the necessary changes for installing the additional hardening measures. Simply follow the regular installation process to benefit from the enhanced security.

## Testing Done

- Analyzed the polkit service using systemd-analyze to ensure the effectiveness of the hardening measures.
- Conducted testing to verify that the polkit service functions as expected after implementing the changes.
- Examined logs

## Checklist

- [ ] Added appropriate changelog entries in the respective `changelog/` directory to reflect the changes made (user-facing change, bug fix, security fix, update).

Please review the changes and provide any feedback or suggestions for improvement.

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/sys-auth/polkit/files/override.conf
+++ b/sdk_container/src/third_party/coreos-overlay/sys-auth/polkit/files/override.conf
@@ -1,11 +1,4 @@
-[Unit]
-Description=Authorization Manager
-Documentation=man:polkit(8)
-
 [Service]
-Type=dbus
-BusName=org.freedesktop.PolicyKit1
-ExecStart=/usr/lib/polkit-1/polkitd --no-debug
 
 # Network Sandboxing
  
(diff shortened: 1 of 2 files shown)


jepio:
Then it's a good idea to align with upstream, by either updating polkit to a version that has the updated file or if that is not available yet - updating the file ourselves.

pothos:
Ah, so it's a mix of backport plus changes, then great, less things to submit upstream. They will know best what settings are valid and what aren't.
That said, I don't see much benefit of any hardening because once you take over the polkit daemon you control the permission granting and can grant malicious actions outside of the service sandbox.

thesamesam:
By the way, we're interested in this sort of thing upstream too (although ideally send it to "real" upstream first): https://archives.gentoo.org/gentoo-dev/message/42e5ea98d30c7c10c103cca0b6a2bafb.

thesamesam:
@krishjainx Hi! Yeah, that sounds good to me. Just make sure you include a link to any relevant upstream MRs/bugs in the patch.

PR #948, opened 2023-06-26
coreos-base/coreos-init: Work around bash regression and handle dd error

pothos:
This pulls in
https://github.com/flatcar/init/pull/97
and
https://github.com/flatcar/init/pull/99
to work around a bash regression and add handling for disk write errors.

## How to use

Backport to Alpha and Beta, maybe also Stable?
Needs to be done through `init` backport branches for Beta/Stable.

## Testing done

So far only manual testing. [Jenkins test](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2048/cldsv/) started with Equinix Metal.

Closes: https://github.com/flatcar/Flatcar/issues/1059


pothos:
Backported to Alpha and Beta (using `flatcar-3602-backport` as `init` branch).

PR #951, opened 2023-06-28
[acct-group/docker] Add acct-group/docker to portage-stable

krishjainx:
# [acct-group/docker] Add acct-group/docker to portage-stable

Add acct-group/docker to portage-stable. Looks like an omission from our side when syncing the docker ebuilds with Gentoo.

## How to use

Dependency to be able to successfully ``emerge app-containers/docker``

## Testing done

With this patch (adding acct-group/docker to portage-stable) I am able to emerge app-containers/docker


dongsupark (approved):
Looks good.
The CI failures are all known issues.

PR #953, opened 2023-06-29
Upgrade Linux Kernel for flatcar-3510 from 5.15.117 to 5.15.119

flatcar-infra:
Subject says it all.


dongsupark:
We need to first release the next Stable with Kernel 5.15.117, https://github.com/flatcar/scripts/pull/928, not this one.
Added `on-hold`.

dongsupark:
Build will fail with Kernel 5.15.119, because https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=34e21b8ff3e6445baedb813b2ba8d306d6f312c8 changed the Hyper-V code. As a result, it fails to apply the z0006 patch.
Will push a fix.

dongsupark (approved):
Now that Stable 3510.2.4 with Kernel 5.15.117 was released, we can merge this PR.

PR #954, opened 2023-06-29
Upgrade Linux Kernel for flatcar-3602 from 5.15.117 to 5.15.119

flatcar-infra:
Subject says it all.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2091/cldsv/


dongsupark:
Build fails due to an issue when applying the z0006 patch.
The next beta will be anyway not 3602, but 3637.1.0. So we do not have to fix this PR.
Added `on-hold`.

However, still we need to fix Stable flatcar-3510.

dongsupark:
It looks like we still need to fix Beta 3602.

krnowak (approved):

PR #959, opened 2023-06-30
[wip] automate the capi image building and publishing

sayanchowdhury:
This is a WIP PR, not yet tested. 

ref https://github.com/flatcar/Flatcar/issues/1104


ci-automation/azure-sig.sh
```diff
@@ -0,0 +1,368 @@
...
+
+[[ -n ${DEBUG:-} ]] && set -o xtrace
+
+# Make sure Azure credentials are given without printing them.
+tracestate="$(shopt -po xtrace || true)"
+set +o xtrace
+_="
+${AZURE_CLIENT_ID}
+${AZURE_CLIENT_SECRET}
+"
```
pothos, line 13: If we don't require them as env vars, we can remove this

pothos: If you want parallel usage through manual invocation, you can move this to the `else` case of the block below.

ci-automation/capi_image.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+#
+# Copyright (c) 2023 The Flatcar Maintainers.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+# >>> This file is supposed to be SOURCED from the repository ROOT. <<<
```
pothos, line 7: If we don't really need sourcing, we can simplify this and expect the script to be invoked directly. Sourcing always modifies shared global state and is hard to reason about.

ci-automation/azure-sig.sh
```diff
@@ -0,0 +1,368 @@
...
+
+  IMAGE_NAME="${FLATCAR_CAPI_IMAGE_NAME}"
+  IMAGE_VERSION="${FLATCAR_VERSION}"
+  GALLERY_NAME="${FLATCAR_CAPI_GALLERY_NAME}"
+  RESOURCE_GROUP_NAME="${PUBLISHING_SIG_RESOURCE_GROUP}"
+
+  # shellcheck disable=SC2310 # This might return 1.
+  if sig-image-version-exists; then
+    return
+  fi
```
pothos, line 82: This is not the wanted behavior when we have to rerun the release step. It should replace/republish.

ci-automation/azure-sig.sh
```diff
@@ -0,0 +1,368 @@
...
+
+  IMAGE_NAME="${FLATCAR_CAPI_IMAGE_NAME}"
+  IMAGE_VERSION="${FLATCAR_VERSION}"
+  GALLERY_NAME="${FLATCAR_CAPI_STAGING_GALLERY_NAME}"
+  RESOURCE_GROUP_NAME="${STAGING_SIG_RESOURCE_GROUP}"
+
+  # shellcheck disable=SC2310 # This might return 1.
+  if sig-image-version-exists; then
+    return
+  fi
```
pothos, line 114: This is not the wanted behavior when we have to rerun the release step. It should replace/republish.

ci-automation/azure-sig.sh
```diff
@@ -0,0 +1,368 @@
...
+
+function ensure-resource-group() {
+  RESOURCE_GROUP_NAME=${RESOURCE_GROUP_NAME:-}
+
+  if ! az group show -n "${RESOURCE_GROUP_NAME}" -o none 2>/dev/null; then
+    az group create -n "${RESOURCE_GROUP_NAME}" -l "${AZURE_LOCATION}"
+  fi
+}
+
+function login() {
```
pothos, line 346: `login` conflicts with `/usr/bin/login`

ci-automation/capi_image.sh
```diff
@@ -0,0 +1,77 @@
...
+    FLATCAR_CHANNEL="$(get_git_channel)"
+    FLATCAR_AZURE_AUTH_CREDENTIALS="${AZURE_AUTH_CREDENTIALS}"
+
+    # Provide a python3 command for the k8s schedule parsing
+    export PATH="$PATH:$PWD/ci-automation/python-bin"
+    k8s_release_versions=$(ci-automation/get_kubernetes_releases.py)
+}
+
+function _capi_image_build_impl() {
+    local arch="$1"
```
pothos, line 51: For the high level entry point maybe let's not require the architecture to be passed from the jenkins job where it would have to have a for loop. Either this script here can build and publish arm64 or it can't because the feature is missing.

ci-automation/get_kubernetes_releases.py
```diff
@@ -0,0 +1,18 @@
...
+http = urllib3.PoolManager()
+r = http.request('GET', url, preload_content=False)
+data = r.data
+r.release_conn()
+
+parsed_data = yaml.safe_load(data)
+
+
+for elem in parsed_data.get('schedules', []):
+    print(elem.get('next').get('release'))
```
pothos, line 18: This should filter out releases with a target in the future. The yaml entry looks like that: `targetDate: 2023-07-12`

PR #979, opened 2023-07-06
Check the url in get_sdk_binhost before echoing

krishjainx:
# Check the url in get_sdk_binhost before echoing

Hi. When running the build_packages script, one can encounter an error such as 'Error fetching binhost package info from.' This pertains to SDK packages (not board packages). Since we have transitioned to the SDK container, the SDK packages are no longer published independently from the container image.

@jepio encouraged me to submit a pull request (PR) for the same

### DIFF
--- a/build_library/toolchain_util.sh
+++ b/build_library/toolchain_util.sh
@@ -201,8 +201,10 @@ get_sdk_binhost() {
         # but normally it is not needed because everything is already part of the tarball.
         # To install the crossdev Rust package, /toolchain-arm64/ is derived from /toolchain/
         # when necessary in install_cross_toolchain().
-        echo "${FLATCAR_DEV_BUILDS_SDK}/${arch}/${ver}/toolchain/"
-        echo "${FLATCAR_DEV_BUILDS_SDK}/${arch}/${ver}/pkgs/"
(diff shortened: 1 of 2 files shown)


jepio:
Can you squash the commits?

jepio:
I've tried this PR and now the following warning is printed onto the console:
```
!!! /etc/portage/binrepos.conf is missing (or PORTAGE_BINHOST is unset), but use is requested.
!!! /etc/portage/binrepos.conf is missing (or PORTAGE_BINHOST is unset), but use is requested.
```
it only makes sense to merge this if it gets rid of this warning as well.

jepio:
> @jepio Hello, Jeremi. It appears that this error has also been resolved through my recent change. Please review it so that we can proceed with the merge. Thank you!

This is not a fix that we can merge. We *do* rely on binpkgs (and these flags) during the sdk build, but not once the sdk is pushed to the container registry. So we can't take this shortcut and remove all these flags.

You could deal with binhost being empty (or only whitespace) and then perform the necessary actions (not sure what exactly is needed to silence the warning): skip passing `--getbinpkg`.

Sidenote: we don't merge main into PR branches, we rebase them.

jepio (approved):

jepio:
There seem to be more similar messages but this is good enough to merge now.

PR #983, opened 2023-07-07
build_sysext: bugfixes, QoL changes, help updated

t-lo:
This change improves `build_sysext` by sourcing a missing lib dependency, adding a number of comfort / quality-of-life options, and updating the output of `--help` accordingly.

The OEM sysext finction in `build_library/vm_image_util.sh` is also updated to use new command line format.

1. Include missing dependency `toolchain_util.sh` to fix an error in `board_options.sh` when `--board` was not specified (`get_board_arch` undefined).
2. Use positional parameters for mandatory arguments:  `sysext_name` and `binary_package` are positional arguments now; the latter can be omitted if `--metapkgs` was specified.
3. `--squashfs_base` is now guessed better and will use the most recent build by default.
4. `--image_builddir` is now guessed and defaults to a `sysext` sub-directory in the directory the squashfs base resides in.
5. A new boolean flag `--ignore_version_mismatch` for the more daring developer was added. The flag will cause the script to continue if a version mismatch between SDK board packages and squashfs base is detected.
6. Error messages were improved for when mandatory parameters were not provided.
7. The `--help` message was improved and adjusted to the new parameters.

Tested with a custom sysext build and with a `image_to_vm.sh` qemu build (for testing the generic OEM sysext build).

### DIFF
--- a/build_sysext
+++ b/build_sysext
@@ -14,7 +14,7 @@ SCRIPT_ROOT=$(dirname "$(readlink -f "$0")")
 assert_inside_chroot
 assert_root_user
 
-default_squashfs_template="$(readlink -f "${SCRIPT_ROOT}/../build/images")/<BOARD>/latest/${FLATCAR_PRODUCTION_IMAGE_SYSEXT_BASE}"
+default_imagedir="$(readlink -f "${SCRIPT_ROOT}/../build/images")/<BOARD>/latest/"
 
 # All these are used to set up the 'BUILD_DIR' variable
 DEFINE_string board "${DEFAULT_BOARD}" \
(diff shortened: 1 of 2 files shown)


build_sysext
```diff
@@ -14,42 +14,55 @@ SCRIPT_ROOT=$(dirname "$(readlink -f "$0")")
 assert_inside_chroot
 assert_root_user
 
+default_squashfs_base="$(readlink -f "${SCRIPT_ROOT}/../build/images/amd64-usr/latest/${FLATCAR_PRODUCTION_IMAGE_SYSEXT_BASE}")"
```
jepio, line 17: hmmm, hardcoding board?

jepio:
can you also add `-noappend` to the `mksquashfs` invocation?

build_sysext
t-lo: That's what I get from copy-pasting my command lines, meh. Will fix.

t-lo:
Fixed hard-coded architecture, added `-noappend` option to `mksquashfs`.

build_sysext
```diff
@@ -14,42 +14,55 @@ SCRIPT_ROOT=$(dirname "$(readlink -f "$0")")
...
 DEFINE_string metapkgs '' \
   "Comma-separated list of meta-packages to build from source and install into sysext image."
+DEFINE_string squashfs_base '' \
+  "The path to the squashfs base image. Defaults to the most current image built in '${default_squashfs_template}'."
 DEFINE_string manglefs_script '' \
   "A path to executable that will customize the rootfs of the sysext image."
-DEFINE_string squashfs_base '' \
-  "The path to the squashfs base image."
+DEFINE_boolean ignore_version_mismatch "${FLAGS_FALSE}" \
+  "Ignore version mismatch between SDK board packages and base squashfs. DANGEROUS."
```
krnowak, line 29: `<insert gif with rick dangerous>`

krnowak (approved):
Cool. Nice improvements, thanks!

The situation with meta packages being passed in a different way than binary packages (a flag with comma-separated list as its value vs. a list of positional parameters) is a bit iffy, but I don't really have anything better to suggest right now.

build_sysext
t-lo: > The situation with meta packages being passed in a different way than binary packages (a flag with comma-separated list as its value vs. a list of positional parameters) is a bit iffy, but I don't really have anything better to suggest right now.

I briefly considered replacing `--metapkgs` with a `--build-if-needed` option that builds the packages if necessary but that's just a different way of solving the issue. I think your approach is more flexible.

t-lo:
@krnowak @jepio One more round please:  `--image_builddir` is now guessed and defaults to a `sysext` sub-directory in the directory the squashfs base resides in. Everything else has been addressed.

This allows a user, after a build, to plainly run
```bash
./build_sysext python dev-lang/python
```
to build a python sysext for the OS image that was just built.

build_sysext
```diff
@@ -59,14 +73,30 @@ FLAGS "$@" || exit 1
...
+  FLAGS_squashfs_base="$(readlink -f "${SCRIPT_ROOT}/../build/images/${FLAGS_board}/latest/${FLATCAR_PRODUCTION_IMAGE_SYSEXT_BASE}")"
+fi
+if [[ ! -f "${FLAGS_squashfs_base}" ]] ; then
+  die "Squashfs base '${FLAGS_squashfs_base}' not found."
+fi
+
+if [[ -z "${FLAGS_image_builddir}" ]]; then
+  FLAGS_image_builddir="$(dirname "${FLAGS_squashfs_base}")/sysext"
+fi
+BUILD_DIR=$(realpath "${FLAGS_image_builddir}")
```
jepio, line 94: Originally I thought we should be taking `BUILD_DIR` from `build_library/build_image_util.sh` but Krish ran into some problems.

krnowak: We dropped using `build_image_util.sh` at my request to replace that with just a build directory parameter, because OEM building is not using any of the output root, build attempts and other stuff that `build_image_util.sh` forces you to do. If someone wishes to use `build_image_util.sh`, he can do so by importing it by himself, and passing the `BUILD_DIR` to `build_sysext` through the `--build_dir` (or rather: `--image_builddir` now).

PR #984, opened 2023-07-07
profiles/coreos: Account for python versions with minor > 10

jepio:
# don't truncate python minor in profile

`sys.version` returns `3.10.12 (main, Jul  6 2023, 22:54:08) [GCC 12.2.1 20230428]`.
The previous expression (sys.version[:3])) truncates 3.10 to 3.1 leading to build failures in some ebuilds that depend on python (like zfs).

## How to use

Don't know which ebuilds (other than sys-fs/zfs) use this but the problem is obvious.

## Testing done

Built zfs sysext.


krnowak (approved):

krnowak:
Next update, when minor is >100 or when major is >10.

PR #997, opened 2023-07-20
Upgrade Linux Kernel for main from 6.1.38 to 6.1.39

flatcar-infra:
Subject says it all.


dongsupark:
See instead https://github.com/flatcar/scripts/pull/1001

PR #1016, opened 2023-07-26
Support OEM systemd-sysext images and Flatcar extensions

pothos:
The vendor tools on the OEM partition weren't updated. We now want to
    ship them as systemd-sysext images which we can easily update. This
    change extends the Flatcar A/B update mechanism to cover the OEM
    systemd-sysext images. The same mechanism is also able to support
    "official" Flatcar extensions, e.g., a ZFS extension.

Additional change:
- GitHub Actions: Use correct version in kola run and skip update test
The kola run didn't pick up the version that was set up in the build
    because the git changes from that step are lost.
    Redo the version setup in the kola run to use the same version, and
    skip the kola update test if no update payload can be found. In the
    future we should copy it over from the GitHub Action artifact.

## How to use

After merging we always need to publish signed OEM payloads during release.

Used PRs:
https://github.com/flatcar/update_engine/pull/24
https://github.com/flatcar/mantle/pull/448
https://github.com/flatcar/bootengine/pull/68
https://github.com/flatcar/init/pull/101

TODO: When merging, create a follow-up GitHub issue for proper update tests on GitHub Actions PR builds (only for the regular update test, that should be easily doable).

## Testing done

See update-engine PR, Jenkins run is [here](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2213/cldsv/)


sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/files/oems/azure
```diff
@@ -0,0 +1,13 @@
+/etc/systemd/system/nvidia.service
```
krnowak, line 1: I assume that the deletion happens after /etc is already overlaid? Also hopefully this won't delete third party files that accidentally had the same name.

krnowak (approved):

sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/files/oems/azure
pothos: Yes, after the overlay. For user files with the same service name that we bring in I think the most practical solution is to document this because the same service name used will have other issues, too, I will add it to the changelog.

PR #1055, opened 2023-08-10
Upgrade Linux Firmware in main from 20230625 to 20230804

flatcar-infra:
Subject says it all.


dongsupark (approved):
Cleared MY_COMMIT as it is now an official release.
CI passed.

PR #1115, opened 2023-09-06
sys-kernel/coreos-modules: Enable dm-verity signature support

jepio:
# Enable dm-verity signature support

This allows checking against the primary kernel keyring, but not the secondary. It is a prerequisite for the ioctl to inject the signature to be supported in the first place.

## How to use

Test with sysext with signature: https://github.com/flatcar/sysext-bakery/pull/24.

## Testing done

Not yet. Will see if ioctl still reports error after an image for this PR is built.


dongsupark:
Moved to backlog. Feel free to update.

PR #1132, opened 2023-09-12
Allow use qcow2 inline compression

krishjainx:
#  Provide the .img qcow2 without wrapped compression by enabling the inline compression

[#1135](https://github.com/flatcar/Flatcar/issues/1135) by @tuxtof

Spoke with @pothos

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -619,7 +619,7 @@ _write_raw_disk() {
 }
 
 _write_qcow2_disk() {
-    qemu-img convert -f raw "$1" -O qcow2 -o compat=0.10 "$2"
+    qemu-img convert -f raw "$1" -O qcow2 -c -o compat=0.10 "$2"
     assert_image_size "$2" qcow2
 }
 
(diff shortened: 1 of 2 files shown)


pothos:
Started a github PR build, we should also test on Jenkins

sayanchowdhury:
Jenkins CI http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2493/cldsv/

sayanchowdhury:
There is one failure: 
```
 --- FAIL: systemd.journal.user (16.95s)
         cluster.go:130: "journalctl --user --unit hello.service | grep Foo" failed: output , status Process exited with status 1
```

pothos:
> There is one failure:
> 
> ```
>  --- FAIL: systemd.journal.user (16.95s)
>          cluster.go:130: "journalctl --user --unit hello.service | grep Foo" failed: output , status Process exited with status 1
> ```

That looks unrelated

pothos:
The jenkins build got cleaned up, can't see the results but I think it's ok to merge if the reported failure was only about an unrelated kola test.
Should we backport this to all channels?

dongsupark:
Let's merge.
I am fine with backports.

pothos:
Let's see how the nightly build turns out before backporting.

pothos:
Backported

PR #1141, opened 2023-09-14
app-misc/mime-types: sync to upstream

t-lo:
Gentoo Upstream commit c2907c4650dc08334953577355e3e7493b62fc5a switches to a new upstream mime-types (Fedora maintained). Upstream's new upstream is mode active and more up to date.

Also, the mime-types-9 source tarball disappeared from most mirrors, making OS builds less reliable.


jepio:
how about we add it to `.github/workflows/portage-stable-package-list` for auto syncing?

krnowak (approved):
Looks good, thanks!

PR #1148, opened 2023-09-18
Add ue-rs to image

jepio:
# Add ue-rs to image

<to be filled by @wrl>

## How to use

## Testing done


sdk_container/src/third_party/coreos-overlay/coreos-base/ue-rs/ue-rs-0.1.0-r1.ebuild
```diff
@@ -0,0 +1 @@
+ue-rs-9999.ebuild
```
krnowak, line 1: I'm not sure if having `ue-rs-9999` makes any sense. Could we have `ue-rs-0.1.0-r1.ebuild` as a regular file and drop the `9999` one?

sdk_container/src/third_party/coreos-overlay/coreos-base/ue-rs/ue-rs-9999.ebuild
```diff
@@ -0,0 +1,228 @@
...
+	windows_x86_64_msvc-0.48.5
+	winreg-0.50.0
+	xmlparser-0.13.5
+	zeroize-1.6.0
+"
+
+inherit coreos-cargo cros-workon systemd
+
+DESCRIPTION="Prototype Omaha Rust implementation"
+HOMEPAGE="https://github.com/coreos/ue-rs"
```
krnowak, line 204:
```suggestion
HOMEPAGE="https://github.com/flatcar/ue-rs"
```

sdk_container/src/third_party/coreos-overlay/coreos-base/ue-rs/ue-rs-0.1.0-r1.ebuild
jepio: i would keep this as is

t-lo (approved):
The ebuild looks sensible and tests run fine. LGTM.

sdk_container/src/third_party/coreos-overlay/coreos-base/ue-rs/ue-rs-0.1.0-r1.ebuild
krnowak: `git blame` should be smart enough to follow renames. But I'm not going to block the PR on this.

krnowak (approved):

PR #1149, opened 2023-09-18
core_sign_update: use pkcs11 openssl engine

tormath1:
To be tested with an actual payload.
SDK: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1275/cldsv/

TODO:
- [x] add the openssl configuration to consume the PCKS11 engine (https://github.com/OpenSC/libp11#using-the-engine-from-the-command-line) (see: https://github.com/flatcar/baselayout/pull/33)
- [x] include the "generate_payload" in this repo ? NOTE: we need to update the key URI to pass the pin-code
- [x] include the "download_payloads" in this repo? 
- [x] provide some documentation (like the /usr/sbin/pcscd daemon that needs to run in foreground) 
- [x] remove the hack around USB group for `acct-user/pcscd` (see the comment: https://github.com/flatcar/scripts/pull/1149#discussion_r1426745596)

The workflow would be the follow: (see: https://github.com/flatcar/flatcar-maintainer-private/pull/9) 
- start the SDK
- start the pcscd daemon
- plug the device
- ./generate_payloads alpha:3815.0.0 ...


core_sign_update
```diff
@@ -136,7 +136,8 @@ i=1
 signature_sizes=""
 for key in "${private_keys[@]}"; do
 	if [[ "${key}" == pkcs11* ]]; then
-		openssl rsautl -engine pkcs11 -pkcs -sign -inkey ${key} -keyform engine -in update.pkcs11-padhash -out update.sig.${i}
+		# NOTE: When we will use OpenSSL 3: use the following URI: pkcs11:id=%${ID}?pin-source=file:/tmp/pin
+		openssl pkeyutl -engine pkcs11 -sign -keyform engine -inkey "pkcs11:id=%${ID}?pin-value=$(cat /tmp/pin)" -in update.pkcs11-padhash -out update.sig.2
```
dongsupark, line 140: Why does it write to `update.sig.2` instead of `update.sig.${i}`?

pothos: And `${key}` is the PKCS#11 URI where I think the ID and pin-value should be part of. So instead of using cat /tmp this has to be specified in the invocation through the private_keys arg.
```suggestion
		openssl pkeyutl -engine pkcs11 -sign -keyform engine -inkey "${key}" -in update.pkcs11-padhash -out update.sig.${i}
```

pothos: In flatcar-build-scripts/generate_payload we pass a hardcoded `pkcs11:object=10;type=private` but we can totally change that to make use of a variable populated from a `read` to enter the pin.

pothos: (while at it we should delete the `dummy.key.pem` cargo cult)

pothos:
+1 for adding download_payloads/generate_payload to this repo. (What I don't like is making it mandatory to run inside the SDK because of of common.sh)

pothos:
> start the pcscd daemon

Inside the SDK or can I use the one from the host? Would it conflict with the one from the host? That was the case before and one had to manually stop it for successful signing.

pothos:
> * ./download_payloads alpha:3815.0.0 ...
> 
> * ./generate_payloads ./data/ keys (with a for loop on the releases in `./data/`)

In the end a big wrapper would be nice that downloads and uploads.

core_sign_update
pothos: > (while at it we should delete the `dummy.key.pem` cargo cult)

needs omitting or using 2 for `--public_key_version` in `core_sign_update`

pothos (commented):
Looks good!

pothos (approved):

pothos:
Let's merge and then remove the scripts from flatcar-build-scripts, otherwise things can get out of sync

sdk_container/src/third_party/portage-stable/acct-user/pcscd/pcscd-0-r2.ebuild
```diff
@@ -0,0 +1,13 @@
+# Copyright 2020-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit acct-user
+
+DESCRIPTION="A user for pcsc-lite"
+ACCT_USER_ID=47
+#ACCT_USER_GROUPS=( pcscd openct usb )
```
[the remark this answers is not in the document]
krnowak: The special handling is needed when the group ID specified in the ebuild is different from one that is already used in Flatcar. In case of `usb` group, the IDs are the same (85), so not sure what's going on here at first. Where's the error message?

krnowak (approved):
I assume that the scripts work. The changes in portage-stable and coreos-overlay look good. The baselayout will need to have its commit ID updated, of course.

One question though: why putting `download_payloads` script in the `data` subdirectory? The `generate payload` scripts could just create the directory and call `download_payloads` from there.

PR #1159, opened 2023-09-21
Upgrade Linux Firmware in main from 20230804 to 20230919

flatcar-infra:
Subject says it all.


dongsupark (approved):
CI passed, except for already known issues.

PR #1169, opened 2023-09-22
Generate reports for OEM and base sysext images, allow reports against last nightly

krnowak:
~Needs https://github.com/flatcar/flatcar-build-scripts/pull/149 first. Tested locally.~
~Needs https://github.com/flatcar/flatcar-build-scripts/pull/153 now.~

This does mostly two things:
- Image changes job prints the changes reports for each of the OEM sysext image and base sysexts too.
- Image changes job can print the changes reports against previous nightly.


pothos:
Would be good to have a Jenkins run for that

krnowak:
> Would be good to have a Jenkins and GA run for that

For sure. I was waiting for the flatcar-build-scripts stuff to go in. Will kick off them soon.

krnowak:
Jenkins: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2585/cldsv

krnowak:
Seems to be working in general. OEM package reports are not working yet, because there are no such reports on bincache - this PR introduces building such reports. On github the OEM reports are useless, because we build only the qemu images, so azure, vmware, ami and others only have error messages about missing files.

Jenkins: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2709/cldsv/
Github artifacts (search for image-changes-reports): https://github.com/flatcar/scripts/actions/runs/6497561930?pr=1169

ci-automation/base_sysexts.sh
```diff
@@ -0,0 +1,16 @@
+# Definitions of base sysexts to be built, for each arch. Used by
+# image.sh and image_changes.sh.
```
pothos, line 2: Why would they differ by arch?

pothos (approved):
Thanks! The invocation of the scripts is "historically grown" now :D

Just one note: On one hand it would be nice to do a hard error when encountering an error but on the other hand this means one wouldn't see the full output. It's ok to leave it as is but I fear it might be broken and people just skip over it because the script gives no final error. Examples are the use of `source <(curl…)` or `local var=$(cmd)` that silence errors.

ci-automation/base_sysexts.sh
krnowak: Ideally they should not differ and for now they do not. Maybe I could add common, amd64-only and arm64-only lists, just like I do for OEM sysexts. This would avoid the duplication of sysext names, but still be ready for an arch-specific base-sysext if such ever appears. Something I could do in follow-up PR. Or maybe just keep it simple.

krnowak:
> Thanks! The invocation of the scripts is "historically grown" now :D

Yeah, generalize some stuff and you end up with a mess like this. :)

> Just one note: On one hand it would be nice to do a hard error when encountering an error but on the other hand this means one wouldn't see the full output. It's ok to leave it as is but I fear it might be broken and people just skip over it because the script gives no final error. Examples are the use of `source <(curl…)` or `local var=$(cmd)` that silence errors.

My aim was that running package-diff or size-change-report.sh or show-changes may fail without failing the whole job. I think it was like this from the beginning, actually. But stuff like `local var=$(stuff)` swallowing errors is something I tried to avoid, so it's certainly something I'd prefer to fix. `source <(curl)` is nasty - looks short and sweet, but it completely swallows errors. I'd need to come up with a different solution.

I have just read a bit about process substitution and error propagation and maybe we could do `source <(curl … || kill $$)`. It seems to be working, but only for non-interactive shells.

krnowak:
Thanks for the review. I'll try to merge it tomorrow. I just realized that the changes in jenkins-os will need to be gated on the new-enough-version of the built image or this PR will need to be backported to other channels.

ci-automation/base_sysexts.sh
pothos: I think we don't really want to have the images differ, any future amd64-only sysext images could also be external, i.e., pulled in at provisioning time with the logic we have for OEM sysext downloads.

Since it's a simple change to make it arch-specific I think it's better to keep one variable here without having to duplicate something for a use case we don't have.

But just a minor detail, feel free to ignore.

pothos:
Oh, yeah, backporting is tricky with the sysext stuff… Maybe some hacky if-else cases could prevent further divergence but I if I recall correctly this ship has sailed already for the image change job and it's already not easy to apply patches cleanly on the Stable/LTS branches, or?

ci-automation/base_sysexts.sh
krnowak: I simplified it.

krnowak:
> Just one note: On one hand it would be nice to do a hard error when encountering an error but on the other hand this means one wouldn't see the full output. It's ok to leave it as is but I fear it might be broken and people just skip over it because the script gives no final error. Examples are the use of `source <(curl…)` or `local var=$(cmd)` that silence errors.

I have split `local var=$(cmd)` into `local var; var=$(cmd)` and replaced the use of `<(curl…)` with curling to a temporary file and sourcing it. I kept `<(printf … | sort)` as I don't expect it to fail.

krnowak:
> Oh, yeah, backporting is tricky with the sysext stuff… Maybe some hacky if-else cases could prevent further divergence but I if I recall correctly this ship has sailed already for the image change job and it's already not easy to apply patches cleanly on the Stable/LTS branches, or?

I decided to not do the backporting as the advantages of doing it are rather slim - once a version hits a LTS/stable/beta/alpha channel, the image changes job is of limited use - almost nothing changes there comparing to the previous nightly in the channel. The important changes (like kernel version and kernel config changes and the like) are still printed.

Instead I went with a cut-off version of 3795 in jenkins os to decide whether to kick off two image-changes jobs or just one.

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2964/cldsv/

krnowak:
CI has passed. So for versions <3795, we will do the old stuff (image changes against last release, done after the image job, no changes in base sysexts and OEM sysexts), while for newer versions the new stuff (image changes against last release and last nightly, done after vms job, display changes in base sysexts and OEM sysexts).

PR #1171, opened 2023-09-22
sys-libs/glibc: update to patchlevel 2.37-rc5

t-lo:
Sync to upstream 2.37-rc5 and apply Flatcar specific patches.

Closes https://github.com/flatcar/Flatcar/issues/1179.

PR #1178, opened 2023-09-25
coreos sys-libs/glibc: Update manifest for changed patch set

krnowak:
Fixes a build error:

```
!!! Fetched file: glibc-2.37-patches-7.tar.xz VERIFY FAILED!
!!! Reason: Insufficient data for checksum verification
!!! Got:      
!!! Expected: BLAKE2B BLAKE2S MD5 RMD160 SHA1 SHA256 SHA3_256 SHA3_512 SHA512 WHIRLPOOL
 * Fetch failed for 'sys-libs/glibc-2.37-r5'
```


t-lo (approved):
I broke this in https://github.com/flatcar/scripts/pull/1171, thank you for fixing @krnowak .

PR #1197, opened 2023-09-27
oem: port digitalocean, equinix metal (packet) and openstack

tormath1:
These OEM do not ship any software, they only use systemd services (`coreos-metadata-sshkeys@.service` and `packet-phone-home.service`).

Need:
- https://github.com/flatcar/init/pull/107

Closes:
- https://github.com/flatcar/Flatcar/issues/1143
- https://github.com/flatcar/Flatcar/issues/1145
- https://github.com/flatcar/Flatcar/issues/1142

I don't think a changelog entry is required as the sysext images are "empty".

---
- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2644/cldsv/


pothos:
> Note: I dropped `brightbox` should we keep it? I don't see any documentation regarding this.

I think it's a good idea to clean up unused/untested OEM setups, best would be in a follow-up and not linked to this PR, or?

sdk_container/src/third_party/coreos-overlay/coreos-base/oem-packet/oem-packet-0.2.2-r2.ebuild
```diff
@@ -0,0 +1,15 @@
...
+DESCRIPTION="Equinix Metal suite for Equinix Metal"
+HOMEPAGE="https://www.equinix.com/"
+SRC_URI=""
+
+LICENSE="GPL-2"
+SLOT="0"
+KEYWORDS="amd64 arm64"
+IUSE=""
+
+OEM_NAME="Packet"
```
pothos, line 15: We can rename this variable, it's used in the shell prompt message but otherwise I think it's not relevant

pothos (commented):
The grub cfg is gone but it has to be moved to the new place

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/coreos-metadata.service
```diff
@@ -1,4 +1,22 @@
...
+ConditionKernelCommandLine=|flatcar.coreos.id=digitalocean
+
+ConditionKernelCommandLine=|flatcar.oem.id=gce
+ConditionKernelCommandLine=|flatcar.coreos.id=gce
+
+ConditionKernelCommandLine=|flatcar.oem.id=openstack
+ConditionKernelCommandLine=|flatcar.coreos.id=openstack
+
+ConditionKernelCommandLine=|flatcar.oem.id=packet
+ConditionKernelCommandLine=|flatcar.coreos.id=packet
```
pothos, line 18: This should be `coreos.oem.id`

pothos (commented):
With OEM sysext images we can also use the migration mechanism to clean up the old OEM contents. (This will also make the sysext images required on update.)

sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/files/oems/packet
```diff
@@ -0,0 +1,3 @@
+/etc/systemd/system/packet-phone-home.service
```
pothos, line 1: Do we want to add `/etc/systemd/system/multi-user.target.requires/coreos-metadata-sshkeys@core.service` here, too, because it's also in `ami`?

pothos: same for DO and openstack

pothos (approved):

PR #1200, opened 2023-09-29
overlay coreos-base/update_engine: fix build issues with gcc 13

dongsupark:
Starting from gcc 13, it is required to explicitly include cstdint, as other standard library header files will not include other headers that were being used internally by the library. So update_engine also has to explicitly include cstdint to avoid build errors.

Pulls in https://github.com/flatcar/update_engine/pull/27.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1087/cldsv/

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)


dongsupark:
SDK and packages of CI passed.

krnowak (approved):

krnowak:
I'll merge it to give today's nightly a chance to build.

PR #1210, opened 2023-10-04
coreos-overlay: Add support for new EKS versions

hhoover:
# Adds S3 download locations for newer EKS versions

Flatcar out of the box only works for EKS 1.23 and older, this adds support for newer versions in the download_kubelet.sh script. This change adds download URLs for EKS 1.24, 1.25, 1.26, 1.27 and 1.28 so images can more easily be used for EKS workers.

Similar to https://github.com/flatcar/coreos-overlay/pull/2110

## How to use

Deploy an image as an EKS worker on cluster versions 1.24-1.28 (or just verify the downloads work)

## Testing done

Tested S3 download URLs using `aws s3 sync` as it is in the script for other versions.


hhoover:
Similar to https://github.com/flatcar/coreos-overlay/pull/2110

jepio:
Thanks for this contribution @hhoover, i've pushed an extra commit to include the PR link in the changelog (i know its annoying to do because you only get the pr number after you open the pr).

hhoover:
@jepio looks like builds failed not related to the PR - did I miss something?

jepio:
No, you did everything all right. The failing test cases don't work for PRs from forks, I just ran the CI to see that things build correctly.

Again, thanks for this PR. We appreciate when users scratch their own itch :)

PR #982, opened 2023-07-07
Ship docker sysext and rip out torcx [WIP]

krishjainx:
# Ship Docker and Containerd Sysext and deprecate torcx

Solves https://github.com/flatcar/Flatcar/issues/443

Related Issues: https://github.com/flatcar/Flatcar/issues/1092, https://github.com/flatcar/Flatcar/issues/1093

Once we have it building in the CI people can download the image from there and play around already.

## How to use

@pothos recommended I file a draft PR for aiding communication

## Testing done

When I attempt to use run build_image script and run the launch script the Flatcar provided sysext  works.
1.

```
core@localhost ~ $ sudo systemctl status docker.socket
● docker.socket - Docker Socket for the API
     Loaded: loaded (/usr/lib/systemd/system/docker.socket; disabled; preset: d>
     Active: active (running) since Sat 2023-07-15 20:18:27 UTC; 40s ago
   Triggers: ● docker.service
     Listen: /run/docker.sock (Stream)
      Tasks: 0 (limit: 7293)
     Memory: 0B
        CPU: 438us
     CGroup: /system.slice/docker.socket

Jul 15 20:18:27 localhost systemd[1]: Starting docker.socket - Docker Socket fo>
Jul 15 20:18:27 localhost systemd[1]: Listening on docker.socket - Docker Socke>
core@localhost ~ $ sudo systemctl status containerd
● containerd.service - containerd container runtime
     Loaded: loaded (/usr/lib/systemd/system/containerd.service; disabled; pres>
     Active: active (running) since Sat 2023-07-15 20:18:27 UTC; 45s ago
       Docs: https://containerd.io
    Process: 1281 ExecStartPre=mkdir -p /run/docker/libcontainerd (code=exited,>
    Process: 1308 ExecStartPre=ln -fs /run/containerd/containerd.sock /run/dock>
   Main PID: 1316 (containerd)
      Tasks: 15
     Memory: 68.6M
        CPU: 253ms
     CGroup: /system.slice/containerd.service
             └─1316 /usr/bin/containerd --config /usr/share/containerd/config.t>

Jul 15 20:18:27 localhost containerd[1316]: time="2023-07-15T20:18:27.957570412>
Jul 15 20:18:27 localhost containerd[1316]: time="2023-07-15T20:18:27.957619335>
Jul 15 20:18:27 localhost containerd[1316]: time="2023-07-15T20:18:27.958557496>
[...]

PR #1216, opened 2023-10-05
Contrib: Deprecate torcx, ship containerd / docker as sysexts

t-lo:
This PR supersedes https://github.com/flatcar/scripts/pull/982 which has been idling for a while. Main motivation of re-issuing the PR is to provide a branch directly on the scripts repo for Flatcar maintainers to collaborate on.

**Features added**<br />
The PR includes a number of improvements to streamline the developer experience with sysexts. The improvements are:
- `build_sysext` can now handle dependencies and create sysexts that depend on packages of other sysexts
- `build_image` takes a command line option of the sysexts to include in the base OS, defaulting to  `containerd` and `docker`. Motivation is to make it easier down the road to move base OS packages into sysexts, and to build leaner OS images.

**Testing**
* Github actions CI / test: https://github.com/flatcar/scripts/actions/runs/6614425214/job/17964395526?pr=1216
* Jenkins CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2791/cldsv/

**Unrelated minor improvements**
* The PR also includes a few unrelated QoL improvements which were implemented along the way to ease development work and testing of this PR:
  * Add [`run_local_tests.sh`](https://github.com/flatcar/scripts/pull/1216/commits/b9ff2f49832edc30d7bbd0a374559bf658c45bda) to lower the bar of running tests locally for the latest image built
  * Make the update payload configurable for the [`qemu_update.sh`](https://github.com/flatcar/scripts/pull/1216/commits/55bf466feb50c9b6c2692085436025d91cc6b9ec) test case. Formerly it was always fetched from bincache; fetch failures were silently ignored and lead to this test being skipped.
  * Add a new flag to `run_sdk_container` to not touch / update the version file ([`-U`](https://github.com/flatcar/scripts/pull/1216/commits/60c90cfaec87316e92fea9f7f1e3304f85b7e8e3))
  * [Fix detection](https://github.com/flatcar/scripts/pull/1216/commits/4dc20d581f8fafcacc0ab142af2e62e5b4bb30b8) of "SDK container image exists in local Docker registry and is up to date" in ci-automation. Formerly, this check always failed for nightly builds, leading to the SDK container being re-downloaded each time an SDK container was newly instanciated (e.g. after running `docker container prune`).

**Dependencies** (squashfs zstd support for Jenkins CI nodes)
* https://github.com/flatcar/scripts/pull/1297
* https://github.com/flatcar/scripts/pull/1296
* https://github.com/flatcar/scripts/pull/1295
* https://github.com/flatcar/flatcar-linux-infra/pull/268

**[ALL DONE] Todo items**
* [x] Update and rename [torcx mantle / kola tests](https://github.com/flatcar/Flatcar/issues/1148) to handle sysext, add test for disabling docker: https://github.com/flatcar/mantle/pull/465
* [x] Remove torcx logic and command line options from `build_packages`, `build_image`, and friends
* [x] Add SLSA provenance generation to docker / containerd sysext builds and ensure SPDX SBOMs, file lists, and licenses are covered
* [x] Update github actions CI and kola workflows to not handle torcx anymore
* [x] Update docker docs https://www.flatcar.org/docs/latest/container-runtimes/getting-started-with-docker/#permanently-running-a-container to include a soft-link `/etc/systemd/system/multi-user.target.wants/docker.service => /usr/lib/systemd/system/docker.service`  for sysext based docker. See `kola/tests/docker/enable_docker.go` for details. : https://github.com/flatcar/flatcar-docs/pull/343
* [x] Remove torcx sanity checks in update_engine post-install script: https://github.com/flatcar/update_engine/pull/30


build_library/build_image_util.sh
```diff
@@ -626,26 +644,18 @@ finish_image() {
...
+  sudo "${SCRIPTS_DIR}/build_sysext" --board="${BOARD}" --image_builddir=${BUILD_DIR} --squashfs_base="${BUILD_DIR}/${image_sysext_base}" --manglefs_script="${SCRIPTS_DIR}/manglefs_docker" docker-flatcar app-containers/docker
+  sudo install -m 0644 -D "${BUILD_DIR}/docker-flatcar.raw" "${root_fs_dir}"/usr/share/flatcar/
+  sudo mkdir -p "${root_fs_dir}"/etc/extensions/
+  sudo ln -sf /usr/share/flatcar/docker-flatcar.raw "${root_fs_dir}"/etc/extensions/docker-flatcar.raw
+  sudo rm ${PORTAGE_CONFIGROOT}/etc/portage/profile/package.provided
+
+  sudo "${SCRIPTS_DIR}/build_sysext" --board="${BOARD}" --image_builddir=${BUILD_DIR} --squashfs_base="${BUILD_DIR}/${image_sysext_base}" --manglefs_script="${SCRIPTS_DIR}/manglefs_containerd" containerd-flatcar app-containers/containerd
+  sudo install -m 0644 -D "${BUILD_DIR}/containerd-flatcar.raw" "${root_fs_dir}"/usr/share/flatcar/
+  sudo ln -sf /usr/share/flatcar/containerd-flatcar.raw "${root_fs_dir}"/etc/extensions/containerd-flatcar.raw
```
jepio, line 658: @pothos does this work?

build_library/dev_container_util.sh
```diff
@@ -110,7 +110,7 @@ create_dev_container() {
   # The remount services are provided by coreos-base/coreos-init
   systemd_enable "${root_fs_dir}" "multi-user.target" "remount-usr.service"
 
-  finish_image "${image_name}" "${disk_layout}" "${root_fs_dir}" "${image_contents}" "${image_contents_wtd}"
+  DEVCONTAINER=1 finish_image "${image_name}" "${disk_layout}" "${root_fs_dir}" "${image_contents}" "${image_contents_wtd}"
```
jepio, line 113: this look unused

t-lo: Yeah, the whole thing needs some cleaning up. This was added in a no-comment commit without justification, and the environment variable does not appear anywhere else in the code.

build_library/build_image_util.sh
t-lo: I also don't quite understand why we first create a docker sysext pretending containerd is installed via `${PORTAGE_CONFIGROOT}/etc/portage/profile/package.provided` and then create a containerd sysext anyway.

pothos: That was done to have a split containerd and Docker sysext setup. This way users can disable Docker but keep our containerd.

pothos: (The pretending of containerd being installed is to build the Docker sysext without containerd in it.)

t-lo: @jepio to answer your question this works because of https://github.com/flatcar/scripts/blob/7c255452fac3e6ef67c076a642d0aa08887cf3e4/build_library/build_image_util.sh#L793-L803.

t-lo: > That was done to have a split containerd and Docker sysext setup. This way users can disable Docker but keep our containerd.

Oh I understand the intent, but I'd argue that if we want stackable / dependent sysexts then we should make the packages shipped with each layer explicit, and re-use these when we build the next "upper" layer.

I've extended (build_sysext.sh)[https://github.com/flatcar/scripts/blob/contrib/torcx-deprecation-docker-sysext/build_sysext#L30-L33] to cover this, and updated (build_image_util.sh)[https://github.com/flatcar/scripts/blob/contrib/torcx-deprecation-docker-sysext/build_library/build_image_util.sh#L640-L655] to use it.

t-lo:
The PR in its current state:
- can build OS images
- does not build any torcx packages or artifacts
- includes docker and containerd sysexts in OS image and update tarball, which have been manually tested and are working (qemu)

build_library/build_image_util.sh
jepio: That looks more complex than it needs to be. How about just mount the whole previous sysext when building the next one.

t-lo: Sysexts do not contain package information since they only cover `/usr` (and/or `/opt`). Installed packages are registered in sub-directories below `/var/db/pkg`.
(I _wish_ this was easier though!)

t-lo:
Added SBOM, licenses, packages and file listing / disk usage inventory information for sysexts.


- packages list and SBOM / licenses JSON now include all packages of the final image, i.e. a combined list of base image and all base OS sysexts.
- Files list and detailed files list include the sysext squashfs files for the base image, and separate sections with files lists for each sysext.
- Disk usage contains both final disk image usage as well as usage of each individual sysext squashfs.

pothos (changes requested):
Please change the name `app-containers_docker.raw` and `app-containers_containerd.raw` to `docker-flatcar.raw` and `containerd-flatcar.raw`. We have documented those for ages in
https://www.flatcar.org/docs/latest/container-runtimes/use-a-custom-docker-or-containerd-version/
and
https://www.flatcar.org/docs/latest/provisioning/sysext/#supplying-your-sysext-image-from-ignition
for a smooth upgrade experience which doesn't break user setups.

Also, `app-containers` is a gentooism that isn't really meaningful to users. While finally better than `app-emulation` we just shouldn't use Gentoo packaging categories for user-facing names. Also, we might want to add more things into the mix and have other Flatcar extensions as well, and the use of Gentoo packaging categories is not a good idea for consistency.

pothos:
> [flatcar_production_image_contents.txt](https://github.com/flatcar/scripts/files/12856456/flatcar_production_image_contents.txt)

The concatenation is not good for automated processing. We use this file for the image change reports and would have to split all these `### Sysext app-containers_docker.raw` headers. Let's have separate files, instead of making our life harder.

t-lo:
> > [flatcar_production_image_contents.txt](https://github.com/flatcar/scripts/files/12856456/flatcar_production_image_contents.txt)
> 
> The concatenation is not good for automated processing. We use this file for the image change reports and would have to split all these `### Sysext app-containers_docker.raw` headers. Let's have separate files, instead of making our life harder.

Listing sysext contents in these files in separate sections was an explicit ask. We additionally ship separate files per sysext, so processing everything should be rather simple: stop parsing `flatcar_production_image_contents.txt` at the first `^#`.

t-lo:
> Please change the name `app-containers_docker.raw` and `app-containers_containerd.raw` to `docker-flatcar.raw` and `containerd-flatcar.raw`. We have documented those for ages in https://www.flatcar.org/docs/latest/container-runtimes/use-a-custom-docker-or-containerd-version/ and https://www.flatcar.org/docs/latest/provisioning/sysext/#supplying-your-sysext-image-from-ignition for a smooth upgrade experience which doesn't break user setups.
> 
> Also, `app-containers` is a gentooism that isn't really meaningful to users. While finally better than `app-emulation` we just shouldn't use Gentoo packaging categories for user-facing names. Also, we might want to add more things into the mix and have other Flatcar extensions as well, and the use of Gentoo packaging categories is not a good idea for consistency.

Got it, will update. It's a bit sad because this convention will make it harder to generically split more apps from the base image into sysexts since we need to maintain app -> name relationships for each, while the current implementation guarantees unique names.

krnowak:
> > > [flatcar_production_image_contents.txt](https://github.com/flatcar/scripts/files/12856456/flatcar_production_image_contents.txt)
> > 
> > 
> > The concatenation is not good for automated processing. We use this file for the image change reports and would have to split all these `### Sysext app-containers_docker.raw` headers. Let's have separate files, instead of making our life harder.
> 
> Listing sysext contents in these files in separate sections was an explicit ask. We additionally ship separate files per sysext, so processing everything should be rather simple: stop parsing `flatcar_production_image_contents.txt` at the first `^#`.

Or maybe even better - create a separate file, like `flatcar_production_image_contents_kitchen_sink.txt`, and leave other files as they were.

t-lo:
> > > > [flatcar_production_image_contents.txt](https://github.com/flatcar/scripts/files/12856456/flatcar_production_image_contents.txt)
> > > 
> > > 
> > > The concatenation is not good for automated processing. We use this file for the image change reports and would have to split all these `### Sysext app-containers_docker.raw` headers. Let's have separate files, instead of making our life harder.
> > 
> > 
> > Listing sysext contents in these files in separate sections was an explicit ask. We additionally ship separate files per sysext, so processing everything should be rather simple: stop parsing `flatcar_production_image_contents.txt` at the first `^#`.
> 
> Or maybe even better - create a separate file, like `flatcar_production_image_contents_kitchen_sink.txt`, and leave other files as they were.

That's what I had initially, then the request came up to create merged lists. I'm not complaining though, the way you propose it is simpler and requires less code. Will update the PR.

pothos:
Maybe the communication about the files got confusing because they serve different purposes and land on different places. For the license JSON it makes sense to have a merged view while for the contents txt files it rather doesn't because they are developer-focused. For the SBOM I think the merged view makes sense, too (even if it doesn't truly reflect what the user finds when opening the image, but I think that live SSH interaction for a scanner would also be a valid use case, so better list the files than not).

t-lo:
Updated the PR. Please find new inventory sample files attached.
[flatcar_production_image_contents.txt](https://github.com/flatcar/scripts/files/12894691/flatcar_production_image_contents.txt)
[flatcar_production_image_contents_wtd.txt](https://github.com/flatcar/scripts/files/12894692/flatcar_production_image_contents_wtd.txt)
[flatcar_production_image_disk_usage.txt](https://github.com/flatcar/scripts/files/12894693/flatcar_production_image_disk_usage.txt)
[flatcar_production_image_licenses.json](https://github.com/flatcar/scripts/files/12894694/flatcar_production_image_licenses.json)
[flatcar_production_image_packages.txt](https://github.com/flatcar/scripts/files/12894695/flatcar_production_image_packages.txt)
[flatcar_production_image_sbom.json](https://github.com/flatcar/scripts/files/12894696/flatcar_production_image_sbom.json)
[containerd-flatcar_contents.txt](https://github.com/flatcar/scripts/files/12894697/containerd-flatcar_contents.txt)
[containerd-flatcar_contents_wtd.txt](https://github.com/flatcar/scripts/files/12894698/containerd-flatcar_contents_wtd.txt)
[containerd-flatcar_disk_usage.txt](https://github.com/flatcar/scripts/files/12894700/containerd-flatcar_disk_usage.txt)
[containerd-flatcar_packages.txt](https://github.com/flatcar/scripts/files/12894701/containerd-flatcar_packages.txt)
[docker-flatcar_contents.txt](https://github.com/flatcar/scripts/files/12894702/docker-flatcar_contents.txt)
[docker-flatcar_contents_wtd.txt](https://github.com/flatcar/scripts/files/12894703/docker-flatcar_contents_wtd.txt)
[docker-flatcar_disk_usage.txt](https://github.com/flatcar/scripts/files/12894705/docker-flatcar_disk_usage.txt)
[docker-flatcar_packages.txt](https://github.com/flatcar/scripts/files/12894706/docker-flatcar_packages.txt)

t-lo:
Most notably I extended the format of `--base-sysexts=`  in [build_image](https://github.com/flatcar/scripts/blob/contrib/torcx-deprecation-docker-sysext/build_image#L36-L37) to support sysext names independent of the packages included. The parameter currently defaults to `containerd-flatcar:app-containers/containerd,docker-flatcar:app-containers/docker` so we get a containerd sysext named `containerd-flatcar.raw` and a docker sysext named `docker-flatcar.raw`.

t-lo:
Filed a [kola / mantle PR](https://github.com/flatcar/mantle/pull/465), rebased to latest main.

run_local_tests.sh
```diff
@@ -0,0 +1,108 @@
...
+  local custom_test_list=false
+
+  # Generate list of all tests for qemu w/o the devcontainer tests.
+  # This will generate globs for top-level test modules, e.g. "cl.update.oem" will become cl.*.
+  # Globs are necessary because tests ignore OS min/max version specification if a test was specified with its full name.
+  # Using globs will prevent tests to be run which aren't meant for the OS version we're testing.
+  if [[ $# -eq 0 ]] ; then
+    tests="$(docker run "${mantle_container}" \
+              kola list --platform qemu \
+              | awk '!/^(devcontainer|Test)/ {if ($1 != "") print gensub(/^([^.]+).*/,"\\1",1,$1) ".*"}' | uniq)"
```
pothos, line 83: Maybe we should add some skipfunc logic to the devcontainer test to not run if there is nothing on bincache? That's what I did for the OEM sysext tests. This way we could use the regular invocation instead of filtering here.

pothos:
Thanks! I've left one more small comment. I didn't have a close look at the mounting yet but in general the changes look good.

run_local_tests.sh
t-lo: I'd rather fix the devcontainer for good :) And make it more robust. It's next on my list of things to be sysext-ified.

t-lo:
Torcx artifact handling was removed from `ci-automation` so Jenkins CI should "just work". Build here: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2748/cldsv/

t-lo:
Meh. `cl.update.payload`  gives me (in the journal of the node that tried to update):
```
Oct 17 12:12:08.935854 update_engine[1420]: Failed to locate torcx binary in /tmp/au_postint_mount.IpwKXz
```

which is from the [flatcar-postinst](https://github.com/flatcar/update_engine/blob/flatcar-master/flatcar-postinst#L298-L325) script. It seems like we need a version check (of `${NEXT_VERSION}`) there, too.

t-lo:
Test build on Jenkins CI is here: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2755/cldsv/

t-lo:
Rebased to latest main, new build: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2774/cldsv/ (also here in GH actions CI)

t-lo:
Jenkins CI is all green! http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2775/cldsv/

t-lo:
Github CI tests are green too!

changelog/changes/2023-10-17-sdk-mantle-kola-sysext.md
```diff
@@ -0,0 +1 @@
+- SDK: bump Mantle / Kola container image version to include torcx removal / sysext tests ([Mantle#466](https://github.com/flatcar/mantle/pull/466), [Mantle#465](https://github.com/flatcar/mantle/pull/465))
```
pothos, line 1: This entry file can be removed because it's rather something for the git log and not for the user-facing release notes

changelog/changes/2023-10-17-update-engine-torcx-removal.md
```diff
@@ -0,0 +1 @@
+- update_engine: remove torcx sanity checks because the OS image removed torcx ([update_engine#30](https://github.com/flatcar/update_engine/pull/30))
```
pothos, line 1: Same here, this is not interesting for the user. What I actually would want to read as a user is that torcx is gone and what to do now - a pointer to https://www.flatcar.org/docs/latest/provisioning/sysext/ should be enough for the details but a good one sentence summary should tell that torcx is removed because systemd-sysext provides the same functionality in a more generic and distro-independent way.

changelog/changes/2023-10-17-sdk-mantle-kola-sysext.md
t-lo: Gotcha, will update.

changelog/changes/2023-10-17-update-engine-torcx-removal.md
t-lo: That's a good point. Will add that and the docs pointer, and mention the relevant changes for OS, update_engine and mantle there.

t-lo:
Added a fix for a long-standing issue in `ci-automation/ci_automation_common.sh` failing to detect local SDK container images if the SDK is from a nightly build.

changelog/changes/2023-10-19-torcx-removal.md
```diff
@@ -0,0 +1 @@
+- torcx was replaced by systemd-sysext in the OS image. Learn more about sysext and how to customise OS images [here](https://www.flatcar.org/docs/latest/provisioning/sysext/). `update_engine` and the `kola` test suite in Mantle have been adapted to this change accordingly. ([scripts#1216](https://github.com/flatcar/scripts/pull/1216), [update_engine#30](https://github.com/flatcar/update_engine/pull/30), [Mantle#466](https://github.com/flatcar/mantle/pull/466), [Mantle#465](https://github.com/flatcar/mantle/pull/465)).
```
pothos, line 1:
```suggestion
- Torcx, the mechanism to switch Docker versions, was replaced systemd-sysext in the OS image. Learn more about sysext and building your own OS extension images [here](https://www.flatcar.org/docs/latest/provisioning/sysext/). `update_engine` and the `kola` test suite in Mantle have been adapted to this change accordingly. ([scripts#1216](https://github.com/flatcar/scripts/pull/1216), [update_engine#30](https://github.com/flatcar/update_engine/pull/30), [Mantle#466](https://github.com/flatcar/mantle/pull/466), [Mantle#465](https://github.com/flatcar/mantle/pull/465)).
```

pothos: I still think that the Mantle/update_engine changes don't belong in the end-user release notes because they are internal

t-lo: Regularly I would agree - but in this case we're introducing significant changes. I just want to make sure we leave a paper trail and we make sufficient noise around it.

pothos: Then let's add more words around the change. The mantle/update_engine mentions are completely irrelevant to users which could rather mean that they skip reading the whole entry because it just looks like noise to them.

pothos: E.g., `You can convert existing Torcx images with the helper script, use the published systemd-sysext images for Docker/Kubernetes from the sysext-bakery, or build your own sysext images for particular Docker versions.`

t-lo: I've elaborated more and added sub-bullet-points to the changelog entry. This isn't really what you asked but I hope it makes it more interesting for users to read, and also adds the pointers we've discussed here in a more user friendly way.

pothos: Thanks, that's much better :)

pothos (approved):

pothos:
The mantle container ref bump is something we normally rather let the Github Action do and this could be dropped here.

sdk_container/.repo/manifests/mantle-container
```diff
@@ -1 +1 @@
-ghcr.io/flatcar/mantle:git-790b16559be109a124536110a6841251d7cacb71
+ghcr.io/flatcar/mantle:git-20a2f8ffee8c8a1a042b1da99f0f59312110f285
```
pothos, line 1: This ref is outdated, `d638477431717aae4d87909d5ee5e39f009a803f` is the latest as of now. I merged the mantle bumps and your changes are in, so you can drop this change here.

pothos:
> Hm, no, seems to be really the compression differences in squashfs vs the tar file:

Not sure if our kernel supports it but we could try `mksquashfs -comp zstd` to reduce the inbuilt sysext image size.

t-lo:
> Interesting, I expected it to reduce the filesystem usage on `/usr` but it didn't:
> 

It makes sense when you think about it: we replace the gzipped torcx tarball baked into `/usr` with 2 compressed squashfs files. I'd expect the sizes to be roughly similar.

> Not sure if our kernel supports it but we could try mksquashfs -comp zstd to reduce the inbuilt sysext image size. (Later, not part of this PR.)

That's a great idea! Our kernel currently does not (`# CONFIG_SQUASHFS_ZSTD is not set` in https://alpha.release.flatcar-linux.net/amd64-usr/current/flatcar_production_image_kernel_config.txt) but we should definitely change that before the next Alpha.

sdk_container/.repo/manifests/mantle-container
t-lo: Rebased on latest main, so this got pulled in automatically. I made sure the ref is correct though.

build_sysext
```diff
@@ -134,11 +143,34 @@ trap cleanup EXIT
...
+    ppath="$(readlink -f "${entry}")"
+    if [[ ! -f "${ppath}" ]] ; then
+      error "--base_pkginfo contains invalid entries."
+      error "Pkginfo file '${ppath}' does not exist."
+      die "Full --base_pkginfo: '${FLAGS_base_pkginfo}'"
+    fi
+
+    pfile="$(basename "${ppath}")"
+    pmdir="${BUILD_DIR}/base-pkginfo/${pfile}"
+    mkdir -p "${pmdir}"
```
krnowak, line 161: Should we check if `pmdir` didn't exist before to avoid silently mounting the squashfs over already existing mount?

build_sysext
```diff
@@ -134,11 +143,34 @@ trap cleanup EXIT
...
+    mount -rt squashfs -o loop,nodev "${ppath}" "${pmdir}"
+    pkginfo_lowerdirs="${pkginfo_lowerdirs}:${pmdir}"
```
krnowak, line 163: Is order of lowerdirs important here? (It's about unusual right-to-left ordering of layers in overlayfs lowerdir option).

build_sysext
```diff
@@ -186,6 +218,17 @@ if [[ -n "${FLAGS_manglefs_script}" ]]; then
...
+  info "  Creating pkginfo squashfs '${BUILD_DIR}/${SYSEXTNAME}_pkginfo.raw'"
+  mkdir -p "${BUILD_DIR}/img-pkginfo/var/db"
+  cp -R "${BUILD_DIR}/install-root/var/db/pkg" "${BUILD_DIR}/img-pkginfo/var/db/"
+  mksquashfs "${BUILD_DIR}/img-pkginfo" "${BUILD_DIR}/${SYSEXTNAME}_pkginfo.raw" -noappend
+fi
+
+info "Writing ${SYSEXTNAME}_packages.txt"
+ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="${BUILD_DIR}/install-root"\
+      equery --no-color list --format '$cpv::$repo' '*' > "${BUILD_DIR}/${SYSEXTNAME}_packages.txt"
+
```
krnowak, line 231: I think that this code should run before the manglefs script is invoked as it could easily just do `rm -rf "${rootfs}/var"`, resulting in no database and no packages listing.

run_local_tests.sh
```diff
@@ -0,0 +1,108 @@
...
+  local arch="${1}"
+  local parallel="${2}"
+
+  # Read by the mantle container.
+  # The local directory ("pwd") will be mounted to /work/ in the container.
+  cat > sdk_container/.env <<EOF
+export export QEMU_IMAGE_NAME="/work/__build__/images/images/${arch}-usr/latest/flatcar_production_image.bin"
+export QEMU_UEFI_BIOS="/work/__build__/images/images/${arch}-usr/latest/flatcar_production_qemu_uefi_efi_code.fd"
+export QEMU_UPDATE_PAYLOAD="/work/__build__/images/images/${arch}-usr/latest/flatcar_test_update.gz"
+export PARALLEL_TESTS=${parallel}
```
krnowak, line 57:
```suggestion
export QEMU_IMAGE_NAME="/work/__build__/images/images/${arch@Q}-usr/latest/flatcar_production_image.bin"
export QEMU_UEFI_BIOS="/work/__build__/images/images/${arch@Q}-usr/latest/flatcar_production_qemu_uefi_efi_code.fd"
export QEMU_UPDATE_PAYLOAD="/work/__build__/images/images/${arch@Q}-usr/latest/flatcar_test_update.gz"
export PARALLEL_TESTS=${parallel@Q}
```

run_local_tests.sh
```diff
@@ -0,0 +1,108 @@
...
+export QEMU_UPDATE_PAYLOAD="/work/__build__/images/images/${arch}-usr/latest/flatcar_test_update.gz"
+export PARALLEL_TESTS=${parallel}                                                     
+EOF
+
+  export MAX_RETRIES=5
+  export SKIP_COPY_TO_BINCACHE=1
+}
+#--
+
+function run_local_tests() {
```
krnowak, line 65: You are sourcing a file in this function, so I think it would be better to run it in a subshell.
```suggestion
function run_local_tests() (
```

run_sdk_container
```diff
@@ -45,11 +45,13 @@ usage() {
     echo "      -C           - Use an entirely custom container image instead of the SDK's"
     echo "                     $sdk_container_common_registry/flatcar-sdk-[ARCH]:[SDK VERSION]"
     echo "                     Useful for CI."
+    echo "      -U             Do not update the versionfile. Instead, use the version from the versionfile as-is."
```
krnowak, line 48: Lovely, I was thinking about the same today.

build_library/build_image_util.sh
```diff
@@ -260,8 +260,19 @@ image_packages_portage() {
...
+
+# List dependencies for a package runtime dependencies
+
+function package_run_dependencies() (
+  pkg=${1:?}
+  ebuild=$(equery-${BOARD} w "${pkg}")
+  function inherit() { : ; }
+  . "${ebuild}"
+  echo ${RDEPEND}
+)
```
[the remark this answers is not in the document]
t-lo: Umm, I had no idea this was still there :sweat_smile: It's part of a hack Krish introduced when this was still in proof-of-concept stage. Back in the day we hard-wired docker and containerd packages and dependency lists into the software inventory files generated during build.

Will remove this.

build_library/sysext_prod_builder
```diff
@@ -0,0 +1,154 @@
...
+            "${name}" "${grp_pkg}"
+
+  sudo mv "${workdir}/sysext-build/${name}.raw" "${workdir}/sysext-build/${name}_pkginfo.raw" \
+                              "${workdir}/sysext-build/${name}"_*.txt "${output_dir}"
+
+  sudo mkdir -p "${install_root}"/usr/share/flatcar/sysext
+  sudo install -m 0644 -D "${output_dir}/${name}.raw" "${install_root}"/usr/share/flatcar/sysext/
+
+  sudo mkdir -p "${install_root}"/etc/extensions/
+  sudo ln -sf "/usr/share/flatcar/sysext/${name}.raw" "${install_root}/etc/extensions/${name}.raw"
```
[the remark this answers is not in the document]
t-lo: I believe I just used that because that's how we ship `/usr/share/flatcar/etc/` and other Flatcar defaults. This decision was never discussed widely - happy to have that discussion. It's very easy to change this because this file is the only place where that path is generated, filled, and used (for the softlink).

build_library/sysext_prod_builder
```diff
@@ -0,0 +1,154 @@
...
+
+  mkdir -p "${sysext_mountdir}/${name}" \
+           "${sysext_mountdir}/${name}_pkginfo"
+  sudo mount -rt squashfs -o loop,nodev "${sysext_output_dir}/${name}.raw" \
+                                    "${sysext_mountdir}/${name}"
+  sudo mount -rt squashfs -o loop,nodev "${sysext_output_dir}/${name}_pkginfo.raw" \
+                                    "${sysext_mountdir}/${name}_pkginfo"
+
+  sysext_lowerdirs="${sysext_lowerdirs}:${sysext_mountdir}/${name}"
+  sysext_lowerdirs="${sysext_lowerdirs}:${sysext_mountdir}/${name}_pkginfo"
```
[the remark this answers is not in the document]
t-lo: It shouldn't matter _here_ because Gentoo tracks installed packages in separate sub-directories (instead of, say, a single file which is appended to), the respective `..._pkginfo`s  only contain `/var/db/pkgs` sub-directories of packages they ship ([build_sysext#L224](https://github.com/flatcar/scripts/blob/contrib/torcx-deprecation-docker-sysext/build_sysext#L224)), and the sysexts themselves only contain their actual binaries in `/usr`.

Bit I'll for sure make a mental note on this, it's important to know.

build_library/sysext_prod_builder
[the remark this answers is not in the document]
t-lo: Yes, the `trap` with the unmounts is actually the main reason there is a separate file which is run in a sub-bash, instead of adding all this to `build_image_util.sh` (where Krish originally out it).

build_sysext
t-lo: Should be fine here, same reason as for https://github.com/flatcar/scripts/pull/1216#discussion_r1367026876.

build_sysext
t-lo: I think this is fine - it's up to the caller to pass an appropriate `BUILD_DIR` anyway.

build_sysext
t-lo: That would be ... _evil_! Good catch, will move it.

build_sysext
krnowak: I was thinking not in terms of BUILD_DIR, but rather in terms of possible conflicts where basename for two different ppaths will return the same value, thus reusing the same directory for the second ppath.

run_sdk_container
t-lo: Great minds... :smile_cat:

sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
```diff
@@ -37,7 +37,6 @@ src_install() {
 		"${D}"/usr/lib/dracut/modules.d/30ignition/ignition-setup.sh \
 		"${D}"/usr/lib/dracut/modules.d/30ignition/ignition-kargs-helper \
 		"${D}"/usr/lib/dracut/modules.d/30ignition/retry-umount.sh \
-		"${D}"/usr/lib/dracut/modules.d/35torcx/torcx-profile-populate-generator \
```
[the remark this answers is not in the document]
t-lo: Torcx is everywhere. Yes, absolutely, will do. Tracked here: https://github.com/flatcar/Flatcar/issues/1217

sdk_container/src/third_party/coreos-overlay/app-torcx/docker/docker-20.10.ebuild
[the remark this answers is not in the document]
t-lo: In name only; docker is not in sys-apps/torcx anymore. But I'll rename the respective variables to avoid confusion in the future.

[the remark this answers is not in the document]
t-lo: (implying that I'll remove it of course)

ci-automation/ci_automation_common.sh
```diff
@@ -192,9 +192,16 @@ function docker_image_from_buildcache() {
     local id_file_url="https://${BUILDCACHE_SERVER}/containers/${version}/${id_file}"
     local id_file_url_release="https://mirror.release.flatcar-linux.net/containers/${version}/${id_file}"
 
-    if image_exists_locally "${name}" "${version}" ; then
+    local local_image=""
+    if image_exists_locally "/${name}" "${version}" ; then
+        local_image="${name}:${version}"
+    elif image_exists_locally "${CONTAINER_REGISTRY}/${name}" "${version}" ; then
+        local_image="${CONTAINER_REGISTRY}/${name}:${version}"
+    fi
```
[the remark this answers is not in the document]
t-lo: That's a typo - copy+waste from the `elif` case. _Great_ catch Krzesimir!

t-lo:
Restarted CI builds + tests after adding missing bits and pieces for squashfs zstd compression support.

pothos:
I think our Jenkins builders can't mount the resulting squashfs because they don't have the squashfs zstd kernel config set. Using the mksquashfs zstd option has to wait until we either backport the kernel config change to Stable or the new Alpha transitions to Stable.

t-lo:
> I think our Jenkins builders can't mount the resulting squashfs because they don't have the squashfs zstd kernel config set. Using the mksquashfs zstd option has to wait until we either backport the kernel config change to Stable or the new Alpha transitions to Stable.

Maybe there's a way around it, like rebuilding the squashfs kernel module with all options enabled for the Jenkins nodes using the devcontainer. No need to resolve our test infra issues via actual public releases.

pothos:
I think our Jenkins builders can't mount the resulting squashfs because they don't have the squashfs zstd kernel config set. Using the mksquashfs zstd option has to wait until we either backport the kernel config change to Stable or the new Alpha transitions to Stable.

> Maybe there's a way around it, like rebuilding the squashfs kernel module with all options enabled for the Jenkins nodes using the devcontainer. No need to resolve our test infra issues via actual public releases.

With the switch of `/usr` to btrfs we gain much free space. Let's just postpone the squashfs zstd enabling until the Jenkins workers have the kernel features to get this here merged.

t-lo:
> I think our Jenkins builders can't mount the resulting squashfs because they don't have the squashfs zstd kernel config set. Using the mksquashfs zstd option has to wait until we either backport the kernel config change to Stable or the new Alpha transitions to Stable.

But that's what I'm saying - we ship squashfs as a kernel module, it's not compiled into the kernel. So it should be straightforward to rebuild the kmod on our CI servers with zstd enabled, using the devcontainer? It would be a stop-gap solution until we ship zstd support in Beta.

t-lo:
Backported squashfs ZSTD support to current stable, beta, and alpha, and created a helper script to install a systemd service that builds and loads a custom squashfs kernel module on Jenkins nodes. Updated the summary to reflect the new dependencies.

build_library/sysext_prod_builder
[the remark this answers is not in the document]
krnowak: Can we make it `/usr/lib/flatcar/sysext` then?

build_sysext
```diff
@@ -179,13 +219,25 @@ done
...
 if [[ -n "${FLAGS_manglefs_script}" ]]; then
   if [[ ! -x "${FLAGS_manglefs_script}" ]]; then
     die "${FLAGS_manglefs_script} is not executable"
   fi
   "${FLAGS_manglefs_script}" "${BUILD_DIR}/install-root"
 fi
 
+info "Writing ${SYSEXTNAME}_packages.txt"
+ROOT="${BUILD_DIR}/install-root" PORTAGE_CONFIGROOT="${BUILD_DIR}/install-root"\
+      equery --no-color list --format '$cpv::$repo' '*' > "${BUILD_DIR}/${SYSEXTNAME}_packages.txt"
```
krnowak, line 239: Shouldn't this also happen before invoking manglefs?

run_local_tests.sh
```diff
@@ -0,0 +1,108 @@
...
+  local arch="${1}"
+  local parallel="${2}"
+
+  # Read by the mantle container.
+  # The local directory ("pwd") will be mounted to /work/ in the container.
+  cat > sdk_container/.env <<EOF
+export QEMU_IMAGE_NAME="/work/__build__/images/images/${arch}-usr/latest/flatcar_production_image.bin"
+export QEMU_UEFI_BIOS="/work/__build__/images/images/${arch}-usr/latest/flatcar_production_qemu_uefi_efi_code.fd"
+export QEMU_UPDATE_PAYLOAD="/work/__build__/images/images/${arch}-usr/latest/flatcar_test_update.gz"
+export PARALLEL_TESTS=${parallel}
```
krnowak, line 57: The reason why it broke was that my suggestion was wrong, it should have been this:

```suggestion
export QEMU_IMAGE_NAME=/work/__build__/images/images/${arch@Q}-usr/latest/flatcar_production_image.bin
export QEMU_UEFI_BIOS=/work/__build__/images/images/${arch@Q}-usr/latest/flatcar_production_qemu_uefi_efi_code.fd
export QEMU_UPDATE_PAYLOAD=/work/__build__/images/images/${arch@Q}-usr/latest/flatcar_test_update.gz
export PARALLEL_TESTS=${parallel@Q}
```

Though personally I'm usually going the bit longer way and do this:

```bash
local dir="/work/__build__/images/images/${arch}-usr/latest"
local image_name="${dir}/flatcar_production_image.bin"
local uefi_bios="${dir}/flatcar_production_qemu_uefi_efi_code.fd"
local payload="${dir}/flatcar_test_update.gz"
cat > sdk_container/.env <<EOF
export QEMU_IMAGE_NAME=${image_name@Q}
export QEMU_UEFI_BIOS=${uefi_bios@Q}
export QEMU_UPDATE_PAYLOAD=${payload@Q}
…
```

build_sysext
```diff
@@ -27,6 +27,14 @@ DEFINE_string image_builddir '' \
   "Custom directory to build the sysext in. Defaults to a 'sysext' sub-directory of the directory the squashfs base image resides in; '${default_imagedir}/sysext' by default."
 DEFINE_string manglefs_script '' \
   "A path to executable that will customize the rootfs of the sysext image."
+DEFINE_boolean generate_pkginfo "${FLAGS_FALSE}" \
+  "Generate an additional squashfs '<sysext_name>_pkginfo.raw' with portage package meta-information (/var/db ...). Useful for creating sysext dependencies; see 'base_pkginfo' below."
+DEFINE_string base_pkginfo "" \
+  "Colon-separated list of pkginfo squashfs paths / files generated via 'generate_pkginfo' to base this sysext on. The corresponding base sysexts are expected to be merged with the sysext generated."
+DEFINE_string compression "zstd" \
```
pothos, line 34:
```suggestion
DEFINE_string compression "gzip" \
```
It's easier to merge the PR if we switch to `zstd` as default in a follow-up PR that we merge after the next round of releases. Then we don't have to build kernel modules for the Jenkins workers.

pothos: Since the releases already got delayed and there were some concerns regarding release member availability this week, I would say that choosing a fast and easy route is the best.

t-lo: That's OK, Jenkins workers have been updated with a custom kmod already: https://github.com/flatcar/flatcar-linux-infra/pull/268. The kmod will automatically be rebuilt when Jenkins nodes update the OS. The custom kmod service won't run for releases higher than 3760. I've set a reminder for early December to clean up the few artifacts.

I'll start a new build and will update the summary of this PR accordingly.

t-lo:
Addressed Krzesimir's comments, rebased to latest main, restarted Ci tests.

krnowak (approved):

t-lo:
- Github CI is all green
- Jenkins CI is all green
- 2 LGTMs

==> Merging PR. Thank you so much @pothos and @krnowak for your help and support with improving this PR and getting it merged! Also big thanks to @tormath1 for help with the test suite and to @krishjainx for starting this work!

pothos:
Thanks, somehow I assumed the kernel config change would be part of last week's releases and also that we should have this here in Alpha. Eagerly waiting for the next Alpha then^^

PR #1231, opened 2023-10-09
Build docker with go 1.19

jepio:
# Build docker with go 1.19

Docker 20.10.24 expects to be built with go 1.19 and if it isn't certain unix level functionality is broken. We already bumped to building with go 1.19 for alpha-3619.0.0, so this change it present in beta/alpha/main.

Fixes a regression in stable https://github.com/flatcar/Flatcar/issues/1203.

## How to use

## Testing done

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/app-emulation/docker-cli/docker-cli-20.10.24.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/app-emulation/docker-cli/docker-cli-20.10.24.ebuild
@@ -6,7 +6,7 @@ GIT_COMMIT=e78084afe5
 EGO_PN="github.com/docker/cli"
 
 COREOS_GO_PACKAGE="${EGO_PN}"
-COREOS_GO_VERSION="go1.18"
+COREOS_GO_VERSION="go1.19"
 
 inherit bash-completion-r1  golang-vcs-snapshot coreos-go-depend
 
--- a/changelog/bugfixes/2023-10-09-docker-go-1.19.md
+++ b/changelog/bugfixes/2023-10-09-docker-go-1.19.md
@@ -1,0 +1,1 @@
+- Fixed a regression in Docker resulting in file permissions being dropped from exported container images. ([scripts#1231](https://github.com/flatcar/scripts/pull/1231))
--- a/sdk_container/src/third_party/coreos-overlay/app-emulation/docker-proxy/docker-proxy-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/app-emulation/docker-proxy/docker-proxy-9999.ebuild
@@ -5,7 +5,7 @@ EAPI=6
 EGO_PN="github.com/docker/libnetwork"
 
 COREOS_GO_PACKAGE="${EGO_PN}"
-COREOS_GO_VERSION="go1.18"
+COREOS_GO_VERSION="go1.19"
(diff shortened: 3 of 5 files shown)


dongsupark (approved):
Thanks.

Looks good if CI passed. 
Please add changelog as well.

jepio:
Added a changelog and validated that:
* `docker version` reports built with `go 1.19.9`
* the reproducer shows `0/42` (good) instead of `0/0` (bad):
```
docker run --name bug-test alpine ls -l /etc/shadow
docker export bug-test | tar tv etc/shadow
```

jepio:
@dongsupark: LTS will also need this soon?

sdk_container/src/third_party/coreos-overlay/app-emulation/docker-cli/docker-cli-20.10.24.ebuild
```diff
@@ -6,7 +6,7 @@ GIT_COMMIT=e78084afe5
 EGO_PN="github.com/docker/cli"
 
 COREOS_GO_PACKAGE="${EGO_PN}"
-COREOS_GO_VERSION="go1.18"
+COREOS_GO_VERSION="go1.19"
```
pothos, line 9: A fat comment here to always align this with the upstream build system would be nice

pothos: E.g. point to https://github.com/docker/cli/blob/master/Dockerfile#L4

pothos: and https://github.com/moby/moby/blob/master/Dockerfile#L3 etc.

dongsupark:
> @dongsupark: LTS will also need this soon?

No, LTS-2023 (flatcar-3510) still has Docker [20.10.23](https://github.com/moby/moby/releases/tag/v20.10.23), which is still based on Go 1.18.
Only Docker [20.10.24](https://github.com/moby/moby/releases/tag/v20.10.24) or newer has Go [1.19](https://github.com/moby/moby/commit/5091f13a5d027b785084028aedb03beb5b0bd9a6).
No need to backport.

sdk_container/src/third_party/coreos-overlay/app-emulation/docker-cli/docker-cli-20.10.24.ebuild
jepio: I'll add that to main

jepio:
> > @dongsupark: LTS will also need this soon?
> 
> No, LTS-2023 (flatcar-3510) still has Docker [20.10.23](https://github.com/moby/moby/releases/tag/v20.10.23), which is still based on Go 1.18. Only Docker [20.10.24](https://github.com/moby/moby/releases/tag/v20.10.24) or newer has Go [1.19](https://github.com/moby/moby/commit/5091f13a5d027b785084028aedb03beb5b0bd9a6). No need to backport.

Will LTS stay on docker 20.10.23 forever?

dongsupark:
> Will LTS stay on docker 20.10.23 forever?

Usually yes, we try to not touch LTS, unless there is a serious need for another version update in LTS, like a serious bug fix.
What we are regularly doing at the moment in LTS, i.e. updates of Linux Kernel, ca-certificates, mantle containers are already a certain amount of changes.

PR #1238, opened 2023-10-09
coreos-base/misc-files: Make Kubernetes work by default through symlink

pothos:
In the past user had to customize Kubernetes or use a bind mount to make writing the default /usr/libexec/kubernetes/ path work. With systemd-sysext on by default the bind mount doesn't work anymore because it can get lost. A newer workaround is to use a systemd-sysext image that creates a symlink in /usr/libexec/... to redirect to somewhere under /var/.
Instead of relying on workarounds, make Kubernetes work by default on Flatcar by having the symlink be part of the generic image. The target folder will be created through a tmpfiles rule.

(The same paths are used as in https://github.com/flatcar/sysext-bakery/pull/28)
## How to use

Backport to Stable and maybe even LTS?

## Testing done

[Done](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2690/cldsv/):
Checked that the symlink exists under `/usr` and that the target dir is created by a tmpfile rule in `base_image_var.conf` (`base_image_var.conf:d /var/kubernetes/kubelet-plugins/volume/exec 0755 root       root       - -`)

### DIFF
--- a/changelog/changes/2023-10-09-kubernetes-usr-libexec.md
+++ b/changelog/changes/2023-10-09-kubernetes-usr-libexec.md
@@ -1,0 +1,1 @@
+- To make Kubernetes work by default, `/usr/libexec/kubernetes/kubelet-plugins/volume/exec` is now a symlink to the writable folder `/var/kubernetes/kubelet-plugins/volume/exec` ([Flatcar#1193](https://github.com/flatcar/Flatcar/issues/1193))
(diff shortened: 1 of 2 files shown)


krnowak (approved):

pothos:
Backported to Stable, but LTS doesn't have the "misc" ebuild… Should I place it in `coreos-init` instead?

krnowak:
> Backported to Stable, but LTS doesn't have the "misc" ebuild… Should I place it in `coreos-init` instead?

Sure.

PR #1240, opened 2023-10-10
sdk_lib: Workaround docker issue with missing ownership preservation

krnowak:
Just do stuff as root for now.

The workaround can be removed when our builders pick up a stable release with a fix.

CI: (weekly package updates using this commit): http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1103/cldsv/


krnowak:
The workaround seems to be working - the failing sdk container job passed. Needed to restart the follow-up second-phase sdk job, because the worker was terminated or something.

dongsupark (approved):

PR #1272, opened 2023-10-13
update_engine/bootengine: Clean up incomplete move target

pothos:
This pulls in https://github.com/flatcar/update_engine/pull/29 and https://github.com/flatcar/bootengine/pull/76
to fix the problem of not being able to update from ≤ 3619.0.0 on Azure.

## How to use

Backport to Alpha/Beta because this fixes the problem of not being able to update from ≤ 3619.0.0

## Testing done

[Jenkins](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2734/cldsv/)
Provision Flatcar Azure 3619.0.0, then update once, e.g. `sudo flatcar-update -V 3732.0.0`, then update to this built dev image and it should work (`sudo flatcar-update -P flatcar_test_update.gz -E flatcar_test_update-oem-azure.gz -D -V 1.2.3`), while updating to 3745.0.0 or  3732.1.0  (must not the version you are on) would fail and fill the OEM disk.

This also tested the bootengine PR because it will try to move the dev sysext to the OEM partition as active OEM sysext but that fails because the old OEM files only get deleted at the end of the preparations. Moving the inactive prod OEM sysext out would normally be done and could have created enough space but in this case not because the symlink under `/etc` isn't set up and we don't mount the inactive `/usr` to find out what version it is. Hence the moval of the active OEM only happens after the next boot.


pothos:
I didn't bump the ebuild revision because it always creates conflicts when backporting.

krnowak:
> I didn't bump the ebuild revision because it always creates conflicts when backporting.

I know it's a PITA, but we already had weird problems due to not bumping the revisions, so please do bump them. I'm wondering though if by dropping the `9999` version and having actual file instead of a symlink, it would make it easier for git to figure out how to do the right thing during backporting.

krnowak:
For now, backported to alpha only. Probably no point in backporting to beta, as we will do a new beta release from current alpha release this week, so 3732 becomes dead.

pothos:
Thanks!

PR #1277, opened 2023-10-16
backport curl CVE to flatcar-3510 (LTS-2023)

sayanchowdhury:
backport curl CVE to flatcar-3510 (LTS-2023)

CVEs for which patches are added
- CVE-2023-38545
- CVE-2023-38546


krnowak:
There was no need in moving the package to coreos-overlay - there is a user-patches setup in this version of LTS. All that was needed was to add those patches to `sdk_container/src/third_party/coreos-overlay/coreos/user-patches/net-misc/curl/` directory.

PR #1278, opened 2023-10-16
backport curl CVE to flatcar-3033 (LTS-2022)

sayanchowdhury:
CVEs for which patches are added

CVE-2023-38545
CVE-2023-38546


krnowak:
I see that you did a copy of net-misc/curl into coreos-overlay instead of moving it from portage-stable to coreos-overlay. Would be better to have just one copy of the package in either one to avoid confusion.

krnowak (approved):

PR #1280, opened 2023-10-17
overlay cloudsigma: Set tty to RAW

dosomder:
# cloudsigma: Set TTY to RAW mode

According to cloudsigma's documentation at https://docs.cloudsigma.com/en/latest/server_context.html#setting-up-the-virtual-serial-port the TTY used to fetch the userdata / metadata should be in mode RAW.

With the flatcar beta 3732.1.0, fetching the data did not work anymore (the returned JSON was missing some bytes), without setting this mode correctly.

## How to use

- Machine on cloudsigma with flatcar image
- Check output of `oem-cloudinit.service` if SSH keys and hostname were set -> it will fail
- Set mode to RAW
- Run service again -> it should work

## Testing done

We tested it on our machines with a custom built image.


changelog/bugfixes/2023-10-17-cloudsigma-tty-raw.md
```diff
@@ -0,0 +1 @@
+- TTY to fetch server_context is set to RAW mode before running cloudinit ([scripts#1280](https://github.com/flatcar/scripts/pull/1280))
```
krnowak, line 1:
```suggestion
- Set TTY used for fetching server_context to RAW mode before running cloudinit on cloudsigma ([scripts#1280](https://github.com/flatcar/scripts/pull/1280))
```

krnowak (changes requested):
Hi,

Thanks for the PR!

It looks to me more like a bug somewhere inside coreos-cloudinit. It uses github.com/cloudsigma/cepgo to fetch metadata. This in turn uses github.com/tarm/goserial, so I'd say that the issue is somewhere there. But looking at the packages, they seem to be old/unmaintained/deprecated. Meh.

I suppose we could go with this workaround. Could you then bump the revision of the oem-cloudsigma ebuild from 2 to 3, please? So it's `oem-cloudsigma-0.0.3-r3.ebuild`. Thanks!

dosomder:
@krnowak To bump the revision, do I simply rename the file? And you probably mean to r4 since it is currently at r3
https://github.com/flatcar/scripts/blob/main/sdk_container/src/third_party/coreos-overlay/coreos-base/oem-cloudsigma/oem-cloudsigma-0.0.3-r3.ebuild

krnowak:
> @krnowak To bump the revision, do I simply rename the file?

Yep.

> And you probably mean to r4 since it is currently at r3 https://github.com/flatcar/scripts/blob/main/sdk_container/src/third_party/coreos-overlay/coreos-base/oem-cloudsigma/oem-cloudsigma-0.0.3-r3.ebuild

Ah, right. Yeah, should be bumped to -r4. Sorry for the confusion.

dosomder:
Thanks for the fast review @krnowak !

PR #1297, opened 2023-10-23
add zstd support to squashfs (release 3602)

t-lo:


krnowak:
Hm, should we add a changelog entry for this change? (Same of other sibling PRs.)

t-lo:
> Hm, should we add a changelog entry for this change? (Same of other sibling PRs.)

Good point, added.

krnowak (approved):

t-lo:
Tests are green, merging.

PR #1313, opened 2023-10-26
overlay update-ssh-keys: update openssh-keys to 0.6.2

dongsupark:
Update openssh-keys to 0.6.2 to avoid relying on a custom branch.
Also remove a custom patch from the ebuild.
Regenerate CRATES by running cargo ebuild.

Pulls in https://github.com/flatcar/update-ssh-keys/pull/14.

## Testing done

TBD

- Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) (not needed)


pothos:
See comment in https://github.com/flatcar/update-ssh-keys/pull/12#issuecomment-1808828487
This is a regression and we should either revert this or fix the problem before branching the next Alpha.

PR #1315, opened 2023-10-27
Upgrade Containerd in main from 1.7.7 to 1.7.8

flatcar-infra:
Subject says it all.


t-lo:
Closing and re-starting the github action since we had significant changes in the main branch that affect containerd.

PR #1319, opened 2023-11-01
Monthly GLSA metadata 2023-11-01

flatcar-infra:
Updated GLSA metadata


t-lo (approved):
Test are all green; CTRL+F "GLSA" in the build_image CI step shows that the image is not affected by any GLSAs.

We should add a test that checks for GLSAs; created https://github.com/flatcar/Flatcar/issues/1231to track this.

PR #1325, opened 2023-11-02
Upgrade Linux Firmware in main from 20230919 to 20231030

flatcar-infra:
Subject says it all.


t-lo (approved):
LGTM, all tests are green.

PR #1328, opened 2023-11-03
Upgrade Linux Kernel for flatcar-3760 from 6.1.60 to 6.1.61

flatcar-infra:
Subject says it all.


t-lo (approved):
Tests are all green, LGTM.

PR #1391, opened 2023-11-14
Use OpenStack image for Brightbox

pothos:
- Use OpenStack image for Brightbox
    
    The special Brightbox image uses the OpenStack userdata in Ignition but
    lacked Afterburn usage. It actually works to use the OpenStack image and
    directly which also enables Afterburn, thus we can drop the special
    image.
    Don't build a special image for Brightbox but recommend to use OpenStack
    images directly. A symlink is added to help with the download of
    hardcoded user scripts.
- ci-automation/vms: Provide OpenStack image without external compression
    
    For Brightbox we can use the OpenStack image but the import only works
    with unpacked images. After we enabled internal qcow2 compression the
    .gz or .bz2 external compression doesn't provide any benefits and makes
    the import more complicated.
    Provide the OpenStack image without external compression in addition.
    The other files are kept for now but we could also delete them if we
    announce this in advance.

## Testing done

The openstack image was tested on brightbox by @tormath1 

Checked bincache results for [jenkins job](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2913/cldsv/) and now we have a symlink but regular digest files.


ci-automation/vms.sh
```diff
@@ -150,6 +152,7 @@ function _vm_build_impl() {
 
     create_digests "${SIGNER}" "images/latest/"*
     sign_artifacts "${SIGNER}" "images/latest/"*
+    ( cd images/latest ; ln -s flatcar_production_openstack_image.img.bz2 flatcar_production_brightbox_image.img.bz2 )
```
krnowak, line 155: I'm wondering if this should be done before digests and signatures are created, so we still have digests and signatures for the brightbox image.

pothos: Thanks, pushed

pothos:
Because this relies on the openstack oem rework I don't think we can easily backport it fully but we could, e.g., backport the uncompressed OpenStack image to have some consistency.

krnowak (approved):

pothos:
Backport done down to 3510 (LTS) by not removing the full ebuild but only the brightbox flag.

PR #1396, opened 2023-11-16
Upgrade Linux Firmware in main from 20231030 to 20231111

flatcar-infra:
Subject says it all.


t-lo (approved):
All tests passed, LGTM.

PR #1157, opened 2023-09-20
overlay profiles: Fix a couple of issues with SLSA provenance stuff

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/2543/cldsv

- Update to slsa 1.0. This is only partially done, as we still need to provide a proper build type. Maybe we could reuse the Github Actions Workflow (https://slsa-framework.github.io/github-actions-buildtypes/workflow/v1).

- Stop using portageq - its use in ebuilds is banned, so eventually it would stop working. Replace it with our hack.

- Stop trying to get a commit hash of coreos-overlay or portage-stable as if they were submodules. This setup is long gone, so a commit hash of toplevel scripts repo is enough.

- Use zstd for compressing generated JSON files.

--

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [x] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

PR #1399, opened 2023-11-17
profiles: slsa: Remove quadratic complexity in SRC_URI iteration

jepio:
# speed up slsa provenance generation for some packages

SLSA provenance generation iterates over $A (which is a subset of $SRC_URI) and for each of those tries to find a match in $SRC_URI. That's quadratic complexity, and the performance impact is bad because we shell out to a helper utility (basename) for every entry. This is leading to long stalls when generating SLSA for packages with long distfile lists, like go and rust packages. Iterate over SRC_URI once and create a dictionary to speed up subsequent lookups. dev-db/etcdctl is a good candidate for testing.

## How to use

```
time emerge-amd64-usr etcdtctl
```

## Testing done

Before: >10min on my laptop
After: 1min on my laptop (includes build time)


krnowak (approved):
Looks good. You probably could drop the `found` variable now, but I have updated the whole SLSA thing in https://github.com/flatcar/scripts/pull/1157, so I can do it in the PR.

jepio:
@krnowak lets cherry-pick this into your PR then

krnowak:
Cherry picked to #1157.

PR #1400, opened 2023-11-17
coreos-base/coreos-init: Make sshkeys.service more robust

pothos:
This pulls in
https://github.com/flatcar/init/pull/112
to only run coreos-metadata-sshkeys@core.service when not masked and also retry on failure.

## How to use

Backport to Beta (via `flatcar-3745-backport`)

## Testing done

See linked PR


pothos:
Ok, did it in a separate commit for easier backporting

PR #1401, opened 2023-11-17
build_library/sysext_prod_builder: Add debug output

pothos:
We see occasional failures due to a read-only filesystem on GitHub Action runners:
  mkdir: cannot create directory ‘/home/sdk/trunk/src/scripts/artifacts/amd64-usr/developer-3790.0.0+nightly-20231116-2100-5-g49eb1a4c07-a1/rootfs/usr/share/flatcar/sysext’: Read-only file system

Add commands for a debug output.


t-lo (approved):
LGTM. Tested `build_image` locally and in the non-error case the change doesn't break anything. Feel free to merge w/o waiting for the CI build to conclude.

PR #1424, opened 2023-11-27
image_changes: support multiple LTS

tormath1:
When we support two LTS, the oldest one is compared to the newer one (e.g lts-2022 is compared to lts-2023).

We now read the 'lts-info' file to find the right version to compare (e.g lts-2022 is compared to the 'current-2022')

---

To be backported to flatcar-3033 / flatcar-3510

### DIFF
--- a/ci-automation/image_changes.sh
+++ b/ci-automation/image_changes.sh
@@ -51,6 +51,23 @@ function _image_changes_impl() {
     source sdk_container/.repo/manifests/version.txt
     local vernum="${FLATCAR_VERSION}"
 
+    MAJOR_B=$(echo "${FLATCAR_VERSION}" | cut -d . -f 1)
+
+    SUFFIX=
+    if [ "${channel}" = "lts" ]; then
+            curl -fsSLO --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 'https://lts.release.flatcar-linux.net/lts-info'
+            while read -r LINE; do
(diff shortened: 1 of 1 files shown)


ci-automation/image_changes.sh
```diff
@@ -51,6 +51,20 @@ function _image_changes_impl() {
...
     local vernum="${FLATCAR_VERSION}"
 
+    MAJOR_B=$(echo "${FLATCAR_VERSION}" | cut -d . -f 1)
+
+    curl -fsSLO --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 'https://lts.release.flatcar-linux.net/lts-info'
+    while read -r LINE; do
+        # each line is major:year:(supported|unsupported)
+        TUPLE=(${LINE//:/ })
+        MAJOR="${TUPLE[0]}"
+        if [[ "${MAJOR_B}" = "${MAJOR}" ]]; then
```
pothos, line 61:
```suggestion
        if [[ "${MAJOR_B}" = "${MAJOR}" ]] && [[ "${channel}" = "lts" ]]; then
```

krnowak: Or:

```bash
SUFFIX=
if [[ ${channel} = 'lts' ]]; then
    curl …
    while read …; do …; done <lts-info
    rm -f lts-info
fi
```

krnowak (approved):

PR #1434, opened 2023-11-29
Print debug output on any error

krnowak:
This is a follow up to Kai's PR (#1401). It will print debugging output together with a backtrace now. In order to do it, the build_sysext is now using the strict mode too, which enable the backtraces to be printed.


common.sh
```diff
@@ -164,9 +164,40 @@ die_notrace() {
...
+      error "${DIE_PREFIX}!!!!!!!!!!!!!!!!!!!!!!!!!"
+      error "${DIE_PREFIX}!! BEGIN DEBUG OUTPUT: !!"
+      error "${DIE_PREFIX}!!!!!!!!!!!!!!!!!!!!!!!!!"
+      error
+      error "${DIE_PREFIX}== MOUNT =="
+      error "${DIE_PREFIX}==========="
+      error_command_output "${DIE_PREFIX}" mount
+      error
+      error "${DIE_PREFIX}== DF =="
+      error "${DIE_PREFIX}==========="
```
krnowak, line 177:
```suggestion
      error "${DIE_PREFIX}========"
```

I'll fix it after github action is done.

dongsupark (approved):

PR #1442, opened 2023-11-29
build_image_util: Store path list for recreating tmpfiles rules

pothos:
The removal of files in the overlay present in the lowerdir creates
    whiteout entries that mask the lowerdir entries. For those files that
    have a tmpfile rule for creation, a reboot would cause the file to be
    created in the upperdir, meaning this file is not updated from the
    lowerdir when it changes. In addition we have filtered out some tmpfile
    rules that caused upcopies (symlinks and directories) which meant that
    removing the /etc/resolv.conf symlink didn't bring it back after reboot.
    To make files from the lowerdir show up if they have a tmpfile rule that
    normally would recreate them we keep a list of whiteout entries that we
    clean up on boot. This also prevents freezing files because
    systemd-tmpfiles does not need to recreate them in the upperdir.

## How to use

Backport to Stable

## Testing done

see https://github.com/flatcar/bootengine/pull/79


build_library/build_image_util.sh
```diff
@@ -737,9 +737,16 @@ EOF
...
   # like + or = which explicitly recreate files.
-  sudo sed -i '/^[CLd]-*!*-*[ \t]*\/etc\//d' "${root_fs_dir}"/usr/lib/tmpfiles.d/*
+  # But before filtering, first store rules that would recreate missing files
+  # to /usr/share/flatcar/etc-no-whiteouts so that we can ensure that
+  # no overlayfs whiteouts exist for these files (example: /etc/resolv.conf).
+  # These rules are combined with the + modifier in addition.
+  # Other rules like w, e, x, do not create files that don't exist.
+  # Note: '-' must come first in the modifier pattern.
+  grep -Ph '^[fcCdDLvqQpb][-=~^!+]*[ \t]*/etc' "${root_fs_dir}"/usr/lib/tmpfiles.d/* | grep -oP '/etc[^ \t]*' | sudo_clobber "${root_fs_dir}"/usr/share/flatcar/etc-no-whiteouts
+  sudo sed -i '/^[CdDL][-=~^!]*[ \t]*\/etc\//d' "${root_fs_dir}"/usr/lib/tmpfiles.d/*
```
krnowak, line 749: Is this resolv.conf thing new in some way? I kicked off an image and grepping for any /etc entry in tmpfiles shows no results for resolv.conf:

```
core@localhost ~ $ grep VERSION= /etc/os-release 
VERSION=3802.0.0+weekly-package-updates-2
core@localhost ~ $ grep -Ph '^.[-=~^!+]*[ \t]*/etc' /usr/lib/tmpfiles.d/*
d   /etc                              -   -   -   -   -
Z   /etc/passwd                       -   -   -   -   -
Z   /etc/shadow                       -   -   -   -   -
Z   /etc/group                        -   -   -   -   -
Z   /etc/gshadow                      -   -   -   -   -
L+  /etc/ld.so.conf     -   -   -   -   ../usr/lib/ld.so.conf
L+ /etc/mtab - - - - ../proc/self/mounts
f       /etc/flatcar/update.conf       0644    -    -    -       -
f^ /etc/motd.d/50-provision.conf - - - - login.motd
f^ /etc/issue.d/50-provision.conf - - - - login.issue
f^ /etc/hosts - - - - network.hosts
f /etc/environment		0755 root root	-	-
r /etc/dbus-1/system.d/UpdateEngine.conf - - - 0 -
```

krnowak (approved):

pothos:
Tested with https://github.com/flatcar/mantle/pull/488

PR #1445, opened 2023-11-30
gcp: fixed OS Login enabling

tormath1:
Noticed in logs:
```
Nov 20 19:28:42.573152 enable-oslogin[1122]: /etc/ssh/sshd_config does not include configuration snippets in /etc/ssh/sshd_config.d. Not enabling OS Login
```

The check: https://github.com/flatcar/scripts/blob/0d6e3aaca9e40106763979b39a5dd09cfae9dd28/sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/files/bin/enable-oslogin#L21-L23 is not passing because of the duplicate `/`:
```
$ sudo cat /etc/ssh/sshd_config | grep Include
# Make sure that all Match options are below this Include!
Include "//etc/ssh/sshd_config.d/*.conf"
```

## Testing done

- [x] CI: http://192.168.42.7:8080/job/container/job/packages_all_arches/3029/cldsv/

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
+++ b/sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/files/units/oem-gce-enable-oslogin.service
@@ -8,6 +8,7 @@ Before=sshd.socket
 
 [Service]
 Type=oneshot
+RemainAfterExit=true
 
 ExecStart=/oem/bin/enable-oslogin
 


sdk_container/src/third_party/coreos-overlay/net-misc/openssh/openssh-9.5_p1-r1.ebuild
```diff
@@ -229,6 +229,7 @@ src_test() {
 }
 
 insert_include() {
+	# note: includedir must be an absolute path (i.e, starting with /)
```
pothos, line 232: Maybe add a check instead of a note

sdk_container/src/third_party/coreos-overlay/net-misc/openssh/openssh-9.5_p1-r1.ebuild
```diff
@@ -232,6 +232,10 @@ insert_include() {
 	local src_config="${1}" options="${2}" includedir="${3}"
 	local name copy regexp_options regexp lineno comment_options
 
+	if [[ "${includedir}" =~ ^/.* ]]; then
```
pothos, line 235:
```suggestion
	if [[ ! "${includedir}" =~ ^/.* ]]; then
```

pothos: or `if ! [[` - both should work

pothos: Do you want to commit the suggestion?

pothos:
A backport to Beta would be good

PR #1467, opened 2023-12-04
overlay coreos-base/afterburn: update to 5.5.0

dongsupark:
Update `coreos-base/afterburn` to the latest release [5.5.0](https://github.com/coreos/afterburn/releases/tag/v5.5.0).

Adjust Flatcar patches for 5.5.0, it is necessary to add again the `hostname` crate, which is not included in 5.5 any more by default.

Remove unnecessary patches for 5.5.0.
  * `0003-encode-information-for-systemd-networkd-wait-online.patch`: it is already merged in upstream, https://github.com/coreos/afterburn/commit/d2cc340038e8cc1625d58802be644a7cb3eea1ae.
  * `0004-lto-thin.patch`: it is not needed any more, since upstream does not enable `lto = true` any more in https://github.com/coreos/afterburn/commit/464c7f9f0a8fe7b7196a7724df183a6b13c819c9.

Needs to be cherry-picked to Beta, `flatcar-3760`.

Fixes https://github.com/flatcar/Flatcar/issues/1049.

## Testing done

TBD


sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/0003-vmware-arm64-unsupported.patch
```diff
@@ -0,0 +1,41 @@
...
+diff --git a/src/providers/vmware/unsupported.rs b/src/providers/vmware/unsupported.rs
+index 99dace52..c4c04f20 100644
+--- a/src/providers/vmware/unsupported.rs
++++ b/src/providers/vmware/unsupported.rs
+@@ -7,4 +7,8 @@ impl VmwareProvider {
+     pub fn try_new() -> Result<Self> {
+         bail!("unsupported architecture");
+     }
++
++    pub fn parse_netplan_config(&self) -> Result<Option<String>> {
```
pothos, line 35: Can you do an upstream PR for that?

dongsupark: https://github.com/coreos/afterburn/pull/1022

pothos: Thanks

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/0004-lto-thin.patch
```diff
@@ -1,29 +0,0 @@
...
-diff --git a/Cargo.toml b/Cargo.toml
-index e5b3dc30..46bbda69 100644
---- a/Cargo.toml
-+++ b/Cargo.toml
-@@ -22,7 +22,7 @@ name = "afterburn"
- path = "src/main.rs"
- 
- [profile.release]
--lto = true
-+lto = false
```
pothos, line 23: To reduce the binary size we can set `codegen-units = 1` here, which is similar to thin LTO.

Full LTO would be ideal for binary size but since it's currently unclear if that works well with arm64, let's look at that later.

dongsupark: `codegen-units = 1` looks like a good idea. It reduces binary size of amd64 at least by 50MiB.

pothos: I think we also strip debug symbols but I need to find out where because the Cargo.toml doesn't do it. To get the same results directly we can set `strip = true` here, too.

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/0004-cargo-codegen-units-release-profile.patch
```diff
@@ -0,0 +1,28 @@
...
+ 1 file changed, 2 insertions(+)
+
+diff --git a/Cargo.toml b/Cargo.toml
+index da74d909..2b12776d 100644
+--- a/Cargo.toml
++++ b/Cargo.toml
+@@ -32,6 +32,8 @@ path = "src/main.rs"
+ [profile.release]
+ # We assume we're being delivered via e.g. RPM which supports split debuginfo
+ debug = true
```
pothos, line 20: I think we can remove the `debug = true` because we throw the result away. Setting `strip = true`  would also make it clearer what the end result on the image will look like.

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/0004-lto-thin.patch
dongsupark: Done.

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/0004-cargo-codegen-units-release-profile.patch
dongsupark: Done.

Yes, both adding `strip = true` and removing `debug = true` help reduce the binary size.
After being installed, it looks anyway nearly the same as before.

pothos (approved):
Thanks!

dongsupark:
CI passed for all cloud providers.

PR #1505, opened 2023-12-18
sdk_lib/sdk_init_selfcontained.sh: fix access rights

t-lo:
This change fixes access rights / file ownership changes in the self-contained SDK init script.

### DIFF
--- a/sdk_lib/sdk_init_selfcontained.sh
+++ b/sdk_lib/sdk_init_selfcontained.sh
@@ -26,8 +26,12 @@ function clone_version() {
 }
 # --
 
+# Nothing mounted on /mnt/host in self-contained mode
+sudo chown -R sdk:sdk /home/sdk /mnt/host/source
+
 version="$(source /mnt/host/source/.repo/manifests/version.txt; echo $FLATCAR_VERSION)"
 
-mkdir -p /home/sdk/trunk/src/third_party/
+rmdir /mnt/host/source/src/third_party
+ln -s /mnt/host/source/src/scripts/sdk_container/src/third_party /mnt/host/source/src/
 
 clone_version scripts /home/sdk/trunk/src/scripts "$version"


sdk_lib/sdk_init_selfcontained.sh
```diff
@@ -26,8 +26,12 @@ function clone_version() {
...
 # --
 
+# Nothing mounted on /mnt/host in self-contained mode
+sudo chown -R sdk:sdk /home/sdk /mnt/host/source
+
 version="$(source /mnt/host/source/.repo/manifests/version.txt; echo $FLATCAR_VERSION)"
 
-mkdir -p /home/sdk/trunk/src/third_party/
+rmdir /mnt/host/source/src/third_party
+ln -s /mnt/host/source/src/scripts/sdk_container/src/third_party /mnt/host/source/src/
```
pothos, line 35: That looks a bit strange because the link points to something under it, maybe one of them should be /home/sdk/trunk/src?

t-lo: No, this is correct - it's a bit convoluted: `/home/sdk/trunk` is a soft-link to `/mnt/host/source` because some parts of the SDK point to files below that directory (most notably the `repo.conf/coreos.conf` ones).

Then again, `/home/sdk/trunk/src/scripts` is the checkout point of the scripts repo, and `/home/sdk/trunk/src/scripts/sdk_container/src/third_party` has the ebuild repos which are expected at `/mnt/host/source/src/third_party`.

We do something similar in `run_sdk_container` to make `third_party` appear in ``/mnt/host/source/src/`: https://github.com/flatcar/scripts/blob/main/run_sdk_container#L141

pothos: Ah, I got confused by the trailing slash, so this is creating a symlink names `third_party` in `src`

pothos (approved):

PR #1542, opened 2024-01-04
Upgrade dev-lang/rust and virtual/rust in main from 1.74.1 to 1.75.0

flatcar-infra:
Subject says it all.


pothos:
Started a two-phase SDK build in [jenkins](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1246/cldsv/)

pothos:
I pushed patches for build failures. They are not part of the "apply flatcar patches" commit because I expect this to be only needed until Gentoo updates to 1.75.

pothos:
There is an alternative fix in https://github.com/rust-lang/rust/pull/119582 - will need to test if that works, too, but the current one looks good so far for our usage.

dongsupark (approved):
Both upstream approaches are fine.

PR #1571, opened 2024-01-15
coreos-cloudinit: fix hostname reset bug

jepio:
# coreos-cloudinit: fix hostname reset bug

When coreos-cloudinit runs on a platform where afterburn sets the hostname it would reset the hostname to localhost. Fix that bug, by checking that the hostname is non-empty before applying. Pulls in https://github.com/flatcar/coreos-cloudinit/pull/25.

See: https://github.com/flatcar/Flatcar/issues/1262

## How to use

Boot built image on Azure without providing any userdata to see if the hostname gets reset.

## Testing done


pothos (approved):

jepio:
http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3274/cldsv/

Will merge/backport as soon as tests finish.

jepio:
tested - this now works

PR #1572, opened 2024-01-15
app-emulation/amazon-ssm-agent: use upstream makefile

tormath1:
By maintaining our own build logic, we missed the fact that the agent was now split in two parts:
* amazon-ssm-agent in `./core`
* ssm-agent-worker in `./agent`

With the runtimeconfig passed from the `core` to the `agent`.

We were currently shipping `ssm-agent-worker` as the `amazon-ssm-agent` hence some errors we had.

```
amazon-ssm-agent[8929]: 2023-12-28 15:29:32 WARN [ssm-agent-worker] failed to read runtime config 'identity_config.json': open /var/lib/amazon/ssm/runtimeconfig/identity_config.json: no such file or directory
```

- [ ] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3275/cldsv/

Locally tested but a last test with the image from the CI will be good.

Closes: https://github.com/flatcar/Flatcar/issues/1307 https://github.com/aws/amazon-ssm-agent/issues/554


sdk_container/src/third_party/coreos-overlay/app-emulation/amazon-ssm-agent/amazon-ssm-agent-3.2.985.0-r3.ebuild
```diff
@@ -0,0 +1,49 @@
...
+	ln -s ${PWD}/vendor/src/* ${PWD}/vendor/
+}
+
+src_compile() {
+	go_export
+
+	# set agent release version
+	BRAZIL_PACKAGE_VERSION=${PV} ${EGO} run ./agent/version/versiongenerator/version-gen.go
+	# build all the tools
+	if [[ "${ARCH}" != "amd64" ]]; then
```
jepio, line 34:
```suggestion
	if [[ "${ARCH}" == "arm64" ]]; then
```

sdk_container/src/third_party/coreos-overlay/app-emulation/amazon-ssm-agent/amazon-ssm-agent-3.2.985.0-r2.ebuild
```diff
@@ -1,57 +0,0 @@
...
-src_prepare() {
-	default
-	ln -s ${PWD}/vendor/src/* ${PWD}/vendor/
-}
-
-src_compile() {
-	go_export
-
-	# this is replication of commands from the vendor makefile
-	# but without network activity during build phase
```
pothos, line 32: The reason for the duplication was the network access, is this not the case anymore or if it is the case, are we fine with network access? We could also patch the Makefile instead of duplicating the commands.

jepio (approved):

PR #1579, opened 2024-01-16
changelog: Update a security entry for openssl

krnowak:
The release with this changelog was yet done, so we can safely update it.

Closes https://github.com/flatcar/Flatcar/issues/1320.


krnowak:
Actually, 3.0.12 is affected by this CVE too, so closing the PR.

PR #1580, opened 2024-01-16
coreos-base/(ue-rs|update_engine): Control extracted OEM payload name

pothos:
This pulls in https://github.com/flatcar/ue-rs/pull/49 and https://github.com/flatcar/update_engine/pull/36 to fix the handling of OEM update payloads in a Nebraska response with self-hosted packages. Before this fix the processing failed and had to rely on the fallback download from the release server.

## How to use

Backport to new Alpha

## Testing done

Tested that updating to the built payload works when using this patch for flatcar-update and running `sudo ./flatcar-update -V 1.2.3 -D -P flatcar_test_update.gz -E extrafile-amd64-3760.1.1-oem-qemu.gz`:
```
diff --git a/bin/flatcar-update b/bin/flatcar-update
index d1ca303..42bf086 100755
--- a/bin/flatcar-update
+++ b/bin/flatcar-update
@@ -59,9 +59,6 @@ while true; do
     if [ "$1" = "" ]; then
       echo "Error: --extension must not have an empty value" > /dev/stderr ; exit 1
     fi
-    if [[ ! "$(basename -- "$1")" =~ ^(flatcar|oem).*gz$ ]]; then
-      echo "Error: --extension expects paths to files named oem-OEMID.gz or flatcar-NAME.gz (with possible 'flatcar_test_update-' prefix), found: $1" > /dev/stderr ; exit 1
-    fi
     EXTENSIONS+=("$1")
     ;;
   -L|--listen-port-1)
@@ -121,8 +118,6 @@ OEMID=$({ grep -m 1 -o "^ID=.*" /usr/share/oem/oem-release 2> /dev/null || true
 if [ "${OEMID}" != "" ] && { [ -e "/usr/share/flatcar/oems/${OEMID}" ] || [ -e "/usr/share/oem/sysext/active-oem-${OEMID}" ]; }; then
   if [ "$PAYLOAD" = "" ]; then
     EXTENSIONS+=("/var/tmp/flatcar-update/oem-${OEMID}.gz")
-  elif ! echo " ${EXTENSIONS[*]} " | grep -q -P "[ /](flatcar_test_update-)?oem-${OEMID}.gz "; then # Surrounded with space to only match base name
-    echo "Error: system requires '${OEMID}' OEM extension but not passed in --extension" > /dev/stderr ; exit 1
   fi
 fi
 for NAME in $(grep -h -o '^[^#]*' /etc/flatcar/enabled-sysext.conf /usr/share/flatcar/enabled-sysext.conf 2> /dev/null | grep -v -x -f <(grep '^-' /etc/flatcar/enabled-sysext.conf 2> /dev/null | cut -d - -f 2-) | grep -v -P '^(-).*'); do
```

Jenkins [build](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3292/cldsv/)


jepio:
We need this in stable too, i'm going to prepare a PR with a backport.

jepio:
Stable isn't using ue-rs yet.

@pothos DO you prefer to update stable to ue-rs or https://github.com/flatcar/update_engine/pull/39?

PR #1608, opened 2024-01-26
ci-automation/garbage_collect.sh: Add min age, remove orphan directories

t-lo:
This change improves the build cache garbage collector to remove orphaned artifact directories - i.e. directories to which no version tag exists in the scripts repo.

SDK containers built by Github actions (using update_sdk_container) are igored by this change because these are handled in a separate garbage collection script.

Also, a new command line parameter has been added to remove artifacts older than the specified number of days (defaulting to 14):
- If neither number of builds nor max age is specified, the script defaults to 50 builds to keep, and a max age of 14 days. The max age overrides the number of builds to keep, so more than 50 builds may be kept.
- If only the number of builds to keep is specified, the max age is set to "0" (i.e. today).
- If both are specified, max age again overrides number of builds to keep.

Lastly, the change updates `ci-automation/garbage_collect_github_ci_sdk.sh` to also feature a min_age parameter and passes the parameter from ``garbage_collect.sh` to `ci-automation/garbage_collect_github_ci_sdk.sh`.

## How to use

```bash
bash -c "source ci-automation/garbage_collect.sh ; DRY_RUN=y garbage_collect ; "
bash -c "source ci-automation/garbage_collect.sh ; DRY_RUN=y garbage_collect 45; "
bash -c "source ci-automation/garbage_collect.sh ; DRY_RUN=y garbage_collect 1 12; "

bash -c "source ci-automation/garbage_collect_github_ci_sdk.sh ; DRY_RUN=y garbage_collect_github_ci 1 236; "
bash -c "source ci-automation/garbage_collect_github_ci_sdk.sh ; DRY_RUN=y garbage_collect_github_ci 10 200; "
```

## Testing done

Ran the above 5 commands and attached logs:
* remove orphans - [remove_orphans_run.txt](https://github.com/flatcar/scripts/files/14065698/remove_orphans_run.txt) (GH clean-up removed from log for readability)
* remove all but last 45 builds (orphan purging and GH clean-up removed from log for readability) - [remove_all_but_45_run.txt](https://github.com/flatcar/scripts/files/14065726/remove_all_but_45_run.txt)
* remove all but 1 builds but minimum age set to 12 days (orphan purging and GH clean-up removed from log for readability) -  [remove_all_older_than_12_days_run.txt](https://github.com/flatcar/scripts/files/14065727/remove_all_older_than_12_days_run.txt)
* Github actions cleanup of artifacts older than 236 days - [remove_gh_ci_sdk_older_236_run.txt](https://github.com/flatcar/scripts/files/14066268/remove_gh_ci_sdk_older_236_run.txt)
* Github actions cleanup of artifacts older than 200 days, keep at least 10 - [remove_gh_ci_sdk_older_200_keep_10_run.txt](https://github.com/flatcar/scripts/files/14066273/remove_gh_ci_sdk_older_200_keep_10_run.txt)

## Follow-up tasks
- [ ] Integrate max age parameter with jenkins-os

### DIFF
--- a/ci-automation/garbage_collect.sh
+++ b/ci-automation/garbage_collect.sh
@@ -8,10 +8,27 @@
 #
 # garbage_collect() should be called after sourcing.
 #
+# The garbage collector will remove artifacts of all NON-RELEASE versions from the build cache
+# which BOTH
+#   * exceed the number of builds to keep (defaults to 50)
+#   AND
+#   * are older than the minimum purge age (14 days by default)
+#
+# Note that the min age threshold can lead to MORE than 50 builds being kept if this script
+#   is run with its default values.
+#
+# Additionally, the garbage collector will remove all artifacts and directories that do not have
+# a version TAG in the scripts repository.
+#
 #  OPTIONAL INPUT
 #  - Number of (recent) versions to keep. Defaults to 50.
+#           Explicitly setting this value will reset the minimum age (see below) to 0 days.
+#  - Minimum age of version tag to be purged, in days. Defaults to 14.
 #  - PURGE_VERSIONS (Env variable). Space-separated list of versions to purge
 #            instead of all but the 50 most recent ones.
+#            Setting this will IGNORE minimum age and number of versions to keep.
+#            NOTE that only dev versions (not official releases) may be specified.
+#            This is to prevent accidental deletion of official release tags from the git repo.
 #  - DRY_RUN (Env variable). Set to "y" to just list what would be done but not
 #            actually purge anything.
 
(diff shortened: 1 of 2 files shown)


ci-automation/garbage_collect.sh
```diff
@@ -269,6 +271,6 @@ function _garbage_collect_impl() {
     echo
 
     source ci-automation/garbage_collect_github_ci_sdk.sh
-    garbage_collect_github_ci
+    garbage_collect_github_ci 1 "${min_age_days}"
```
krnowak, line 274: Just one SDK build to keep? Asking to make sure, especially that the default was 20. I do realize that probably more will be kept until they are older than 14 days or so.

ci-automation/garbage_collect.sh
```diff
@@ -38,26 +57,60 @@ function garbage_collect() {
...
+    fi
+    echo
 
     if [ -z "${purge_versions}" ] ; then
-        keep="$((keep + 1))" # for tail -n+...
+        # Generate a list "<timestamp> | <tagname>" from all repo tags that look like dev versions
+        local versions_detected="$(git tag -l --sort=-committerdate \
+                                          --format="%(creatordate:format:%Y-%m-%d) | %(refname:strip=2)" \
+                | grep -E '.*\| (main|alpha|beta|stable|lts)-[0-9]+\.[0-9]+\.[0-9]+-.*' \
+                | grep -vE '(-pro)$')"
```
krnowak, line 89: Why excluding the `-pro` build tag? Some artifact from the past?

ci-automation/garbage_collect.sh
```diff
@@ -38,26 +57,60 @@ function garbage_collect() {
...
+                                    next
+
+                                print $3
+                                }')"
     else
-        # make sure we only accept dev versions
+        # User-provided version list, make sure we only accept dev versions
         purge_versions="$(echo "${purge_versions}" | sed 's/ /\n/g' \
                             | grep -E '(main|alpha|beta|stable|lts)-[0-9]+\.[0-9]+\.[0-9]+\-.*' \
                             | grep -vE '(-pro)$')"
```
krnowak, line 112: Same here for excluding `-pro` from purges.

ci-automation/garbage_collect.sh
```diff
@@ -90,7 +143,7 @@ function _garbage_collect_impl() {
         local os_docker_vernum="$(vernum_to_docker_image_version "${FLATCAR_VERSION}")"
 
         # Remove container image tarballs and SDK tarball (if applicable)
-        #
+        # Keep in sync with "orphaned direcrories" clean-up below.
```
krnowak, line 146:
```suggestion
        # Keep in sync with "orphaned directories" clean-up below.
```

ci-automation/garbage_collect.sh
t-lo: The "1" just means that at least one is kept; for everything else the 14-day limit will apply.

ci-automation/garbage_collect.sh
t-lo: Yes, to prevent deleting the release tags of the `-pro` versions.

ci-automation/garbage_collect.sh
t-lo: We would need to rework tag removal and exclude any pattern that matches an official release version. Note that versions that look like official releases are skipped entirely - the garbage collector only cares about dev builds and nightlies. (Releases are supposed to be cleaned up manually, after pushing release binaries to the official mirrors.)

ci-automation/garbage_collect.sh
krnowak: This is a bit too all-catching as it will prevent GC of any dev build that ends with -pro (so a build tag like `weekly-updates-like-a-pro` won't be GC'd), but OTOH we tend not to put -pro into our build tags, so I suppose it's fine.

krnowak (approved):

ci-automation/garbage_collect.sh
t-lo: Good point, the match could be more precise.

PR #1616, opened 2024-01-30
ci-automation: improve bincache garbage collection, cover cached release artifacts

t-lo:
This PR extends the garbage collector for the build cache server to remove cached release artifacts. Release artifacts are copied to the official mirrors and do not need to remain on the build cache after a release was published.

By default, the 10 latest releases per channel (including LTS and previous LTS) are kept.

Also excluded from garbage collection are:
 - Emerging new major releases (i.e. major number larger than the latest Alpha release)
 - channel progressions (major number exists in the lists of releases to keep but minor is bigger than any release)
 - patch releases (major and minor exist in list of releases to keep but patch level is newer than in any release)
 - SDKs (tarballs and containers) of any release in the list of releases to keep; i.e. the SDK in <MAJOR>.0.0 for any release to keep.

Furthermore, this change unsets `POSIXLY_CORRECT` for all `awk` calls that use `gensub`. This fixes an issue with restricted environments like e.g. the systemd-run env we use in our ci-automation.

# How to use
```bash
bash -c "source ci-automation/garbage_collect_releases.sh ; DRY_RUN=y garbage_collect_releases;"
```

# Testing done
Ran the above command and captured the output:  [remove_releases.txt](https://github.com/flatcar/scripts/files/14099392/remove_releases.txt). Reviewed the output; looks sane.

### DIFF
--- a/ci-automation/garbage_collect_releases.sh
+++ b/ci-automation/garbage_collect_releases.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright (c) 2021 The Flatcar Maintainers.
+# Copyright (c) 2024 The Flatcar Maintainers.
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
@@ -10,7 +10,7 @@
 #
 #  OPTIONAL INPUT
 #  - Number releases to keep per channel. Defaults to 10.
-#  - Number of LTS channels to keep. Defaults to 3 (i.e. the current and the previous (deprecated) LTS).
+#  - Number of LTS channels to keep. Defaults to 2 (i.e. the current and the previous (deprecated) LTS).
 #  - DRY_RUN (Env variable). Set to "y" to just list what would be done but not
 #            actually purge anything.
 
@@ -46,8 +46,7 @@ function _garbage_collect_releases_impl() {
     local keep_versions
     mapfile -t keep_versions < <(unset POSIXLY_CORRECT; \
         curl -s "${RELEASES_JSON_FEED}" \
-        | jq '[keys]' \
-        | sed -n 's/.*"\([0-9]\+\.[0-9]\+\.[0-9]\+\)".*/\1/p' \
+        | jq -r 'keys_unsorted | .[] | match("[0-9]+\\.[0-9]+\\.[0-9]+") | .string' \
         | sort -Vr \
         | awk -v keep="${keep_per_chan}" -v lts="${keep_lts_releases}" '
             {
--- a/ci-automation/garbage_collect_releases.sh
+++ b/ci-automation/garbage_collect_releases.sh
@@ -98,58 +98,65 @@ function _garbage_collect_releases_impl() {
                                  "ls -1 ${BUILDCACHE_PATH_PREFIX}/${dir} | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$'"); do
             local o_fullpath="${fullpath}/${version}"
 
-            # skip if version is marked for keeping OR if it's a new release about to be published
+            # Go through all deletion candidate versions in directory
+            #   skip if candidate version is marked for keeping OR if it's a new release about to be published
+            #   or if it ships an SDK used by a version in the keep list.
             if printf "%s\n" "${keep_versions[@]}" \
-               | { unset POSIXLY_CORRECT ; awk -v version="${version}" -v path="${dir}" '
+               | { unset POSIXLY_CORRECT ; awk -v candidate_version="${version}" -v path="${dir}" '
                 BEGIN {
-                    vmajor = gensub("([0-9]+)\\.[0-9]+\\.[0-9]+","\\1","g", version) + 0
-                    vminor = gensub("[0-9]+\\.([0-9]+)\\.[0-9]+","\\1","g", version) + 0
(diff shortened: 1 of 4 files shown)


ci-automation/garbage_collect_releases.sh
```diff
@@ -0,0 +1,180 @@
...
+                    minor = gensub("[0-9]+\\.([0-9]+)\\.[0-9]+","\\1","g") + 0
+                    patch = gensub("[0-9]+\\.[0-9]+\\.([0-9]+)","\\1","g") + 0
+
+                    if (   ((path == "sdk/amd64") || (path == "containers")) \
+                        && (vmajor == major) && (vminor == 0) && (vpatch == 0) ) {
+                        print ""
+                        print "## Skipping " version " in " path " because it contains the SDK for release " $0 " in keep list."
+                        ret = 0
+                        exit
+                    }
```
krnowak, line 125: That's tricky. We seldom had a case where we had to rebuild SDK for some patch release, so SDK ended up having a version different than `<MAJOR>.0.0`.

t-lo:
(I'm not around for more than a week, feel free to amend and merge this PR as you see fit)

ci-automation/garbage_collect_releases.sh
```diff
@@ -0,0 +1,180 @@
...
+    echo
+    echo "Number of versions to keep per channel: '${keep_per_chan}'"
+    echo "Number of LTS major releases to keep: '${keep_lts_releases}'"
+    echo
+
+    source ci-automation/ci_automation_common.sh
+    local sshcmd="$(gen_sshcmd)"
+
+    local keep_versions
+    mapfile -t keep_versions < <(unset POSIXLY_CORRECT; \
```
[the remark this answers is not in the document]
t-lo: Good catch, I'll rewrite.

ci-automation/garbage_collect_releases.sh
t-lo: Yeah, I agree, the check is too rigorous. We should skip all versions that have a path below `/sdk/smd64` and that match a major version on the keep list.

t-lo:
Addressed all comments, ran another dry run using
```bash
bash -c "source ci-automation/garbage_collect_releases.sh ; DRY_RUN=y garbage_collect_releases;"
```

Output: 
[out.txt](https://github.com/flatcar/scripts/files/14241513/out.txt)

ci-automation/garbage_collect_releases.sh
```diff
@@ -0,0 +1,183 @@
...
+
+    echo
+    echo "Number of versions to keep per channel: '${keep_per_chan}'"
+    echo "Number of LTS major releases to keep: '${keep_lts_releases}'"
+    echo
+
+    source ci-automation/ci_automation_common.sh
+    local sshcmd="$(gen_sshcmd)"
+
+    local keep="$(
```
krnowak, line 46: Quotes in assignment are not necessary.
```suggestion
    local keep=$( 
```

ci-automation/garbage_collect_releases.sh
```diff
@@ -0,0 +1,183 @@
...
+                            print version
+                        chan_count[chan_num] = chan_count[chan_num] + 1
+                    } else {
+                        if (    (chan_count["lts"][major] < keep) \
+                             && (length(chan_count["lts"]) <= lts) )
+                            print version
+                        chan_count["lts"][major] = chan_count["lts"][major] + 1
+                    }
+                } '
+    )"
```
krnowak, line 70:
```suggestion
    )
```

ci-automation/garbage_collect_releases.sh
```diff
@@ -0,0 +1,183 @@
...
+        echo
+        echo "## Processing '${fullpath}'"
+        echo "---------------------------"
+        for version in $($sshcmd "${BUILDCACHE_USER}@${BUILDCACHE_SERVER}" \
+                                 "ls -1 ${BUILDCACHE_PATH_PREFIX}/${dir} | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$'"); do
+            local o_fullpath="${fullpath}/${version}"
+
+            # skip if version is marked for keeping OR if it's a new release about to be published
+            if printf "%s\n" "${keep_versions[@]}" \
+               | { unset POSIXLY_CORRECT ; awk -v version="${version}" -v path="${dir}" '
```
krnowak, line 103: Can I ask you to rename some of the variables? It would make it easier to track which version, major, minor and patch is which. What I mean is something like renaming `version` to `bincache_version`, `v{major,minor,patch}` to `bincache_{major,minor,patch}`. And then possibly having an assignment like `keep_version = $0` and variables `keep_{major,minor,patch}`. `major_alpha` probably could be `keep_major_alpha`.

krnowak (approved):

ci-automation/garbage_collect_releases.sh
t-lo: Yes, good point. Tbh I got confused myself when updating the PR after one week of vacation :sweat_smile:

t-lo:
Addressed all comments, ran another dry run using
```bash

bash -c "source ci-automation/garbage_collect_releases.sh ; DRY_RUN=y garbage_collect_releases;"
```

Output: [out.txt](https://github.com/flatcar/scripts/files/14284257/out.txt)

PR #1628, opened 2024-02-06
docker 24.0.9, runc 1.1.12, containerd 1.7.13 for beta

krnowak:


jepio (approved):

pothos:
Thanks!
(I realize that containerd is in the coreos-overlay because it uses a version that wasn't available at that time but in general it looks like we could move it back to portage-stable, or?)

krnowak:
> Thanks! (I realize that containerd is in the coreos-overlay because it uses a version that wasn't available at that time but in general it looks like we could move it back to portage-stable, or? - Not for the release branch but in main)

Yeah, we could move it to portage-stable now. We could also drop app-containers/docker-proxy from portage-stable as it is not pulled in any more.

krnowak:
Any clue what the failure in `extra-test.[first_dual].cl.update.oem` is about? Same happened in https://github.com/flatcar/scripts/pull/1629.

pothos:
Looks unrelated because it downloads something from bincache instead of an artifact of this PR
```
 first_dual: ++ curl --head -o /dev/null -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 https://bincache.flatcar-linux.net/images/amd64/3815.1.0/flatcar_test_update-oem-azure.gz
```

pothos:
> Looks unrelated because it downloads something from bincache instead of an artifact of this PR
> 
> ```
>  first_dual: ++ curl --head -o /dev/null -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 https://bincache.flatcar-linux.net/images/amd64/3815.1.0/flatcar_test_update-oem-azure.gz
> ```

Maybe the `.github/` workflows are out of sync with `main`.

krnowak:
> > Looks unrelated because it downloads something from bincache instead of an artifact of this PR
> > ```
> >  first_dual: ++ curl --head -o /dev/null -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 https://bincache.flatcar-linux.net/images/amd64/3815.1.0/flatcar_test_update-oem-azure.gz
> > ```
> 
> Maybe the `.github/` workflows are out of sync with `main`.

Alright, I'll ignore it for now.

Kicked off the CI again, because containerd insisted on pulling in the old version of runc. Same thing done for alpha and main.

krnowak:
> > Kicked off the CI again, because containerd insisted on pulling in the old version of runc. Same thing done for alpha and main.
> 
> For a more simple approach to problems like this I would always delete the old ebuilds to prevent them from being used accidentally.

That would fail the build, as containerd had a dependency on any revision of runc 1.1.9. Admittedly that would be more noticable than just looking at the image reports.

PR #1629, opened 2024-02-06
docker 24.0.9, runc 1.1.12, containerd 1.7.13 for alpha

krnowak:


pothos:
On a first look the `extra-test.[first_dual].cl.update.oem` failure seems to be related to the `.github` or ci integration folder being out of sync or something because it downloads things from bincache. If there were regressions we see them in Jenkins. I think it's safe to ignore for now.

pothos (approved):

PR #1630, opened 2024-02-06
docker 24.0.9, runc 1.1.12 for main

krnowak:


pothos:
Do you think the test failure is due to the update to 25.0.2?  (by the way, the PR title needs correction)
If that's the case, for a quick solution I would suggest to stick to 24.

krnowak:
> Do you think the test failure is due to the update to 25.0.2? (by the way, the PR title needs correction) If that's the case, for a quick solution I would suggest to stick to 24.

Yep, just found out that [device-mapper storage driver was removed in v25](https://github.com/docker/cli/blob/v25.0.0/docs/deprecated.md#device-mapper-storage-driver). Will stick to v24 for now. But we will probably need to think about a possible migration path of docker storage from devicemapper to overlay2 maybe using a postinst script.

pothos:
> > Do you think the test failure is due to the update to 25.0.2? (by the way, the PR title needs correction) If that's the case, for a quick solution I would suggest to stick to 24.
> 
> Yep, just found out that [device-mapper storage driver was removed in v25](https://github.com/docker/cli/blob/v25.0.0/docs/deprecated.md#device-mapper-storage-driver). Will stick to v24 for now. But we will probably need to think about a possible migration path of docker storage from devicemapper to overlay2 maybe using a postinst script.

Oh, yes, a similar topic was the Docker btrfs storage backend: https://www.flatcar.org/releases#release-3794.0.0

pothos:
What was the revert commit about? I think we can drop it because the nightly would overwrite again, or?

pothos (approved):

krnowak:
> What was the revert commit about? I think we can drop it because the nightly would overwrite again, or?

Ah, I'll drop it. It was there, because nightly SDK was not yet available so CI was failing. With the revert, CI pulled a day older SDK.

PR #1631, opened 2024-02-06
changelog: remove cloudsigma custom cloudinit service configuration

daMupfel:
# CloudSigma - Remove unnecessary cloudinit service override configuration

This is related to our first attempt to fix CloudSigma cloudinit issues here: https://github.com/flatcar/scripts/pull/1280

CloudSigma uses a serial port to transfer metadata into a running instance. Because the service was overwritten with the configuration "restart" it was possible that the previous instance got interrupted. This left the serial connection in an non-deterministic state as the next user would get output from the previous request. The service in general was not needed as the cloud oem is already set to cloudsigma within grub.config.

## How to use

- Use adapted image in CloudSigma
- Attach ssh private keys to the instance
- Check that private keys are loaded on boot

## Testing done

Since I lack the understanding of building flatcar, I modified the existing oem image directly and uploaded it to CloudSigma for testing.


changelog/bugfixes/2024-02-06-cloudsigma-remove-custom-cloudinit-service-config.md
```diff
@@ -0,0 +1 @@
+- Remove custom CloudSigma coreos-cloudinit service configuration since it will be called with the cloudsigma oem anyway. The restart of the service can also cause the serial port to be stuck in an non deterministic state which breaks future runs.
```
krnowak, line 1:
```suggestion
- Removed custom CloudSigma coreos-cloudinit service configuration since it will be called with the cloudsigma oem anyway. The restart of the service can also cause the serial port to be stuck in an nondeterministic state which breaks future runs.
```

krnowak (commented):
Hi, thanks for the PR!

We don't have a CI running on cloudsigma instance, so I'll take your word that this change is correct. :) Had some nitpicks on the changelog wording.

daMupfel:
Hi,

I adapted the wording according to your suggestion.

Regarding testing, I tested this multiple times with our custom image (fresh vm and disk iamge), and it so far seems to work. 
Since this (if I am correct with my observation) is a timing related issue, it is sometimes hard to reliably reproduce.

I came to the conclusion that reading the serial port was interrupted, because I saw the coreos-cloudinit service being stopped/restarted while watching the boot mechanism through VNC. 

If you then later read the remaining output from the serial port, you can observe that it returns values from a previous request.

If I understand the mechanism correctly than I don't think this change should break anything as the service is called anyway just with the configured oem in the grub config (which is the same) as the manually provided one. The other line regarding setting the tty mode was introduced by us and sadly did not fix the issue.

Regards,

David

krnowak (approved):
Looks good, thanks!

PR #1654, opened 2024-02-13
bootengine: fix kmod-static-nodes

ader1990:
Update the bootengine commit id to use the fix from: https://github.com/flatcar/bootengine/pull/85

Fixes: https://github.com/flatcar/Flatcar/issues/1336

## Testing done

### DIFF
--- a/changelog/bugfixes/2024-02-13-fix-kmod-static-nodes-creation.md
+++ b/changelog/bugfixes/2024-02-13-fix-kmod-static-nodes-creation.md
@@ -1,0 +1,2 @@
+- Resolved kmod static nodes creation in bootengine ([bootengine#85](https://github.com/flatcar/bootengine/pull/85))
+- Fixes kubevirt vm creation by ensuring that /dev/vhost-net exists ([Flatcar#1336](https://github.com/flatcar/Flatcar/issues/1336))
--- a/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="https://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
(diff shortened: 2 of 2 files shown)


pothos (commented):
A changelog entry in `changelog/bugfixes/2024-02-13-TOPIC.md` would be good with something like `- Resolved problem causing fallout ([bootengine#85](https://github.com/flatcar/bootengine/pull/85))`

ader1990:
Note that this issue is present also on Flatcar versions up to 3510 and maybe older versions (I could track it down up to 3510 using a Hyper-V vm. 3374 did not properly boot because of ignition/cloud-init implementation).

changelog/bugfixes/2024-02-13-fix-kmod-static-nodes-creation.md
```diff
@@ -0,0 +1 @@
+- Resolved kmod static nodes creation in bootengine ([bootengine#85](https://github.com/flatcar/bootengine/pull/85))
```
pothos, line 1: Maybe mention kubevirt here for users to understand?

ader1990: Added a new line below mentioning kubevirt, also added the same line into the commit message for easier search

pothos (approved):

changelog/bugfixes/2024-02-13-fix-kmod-static-nodes-creation.md
pothos: Thanks, just FYI: The order may change in the release notes but I think it's ok in this case (I normally do only a single long line for that reason)

pothos:
We can do a backport by picking the bootengine commit into backport branches like `flatcar-3510-backport` and then using the latest commit of these branches in the ebuild file when backporting this change.

ader1990:
> We can do a backport by picking the bootengine commit into backport branches like `flatcar-3510-backport` and then using the latest commit of these branches in the ebuild file when backporting this change.

I suggest to have this change tested better (a few test iterations of the HEAD) before backporting it to the beta / stable branches. For the moment, I am the only person affected by the kubevirt issue.

Thanks.

ader1990:
Note that this PR has not been backported yet and only Flatcar version >= 3913 has the fix.

PR #1655, opened 2024-02-13
sysext: add built-in Incus sysext image

tormath1:
In this PR, we add Incus support as opt-out built-in sysext.

Note for reviewers:
* Most of this PR is adding new packages into portage-stable
* One commit must be upstreamed: `app-containers/incus: fix cross-compilation issue` (https://github.com/gentoo/gentoo/pull/42095)
* With this sysext, we're adding new groups and users - I tried to leverage `userdb` to dynamically load and unload those groups when we load and unload the sysext image (note: it seems a soft reboot might still be necessary)
* I have one concern for the `static-libs` useflag enabled for `sys-libs/libcap` as it seems we don't use this flag for the generic image but for the SDK - that said, the CI output does not show any change regarding the libcap. 
* documentation: https://github.com/flatcar/flatcar-website/pull/423

## How to use

(This is optional to use ZFS)

```yaml
---
variant: flatcar
version: 1.1.0
storage:
  files:
    - path: /etc/flatcar/enabled-sysext.conf
      contents:
        inline: |
          incus
          zfs
  links:
    - path: /etc/extensions/docker-flatcar.raw
      target: /dev/null
      overwrite: true
    - path: /etc/extensions/containerd-flatcar.raw
      target: /dev/null
      overwrite: true
```

Once on the instance, you can use a preseed or follow the interactive `incus admin init`:
```bash
core@localhost ~ $ systemctl status incus
● incus.service - Incus - main daemon
     Loaded: loaded (/usr/lib/systemd/system/incus.service; indirect; preset: disabled)
     Active: active (running) since Thu 2025-05-15 07:54:30 UTC; 40s ago
...
core@localhost ~ $ incus admin init
core@localhost ~ $ incus launch images:ubuntu/22.04 ubuntu-01
Launching ubuntu-01
core@localhost ~ $ incus list
+-----------+---------+----------------------+------+-----------+-----------+
|   NAME    |  STATE  |         IPV4         | IPV6 |   TYPE    | SNAPSHOTS |
+-----------+---------+----------------------+------+-----------+-----------+
| ubuntu-01 | RUNNING | 10.126.29.162 (eth0) |      | CONTAINER | 0         |
+-----------+---------+----------------------+------+-----------+-----------+
core@localhost ~ $ incus exec ubuntu-01 cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
core@localhost ~ $ incus storage list
+---------+--------+-------------+---------+---------+
|  NAME   | DRIVER | DESCRIPTION | USED BY |  STATE  |
+---------+--------+-------------+---------+---------+
| default | zfs    |             | 3       | CREATED |
+---------+--------+-------------+---------+---------+
core@localhost ~ $ zpool list
NAME      SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
default  4.50G   264M  4.24G        -         -     1%     5%  1.00x    ONLINE  -
```

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5981/cldsv/
- [x] Add new packages to `.github/workflows/portage-stable-packages-list`
- [x] rebase on https://github.com/flatcar/scripts/pull/1742
- [x] review downstream patch (tc-is-cross-compiler) + wait for Go1.21 upgrade (https://github.com/gentoo/gentoo/pull/36323)
- [x] Inspect USE flags

Closes: https://github.com/flatcar/Flatcar/issues/1319

### DIFF
--- a/changelog/changes/2025-05-14-incus.md
+++ b/changelog/changes/2025-05-14-incus.md
@@ -1,0 +1,1 @@
+- Provided an Incus Flatcar extension as optional systemd-sysext image with the release. Write 'incus' to `/etc/flatcar/enabled-sysext.conf` through Ignition and the sysext will be installed during provisioning. ([scripts#1655](https://github.com/flatcar/scripts/pull/1655))


ci-automation/base_sysexts.sh
```diff
@@ -8,4 +8,5 @@ fi
 ciabs_base_sysexts=(
     'containerd-flatcar:app-containers/containerd'
     'docker-flatcar:app-containers/docker&app-containers/docker-cli'
+    'incus-flatcar:app-containers/incus'
```
pothos, line 11: I think base_sysext is not the right place, or? We would want this to be built after the prod image and not be part of it. I would extend `build_image` to call `build_sysext` from a new if-block after `if [[ "${PROD_TAR}" -eq 1 ]]; then` controlled by an argument to build sysexts or not (default could be yes?).
The built sysext images needs to be treated like an OEM sysext, to get uploaded and signed the same way. The naming scheme is `flatcar-NAME.raw`

sdk_container/src/third_party/portage-stable/app-containers/incus/incus-0.6-r1.ebuild
```diff
@@ -124,7 +124,9 @@ src_configure() { :; }
 
 src_compile() {
 	export GOPATH="${S}/_dist"
-	export CGO_LDFLAGS_ALLOW="-Wl,-z,now"
+	export CGO_LDFLAGS_ALLOW="-Wl,-z,now,-L,-rpath-link"
+	export CGO_CFLAGS="${CGO_CFLAGS} -I${ESYSROOT}/usr/include"
+	export CGO_LDFLAGS="${CGO_LDFLAGS} -L${ESYSROOT}/usr/lib64 -Wl,-rpath-link,${ESYSROOT}/usr/lib64"
```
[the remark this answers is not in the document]
krnowak: Maybe gate it by some check if we are cross-compiling (`tc-is-cross-compiler` in toolchain-funcs.eclass)? I suppose that the typical, native compilation doesn't have that problem.

pothos:
> Same as podman, we need to generate subuid / subgid

Would be good to use the same approach as done there to make sure that the extensions don't conflict.

sdk_container/src/third_party/coreos-overlay/coreos-devel/board-packages/board-packages-0.0.1.ebuild
krnowak, line 1: I think a revbump of the symlink is missing.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/app-containers/incus
```diff
@@ -0,0 +1,3 @@
+cros_pre_src_compile_pkgconfig() {
```
krnowak, line 1: This looks like a fix candidate for upstream, no?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```diff
@@ -101,3 +101,9 @@ sys-apps/azure-vm-utils
 =sys-fs/fuse-3.17.2 ~arm64
 =sys-libs/libsemanage-3.7 ~arm64
 =sys-process/audit-4.0.2-r1 ~arm64
+=app-containers/incus-6.0.3-r2 ~arm64
+=app-containers/lxc-6.0.3 ~arm64
+=dev-util/xdelta-3.0.11-r1 ~arm64
+=dev-libs/raft-0.22.1 ~arm64
+=dev-libs/cowsql-1.15.7 ~arm64
+=sys-fs/lxcfs-6.0.3-r1 ~arm64
```
krnowak, line 109: Can I ask to keep those entries sorted within a file?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.use
```diff
@@ -176,3 +176,19 @@ sys-apps/systemd -gcrypt
 
 # Make dracut module available for initrd build.
 sys-apps/azure-vm-utils dracut
+
+# Enable SELinux for lxc
+app-containers/lxc selinux
+
+# required by app-containers/incus-6.0.3-r1::portage-stable
+# required by incus (argument)
+>=dev-libs/raft-0.22.1 lz4
```
krnowak, line 185: Could this be

```suggestion
dev-libs/raft lz4
```

?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.use
```diff
@@ -176,3 +176,19 @@ sys-apps/systemd -gcrypt
...
+
+# required by app-containers/incus-6.0.3-r1::portage-stable
+# required by incus (argument)
+>=dev-util/xdelta-3.0.11-r1 lzma
```
krnowak, line 189: Could this be

```suggestion
dev-util/xdelta lzma
```

?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.use
```diff
@@ -176,3 +176,19 @@ sys-apps/systemd -gcrypt
...
+
+# required by app-containers/lxc-6.0.3::portage-stable[tools,caps]
+# required by app-containers/incus-6.0.3-r1::portage-stable
+# required by incus (argument)
+>=sys-libs/libcap-2.71 static-libs
```
krnowak, line 194: Could this be

```suggestion
sys-libs/libcap static-libs
```

?

sdk_container/src/third_party/coreos-overlay/coreos/config/env/acct-group/incus
```diff
@@ -0,0 +1,8 @@
+cros_post_src_install_add_userdb_record(){
```
krnowak, line 1: Could you add comments what this hook is doing? Thanks. (Same for `acct-group/incus-admin` and `acct-group/lxc` hooks too, please.)

krnowak (approved):

PR #1661, opened 2024-02-13
core_sign_update: Work with a mapped pcscd socket

pothos:
The sdk user in the container might rejected by the pcscd on the host. Work around that by running the openssl command as root.

## Testing done

Used for current release


krnowak:
Or maybe the SDK user should be added to the pcscd group.

krnowak (approved):

PR #1665, opened 2024-02-15
sys-kernel/dracut: Sync from Gentoo and apply Flatcar patches

ader1990:
sys-kernel/dracut needs to be upgraded to a version that supports systemd-executor binary.
See: https://github.com/dracutdevs/dracut/pull/2535/commits/62fd8ebe4776dbb60ca04a865c2d160e45d65e91

Without the systemd-executor binary, the upgrade to systemd 255 is not possible,
as it leads to a failure during the initramfs stage - `executable not found systemd-executor`.


krnowak:
Thanks! Could you please drop all the unused ebuilds and entries from Manifest as a part of the "sync with Gentoo" commit? It's a hindrance to find out which one is actually modified by us.

ader1990:
> Thanks! Could you please drop all the unused ebuilds and entries from Manifest as a part of the "sync with Gentoo" commit? It's a hindrance to find out which one is actually modified by us.

sure, will do.

ader1990:
Closing this PR for the moment, as https://github.com/flatcar/scripts/pull/1667 should be enough for systemd 255 upgrade. After https://github.com/flatcar/scripts/pull/1667 is merged and systemd is upgraded, this PR can be reopened.

PR #1667, opened 2024-02-16
sys-kernel/dracut: add systemd-executor binary

ader1990:
Without the systemd-executor binary, the upgrade to systemd 255 is not possible,
as it leads to a failure during the initramfs stage - executable not found systemd-executor.

Upgrade of dracut will be left for the future: https://github.com/flatcar/scripts/pull/1665, as dracut has to be upgraded after systemd 255 has been already upgraded, because of these two commits that will break bootengine and the CI:

https://github.com/dracutdevs/dracut/commit/7528d84de84d9c1fb7d5f54712c692600e21b044
https://github.com/dracutdevs/dracut/commit/5912f4fbc036cc36b9507c16dddef1ded1556572


pothos (approved):

ader1990:
Let's keep this PR in draft state until the systemd 255 PR builds and runs correctly (I will rebase it on top of this one).

ader1990:
PR is ready to merge. The systemd 255 still has some issues with dracut-cmdline-ask.service and vconsole-setup, which are unrelated to this PR. A dracut version upgrade is still needed, but will be done is a separate PR.

PR #1679, opened 2024-02-19
sys-apps/systemd: upgrade from version 252 to version 255

ader1990:
Upgrades systemd 252 to systemd 255.

This is WoIP branch, the following issues are still to be resolved:

  * during initrd stage - dracut-cmdline-ask.service fails because systemd-vconsole-setup.service fails (/usr/bin/loadkeys fails) - fixed by adding the dracut internationalization module
  * check if there are some unnecessary code changes for the kbd module - resolved by sticking with the i18n dracut module - fixed by adding a stripped down patch to dracut i18n module
 
Fixes: https://github.com/flatcar/Flatcar/issues/1269
Requires: https://github.com/flatcar/bootengine/pull/87
Requires a mantle fix or systemd-resolved MulticastDNS disabled: https://bugs.launchpad.net/cloud-images/+bug/2038894/comments/4. Fixed by disabling MulticastDNS.


ader1990:
Required to be merged first: https://github.com/flatcar/bootengine/pull/87.
@pothos suggested to add a new dracut module to only install the required binary loadkeys, and not all the i18n dracut module.

ader1990:
> Failed test case ` cl.network.listeners` needs to be fixed in mantle https://github.com/flatcar/mantle/blob/bd2e8e962592dac502d46d998808fd50b60365d2/kola/tests/misc/network.go#L182, as systemd-resolved now also listens on UDP port 5353. If we do not want this, we can remove the listener: https://bugs.launchpad.net/cloud-images/+bug/2038894/comments/4

https://man.archlinux.org/man/resolved.conf.5#OPTIONS
```
MulticastDNS=

Takes a boolean argument or "resolve". Controls Multicast DNS support (RFC 6762[2]) on the local host.
If true, enables full Multicast DNS responder and resolver support.
 If false, disables both. If set to "resolve", only resolution support is enabled, but responding is disabled.
 Note that [systemd-networkd.service(8)](https://man.archlinux.org/man/systemd-networkd.service.8.en) 
also maintains per-link Multicast DNS settings. Multicast DNS will be enabled on a link only if the per-link 
and the global setting is on.
Added in version 234.
```

jepio:
> Note on the size increase: amd64-generic-image .zip increased by 5MB and vmlinuz increased by 3MB.

Thats a lot for vmlinuz - what exactly is responsible for the size increase in vmlinuz?

jepio:
> now also listens on UDP port 5353.
> If we do not want this, we can remove the listener

I would disable the listener. 99% of Flatcar users have no use for multicast dns, and homelabbers can enable it manually.

ader1990:
> > now also listens on UDP port 5353.
> > If we do not want this, we can remove the listener
> 
> I would disable the listener. 99% of Flatcar users have no use for multicast dns, and homelabbers can enable it manually.

https://github.com/flatcar/scripts/pull/1679/commits/37746a40b0794ef5769ea0be4039258a4672147c -> this should disable it, tests are running.

ader1990:
The Flatcar image built by https://github.com/flatcar/scripts/actions/runs/7970403527/job/21757956307?pr=1679 was tested on arm64 using https://github.com/cloudbase/BMK/tree/flatcar_sysext, the deployment was succesful:

* k8s cluster deployed using k8s sysext on two Ampere nodes
* cillium
* rook / ceph (2 node ceph cluster)
* kubevirt (tested with Fedora VMs using ceph pvc)

ader1990:
> > Note on the size increase: amd64-generic-image .zip increased by 5MB and vmlinuz increased by 3MB.
> 
> Thats a lot for vmlinuz - what exactly is responsible for the size increase in vmlinuz?

Hello @jepio,

According to my comparison from here: https://github.com/flatcar/scripts/pull/1684#issuecomment-1957376741, systemd and kbd/i8in dracut module are the responsible. Makes sense, as the newer version added more systemd binaries / services in the initrd (like systemd-executor, systemd-vconsole-setup).

The vmlinuz from https://github.com/flatcar/scripts/actions/runs/7970403527/job/21757956101?pr=1679 is 57236 KB vs 56868 KB the i18n one, which is only 0.5 MB more.

Basically 1.5 MB more for the systemd upgrade 252 -> 255 (vanilla one is 55860 KB).

This size increase can be reduced by adding a dracut special module for the loadkeys binary instead of installing the dracut i18n module (with the caveat that the new module might introduce technical debt for the future ). @jepio @pothos should I try with a dracut special module especially for this?

Thank you.

ader1990:
> > > Note on the size increase: amd64-generic-image .zip increased by 5MB and vmlinuz increased by 3MB.
> > 
> > 
> > Thats a lot for vmlinuz - what exactly is responsible for the size increase in vmlinuz?
> 
> Hello @jepio,
> 
> According to my comparison from here: [#1684 (comment)](https://github.com/flatcar/scripts/pull/1684#issuecomment-1957376741), systemd and kbd/i8in dracut module are the responsible. Makes sense, as the newer version added more systemd binaries / services in the initrd (like systemd-executor, systemd-vconsole-setup).
> 
> The vmlinuz from https://github.com/flatcar/scripts/actions/runs/7970403527/job/21757956101?pr=1679 is 57236 KB vs 56868 KB the i18n one, which is only 0.5 MB more.
> 
> Basically 1.5 MB more for the systemd upgrade 252 -> 255 (vanilla one is 55860 KB).
> 
> This size increase can be reduced by adding a dracut special module for the loadkeys binary instead of installing the dracut i18n module (with the caveat that the new module might introduce technical debt for the future ). @jepio @pothos should I try with a dracut special module especially for this?
> 
> Thank you.

Please see https://github.com/flatcar/scripts/pull/1684#issuecomment-1963603717. I suggest to go with the upstream tested versions of systemd + dracut implementation in the case of systemd-vconsole-setup. I am afraid that if I continue to keep fixing the stripped down implementation, I will reach to a functional implementation that is the same as the upstream one.

pothos:
> Please see [#1684 (comment)](https://github.com/flatcar/scripts/pull/1684#issuecomment-1963603717). I suggest to go with the upstream tested versions of systemd + dracut implementation in the case of systemd-vconsole-setup. I am afraid that if I continue to keep fixing the stripped down implementation, I will reach to a functional implementation that is the same as the upstream one.

Makes sense - I guess we look for other tricks to reduce the size. E.g., we could remove unused kernel modules or we could compress the kernel more (there are new options in 6.x).

ader1990:
> > Please see [#1684 (comment)](https://github.com/flatcar/scripts/pull/1684#issuecomment-1963603717). I suggest to go with the upstream tested versions of systemd + dracut implementation in the case of systemd-vconsole-setup. I am afraid that if I continue to keep fixing the stripped down implementation, I will reach to a functional implementation that is the same as the upstream one.
> 
> Makes sense - I guess we look for other tricks to reduce the size. E.g., we could remove unused kernel modules or we could compress the kernel more (there are new options in 6.x).

I think we can revisit this part on the next overhaul of the Flatcar initrd dracut modules, as the tmpfiles dev early services and maybe more modules need some attention (removal). As it seems right now, systemd will just add more and more binaries / size as time passes because of the feature set increase.

ader1990:
PR is ready to be reviewed.

pothos:
Ive also start a Jenkins build: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3535/cldsv/
The test reports will show up on bincache

pothos:
The main blocker I see is that we have passed the 50% mark for the boot partition:
```
  File    Size  Used Avail Use% Type
 +/boot   127M   64M   63M  51% vfat
```
This is really problematic because the kernel should fit twice for the A/B update to work. Currently it's still possible to update but I fear that simple bugfix/security updates would reach the limit and lead to a situation where we can't release quickly.
I suggest we look into the following areas:
- Review the kernel config for size tweaks - the new generic compression option could help and we can also disable things we don't need/want to support
- Review firmware and kernel modules in the initrd - the smallest combination would be to only have what we need to mount the `/usr` partition and from there one we could load the other kernel modules from there. To reduce confusion I we could even try to compile all things we need into the kernel instead of having modules, and then we can bind mount the kernel module folder from `/usr` into the initrd.
- Review binaries in the initrd - here again the smallest combination is what we need to mount the `/usr` partition and from that point on we can rely on the binaries from there through the wrapper trick we currently use for some stuff already.

ader1990:
With the commit for ader1990/bootengine https://github.com/ader1990/bootengine/commit/5fcbbee367b58d5efc08a57834c0ac6a555721a3, this is the /boot partition size in KB:

```
/dev/sda1         129039 64571     64468  51% /mnt
```

Without this PR, the /boot partition size is:

```
/dev/sda1         129039 64130 64909  51% /mnt
```

Note the increase of only 441 KB, which tips over the >50%.

jepio:
Sorry for not asking sooner:
> during initrd stage - dracut-cmdline-ask.service fails because systemd-vconsole-setup.service fails (/usr/bin/loadkeys fails) - fixed by adding the dracut internationalization module

Why do we want `dracut-cmdline-ask.service` to run? It doesnt feel like this makes much sense for Flatcar.

pothos:
> I've also started a Jenkins build: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3535/cldsv/ The test reports will show up on bincache

Started a new run http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3637/cldsv/ which should land test results for all platforms on https://bincache.flatcar-linux.net/testing/9999.0.0+ader1990-upgrade-to-systemd-255-v2

pothos:
Can you squash the smaller commits that touch the systemd ebuild folder into one "Apply Flatcar changes" patch? When doing so it would be good to split the one that also changes the `profiles` folder. That way we can (with resolving some conflicts) apply the "Apply Flatcar changes" commit again after syncing from upstream.

sdk_container/src/third_party/coreos-overlay/sys-apps/systemd/systemd-255.3.ebuild
```diff
@@ -572,7 +562,7 @@ multilib_src_install_all() {
 	# Flatcar: Use an empty preset file, because systemctl
 	# preset-all puts symlinks in /etc, not in /usr. We don't use
 	# /etc, because it is not autoupdated. We do the "preset" above.
-	rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
+	# rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
```
pothos, line 565: Why is this removed? This could be the reason why some enablement symlinks end up in the  /etc upperdir

pothos: I guess we could make use of presets if we add a step in the image postprocessing where we evaluate the presets to create symlinks from them.

ader1990: /lib/systemd/system-preset/90-systemd.preset does not exist anymore on this version of systemd.

ader1990: hmm, I think the path is incorrect due to the split-usr usex, need to check again.

ader1990: `rm: cannot remove '/var/tmp/portage/sys-apps/systemd-255.3/image/lib/systemd/system-preset/90-systemd.preset': No such file or directory` - this is the log message.

ader1990: `/var/tmp/portage/sys-apps/systemd-255.3/image/usr/lib/systemd/system-preset/90-systemd.preset` exists.

ader1990: @pothos testing now locally with

```
diff --git a/sdk_container/src/third_party/coreos-overlay/sys-apps/systemd/systemd-255.3.ebuild b/sdk_container/src/third_party/coreos-overlay/sys-apps/systemd/systemd-255.3.ebuild
index 7e8266d7d2..b1584034c7 100644
--- a/sdk_container/src/third_party/coreos-overlay/sys-apps/systemd/systemd-255.3.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-apps/systemd/systemd-255.3.ebuild
@@ -295,7 +295,7 @@ src_configure() {

 # Flatcar: Our function, we use it in some places below.
 get_rootprefix() {
-       usex split-usr "${EPREFIX:-/}" "${EPREFIX}/usr"
+       "${EPREFIX}/usr"
 }
 multilib_src_configure() {
        local myconf=(
@@ -562,8 +562,8 @@ multilib_src_install_all() {
        # Flatcar: Use an empty preset file, because systemctl
        # preset-all puts symlinks in /etc, not in /usr. We don't use
        # /etc, because it is not autoupdated. We do the "preset" above.
-       # rm "${ED}$(usex split-usr '' /usr)/lib/systemd/system-preset/90-systemd.preset" || die
-       insinto $(usex split-usr '' /usr)/lib/systemd/system-preset
+       rm "${ED}/usr/lib/systemd/system-preset/90-systemd.preset" || die
+       insinto /usr/lib/systemd/system-preset
        doins "${FILESDIR}"/99-default.preset

        # Flatcar: Do not ship distro-specific files (nsswitch.conf

```

ader1990:
> Can you squash the smaller commits that touch the systemd ebuild folder into one "Apply Flatcar changes" patch? When doing so it would be good to split the one that also changes the `profiles` folder. That way we can (with resolving some conflicts) apply the "Apply Flatcar changes" commit again after syncing from upstream.

I left the 3 commits separate to have a better idea on what the process was. The changes to the profiles folder is related to the kbd package requirement. I will update the PR with the 3 squashed commits and split the kbd part into two: kbd enable and systemd-vconsole-setup fixes after the systemd preset issue is resolved.

changelog/changes/2024-03-13-systemd-upgrade-from-252-to-255.md
```diff
@@ -0,0 +1 @@
+- Upgraded systemd from version 252 to version 255 ([flatcar/scripts#1679](https://github.com/flatcar/scripts/pull/1679)).
```
pothos, line 1: That entry would rather go into `changelog/updates/` as `- systemd ([255](https://github.com/systemd/systemd/releases/tag/v255))`

ader1990: sure, will do the update. as the two changes are cosmetic, I will wait for the CI to finish before I do the force push.

pothos (approved):
Thanks a lot, looks good!

@krnowak Can you also do a quick review? I wonder why `usex split-usr '' /usr` didn't work.

PR #1684, opened 2024-02-21
sys-kernel/bootengine: fix systemd-vconsole-setup for systemd 255

ader1990:
Use https://github.com/flatcar/scripts/pull/1679 for fixing the upcoming systemd 255 upgrade.

Required-by: https://github.com/flatcar/scripts/pull/1679

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
@@ -10,7 +10,7 @@ CROS_WORKON_REPO="https://github.com"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	CROS_WORKON_COMMIT="a85e1977b29dbe8315733dbe1b1ab3ab84d039a2" # flatcar-master
+	CROS_WORKON_COMMIT="5f73716fe5370ee9a269f080d445c85479b745ab" # flatcar-master
 	KEYWORDS="amd64 arm arm64 x86"
 fi
 
--- a/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
@@ -2,7 +2,7 @@
 # Distributed under the terms of the GNU General Public License v2
 
 EAPI=7
-CROS_WORKON_PROJECT="flatcar/bootengine"
+CROS_WORKON_PROJECT="ader1990/bootengine"
 CROS_WORKON_LOCALNAME="bootengine"
 CROS_WORKON_OUTOFTREE_BUILD=1
 CROS_WORKON_REPO="https://github.com"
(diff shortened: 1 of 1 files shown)


ader1990:
Firstly, let's run the CI to get the size of the kernel / initrd / image without any changes. To be logged here.
Secondly, let's run the CI to get the size of the kernel / initrd / image with the dracut i18n. To be logged here.
Thirdly, if the size increase is beyond an acceptable threshold, to create a special dracut module in bootengine that has only the loadkeys binary, run the CI. To be logged here.

ader1990:
size of the kernel / initrd / image without any changes vanilla

ader1990:
size of the kernel / initrd / image with dracut module i18n vanilla

Difference versus vanilla of the vmlinuz is 1MB increase.

pothos:
Thanks for the numbers, 1MB is quite a lot for the initrd - we reduced the initrd size with many tricks already and I think we need one more here by only inserting the required binary instead of the whole internationalization module.

ader1990:
> Thanks for the numbers, 1MB is quite a lot for the initrd - we reduced the initrd size with many tricks already and I think we need one more here by only inserting the required binary instead of the whole internationalization module.

I looked on the systemd / dracut issues related to this topic, and the amount of discussion / code change is quite high -> https://github.com/systemd/systemd/issues/6689, https://github.com/systemd/systemd/pull/15479, https://github.com/systemd/systemd/commit/e019ea738d63d5f7803f378f8bd3e074d66be08f, https://github.com/systemd/systemd/commit/a6f3a7eb8acb457e2d0a9974a03662870bdf0de0, https://github.com/dracutdevs/dracut/pull/797.

If we go with a non-standard way, we will probably open the gates for random bugs / behaviour during the initrd stage.

Testing under way for the non-standard implementation: https://github.com/ader1990/bootengine/commit/a92ce96dce7a4965943b24385f2d3a14ef07bd81

ader1990:
> > Thanks for the numbers, 1MB is quite a lot for the initrd - we reduced the initrd size with many tricks already and I think we need one more here by only inserting the required binary instead of the whole internationalization module.
> 
> I looked on the systemd / dracut issues related to this topic, and the amount of discussion / code change is quite high -> [systemd/systemd#6689](https://github.com/systemd/systemd/issues/6689), [systemd/systemd#15479](https://github.com/systemd/systemd/pull/15479), [systemd/systemd@e019ea7](https://github.com/systemd/systemd/commit/e019ea738d63d5f7803f378f8bd3e074d66be08f), [systemd/systemd@a6f3a7e](https://github.com/systemd/systemd/commit/a6f3a7eb8acb457e2d0a9974a03662870bdf0de0), [dracutdevs/dracut#797](https://github.com/dracutdevs/dracut/pull/797).
> 
> If we go with a non-standard way, we will probably open the gates for random bugs / behaviour during the initrd stage.
> 
> Testing under way for the non-standard implementation: [ader1990/bootengine@a92ce96](https://github.com/ader1990/bootengine/commit/a92ce96dce7a4965943b24385f2d3a14ef07bd81)

I tried with the stripped down version of the i18n dracut module, using this HEAD https://github.com/ader1990/bootengine/commit/9f37ee524abf2406c2f3ee1539211ccf02e1340b. The loadkeys binary was succesfully installed in the initrd, but the execution of the binary failed with `localhost systemd-vconsole-setup[254]: loadkeys: Unable to open file: us: No such file or directory`. Which means basically that the definition of the https://github.com/dracutdevs/dracut/blob/master/modules.d/10i18n/module-setup.sh has to match really closely with the stripped down version, leading to basically reusing the same dracut module but with the overhead in the future of maintaining it.

At this moment, I can only suggest to use the upstream version of the dracut module, the one that is properly tested to work with systemd.

ader1990:
Closing PR, systemd 255 will use the i18n dracut module directly.

PR #1685, opened 2024-02-21
Prepare SELinux packages in overlay for upstreaming

krnowak:
~Based on and blocked on #1672.~

The weekly updates updated SELinux userspace tools to version 3.6. We have two packages in overlay that were not yet updated - sys-apps/policycoreutils and sys-libs/libsemanage. This PR updates both packages to 3.6 too and tries to remake Flatcar modifications to the ebuilds, so they will still be useful for us and be eventually upstreamable.

The modifications were as follows:

- Drop dependency on python:
  - I tried making it optional by putting it behind the `python` USE flag.
  - In case of policycoreutils, python dependency did come mostly through the Gentoo-specific addition of policycoreutils-extra tarball that has some scripts written in python. For this, I have added a `vanilla` USE flag, that would skip the extra Gentoo stuff.
- Make `/var/lib/selinux` a symlink to `/usr/lib/selinux/policy`.
  - ~I moved that modification to `coreos-base/misc-files`.~
  - This modification is split in two - tmpfiles stuff went to `coreos-base/misc-files`, overriding the path went to the profiles bash hook. More about this in commit messages and comments.

~One thing I'm not sure about is whether to drop python support SELinux only for generic images or for both generic and SDK images. Keeping python support for SELinux with current changes I have made to policycoreutils and libsemanage ebuilds results in needing to pull app-admin/setools, dev-python/networkx and sys-apps/selinux-python packages into SDK.~
We dropped python support for both generic and SDK images - we don't need the python tools.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1342/cldsv/ (more or less passed, kicked off another one, because I dropped one INSTALL_MASK).

--

- sys-apps/policycoreutils: [PROD] [DEV]
  - from 3.5 to 3.6
  - bumped python compat to 3.10..3.11
  - moved the use of tmpfiles to coreos-base/misc-files
  - added vanilla and python USE flags (we enable vanilla and disable python)
    - vanilla tells the ebuild to use vanilla sources, disabling it pulls in some Gentoo extras (which use python)
  - moved the override of selinux policy modules directory into profiles bash hook

- sys-libs/libsemanage: [PROD] [DEV]
  - from 3.5 to 3.6
  - bumped python compat to 3.10..3.11
  - added the enabled-by-default python USE flag (to be upstreamed)
  - changed dep on sys-devel/{bison,flex} to app-alternatives/{yacc,lex}
  - moved config generation into a patch (to be upstreamed)
    - so we can add out own patch that changes the compression setting there

- sys-process/audit: [PROD] [DEV]
  - from 3.0.6-r1 to 3.1.1
  - EAPI 8
  - bumped python compat to 3.9..11
  - moved the use of tmpfiles to coreos-base/misc-files
  - added the io-uring USE flag
    - we probably could think about enabling it
  - dropped most of the modifications
  - release notes: https://github.com/linux-audit/audit-userspace/releases/tag/v3.1.1

- dev-lang/swig:
  - from 4.0.2 to 4.1.1-r1
  - EAPI 8
  - replaced dep on dev-libs/libpcre with dev-libs/libpcre2
  - release notes: https://sourceforge.net/p/swig/news/2022/11/-swig-411-released/

--

- [x] changelog
- [x] image diff


krnowak:
CI passed.

krnowak:
> LGTM. Thanks for working on this last bit of the SELinux automation: now all the SELinux upgrades (policies and core packages) will be done without effort. Allowing us to focus on the other SELinux related topics 😉 Thanks!

Sure thing! I hope that upstream will be willing to take our changes.

> 
> Some notes:
> 
> * Regarding the OpenStack failure, it will disappear if you rebase on `main` but that's not needed of course.

Oh, cool. Good to know.

> * What about opening a ticket on Gentoo https://bugs.gentoo.org/ to get folks input, get a reference when we'll send the Python patch upstream and not forget about it?

Oh yeah, I want to upstream this stuff, obviously. But I was rather thinking about filing a github PR. Would a ticket on bugzilla be better?

PR #1688, opened 2024-02-22
nvidia-drivers: Improve match for NVIDIA GPUs

jepio:
# nvidia-drivers: Improve match for NVIDIA GPUs
The logic for checking if we need to probe depends on the device showing up with the product type in lspci output, which doesn't hold for all sorts of GPUs. The NVIDIA_PRODUCT_TYPE is used for fetching the drivers and is "tesla" across datacenter GPUs.

Switch to matching on vendor id and device class. The values are the same ones that the nvidia driver binds to.

## How to use

## Testing done


jepio:
This addresses #1364

jepio:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3509/cldsv/

krnowak (commented):
I guess that the changes are fine, but I'd like to ask you to bump the revision of the x11-drivers/nvidia-drivers ebuild.

sayanchowdhury (approved):
LGTM. Agreed with @krnowak thoughts.

krnowak (approved):

PR #1700, opened 2024-02-24
Upgrade Linux Kernel for flatcar-3850 from 6.6.16 to 6.6.18

flatcar-infra:
Subject says it all.


dongsupark:
CI fails at build_image only with amd64, added `on-hold`.

PR #1707, opened 2024-02-27
sys-kernel/bootengine: Restore support for custom PXE OEM contents

pothos:
This pulls in https://github.com/flatcar/bootengine/pull/88 to restore looking at the documented /usr/share/oem path in a custom PXE OEM initrd instead of /oem.

## How to use

Backport to Stable

## Testing done

```
tree initrd/
initrd/
└── usr
    └── share
        └── oem
            └── config.ign
(cd initrd; find . -print0 | cpio --null --create --verbose --format=newc > ../merged.cpio )
zcat flatcar_production_pxe_image.cpio.gz >> merged.cpio
gzip merged.cpio
./flatcar_production_pxe.sh -- -append 'flatcar.first_boot=1' -initrd ../merged.cpio.gz
# checked that ignition config got applied
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ on:
         description: |
           Space-separated vendor formats to build.
         required: true
-        default: qemu_uefi
+        default: qemu_uefi pxe
       custom_sdk_version:
         type: string
         required: false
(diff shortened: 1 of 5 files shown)


jepio (approved):

pothos:
With the CI image I can't easily test this. I think it's worth having the PXE image available in the VM images artifacts. It's quite handy for many test cases. I've added a commit, let's see if that works.

PR #1710, opened 2024-02-29
sys-boot/shim: Fix parallel build

jepio:
# sys-boot/shim: Fix parallel build

The bundled gnu-efi build is implemented in a buggy way that can break when built in parallel. We've hit this in the nightly sdk build. Add a patch for it.

The patch has been posted upstream at https://github.com/rhboot/shim/pull/643.

## How to use

## Testing done


krnowak (approved):
Look good overall, but could rename the ebuild instead of making a symlink to it? Thanks!

PR #1717, opened 2024-03-01
Monthly GLSA metadata 2024-03-01

flatcar-infra:
Updated GLSA metadata


dongsupark:
Blocked by samba 4.18.9, [202402-28](https://security.gentoo.org/glsa/202402-28).
Added `on-hold`.

I do not believe the samba issue actually affects Flatcar, but in this case, a manual update to 4.18.9 is necessary.

dongsupark:
See instead https://github.com/flatcar/scripts/pull/1814

PR #1723, opened 2024-03-02
Upgrade Linux Kernel for flatcar-3850 from 6.6.16 to 6.6.19

flatcar-infra:
Subject says it all.


dongsupark:
See instead https://github.com/flatcar/scripts/pull/1724

PR #1730, opened 2024-03-05
Mount btrfs fs with rescue= option

jepio:
# Mount btrfs fs with rescue= option

It looks like 'norecovery' is deprecated and has been removed in the v6.8-rc1 kernel. Replace it with 'rescue=nologreplay', which is a replacement implemented since v5.9. The standalone 'nologreplay' option is also deprecated.

Also need to pull in this:
- [x] https://github.com/flatcar/update_engine/pull/40
but thats tied to us updating to v6.8+ which means we still have time.

## How to use

## Testing done


pothos:
Thanks

ader1990:
Hello @jepio, I am usually running my dev env on Ubuntu latest version aka 24.04 Noble, and this week the kernel got upgraded from 6.6 to 6.8. Without this patch, `./build_image` fails, thanks for the fix.

ader1990 (approved):
LGTM. required for Ubuntu 24.04.

jepio:
Thanks; I've added a commit to bump update_engine, will merge after running tests.

PR #1731, opened 2024-03-06
Discussion about the files in the /boot and initrd

ader1990:
This PR was opened to have a place where to discuss line by line on the files that we can or cannot remove from the /boot partition and initrd. Or maybe move them into the /usr.

This is a follow up on @pothos suggestion to cut things we do not need: https://github.com/flatcar/Flatcar/issues/1381#issuecomment-1969285023


pothos:
Thanks, we can also discuss the kernel config: https://alpha.release.flatcar-linux.net/amd64-usr/current/flatcar_production_image_kernel_config.txt

boot_partition.txt
```diff
@@ -0,0 +1,812 @@
+[   57467608]  ./flatcar/vmlinuz-a*
+[     949440]  ./EFI/boot/bootx64.efi*
+[     855912]  ./EFI/boot/mmx64.efi*
+[     783864]  ./flatcar/grub/x86_64-efi/core.efi.signed*
+[     783864]  ./EFI/boot/grubx64.efi*
+[     782336]  ./flatcar/grub/x86_64-efi/core.efi*
```
jepio, line 6: these are all the same file and i want to say that we only need `EFI/boot/grubx64.efi`, but this needs testing.

pothos: Unless we would also remove them in the update postinst hook these would still occupy space to prevent an update - but even then for some old clients the postinst hook might be too late because the kernel can't even be extracted to `/boot`.

initrd.txt
```diff
@@ -0,0 +1,1397 @@
+[   12539904]  ./rootfs-0/kernel/x86/microcode/GenuineIntel.bin
+[    5343472]  ./rootfs-1/usr/lib64/libcrypto.so.3*
+[    3208664]  ./rootfs-1/usr/lib/systemd/libsystemd-shared-252.so*
+[    2591624]  ./rootfs-1/usr/lib/gcc/x86_64-cros-linux-gnu/13/libstdc++.so.6.0.32*
+[    2058480]  ./rootfs-1/usr/lib/systemd/libsystemd-core-252.so*
+[    1933576]  ./rootfs-1/usr/lib64/libc.so.6*
+[    1845305]  ./rootfs-1/usr/lib/firmware/phanfw.bin
+[    1649076]  ./rootfs-1/usr/lib/firmware/mellanox/mlxsw_spectrum-13.2010.1006.mfa2
```
pothos, line 8: In theory we could the network firmware and kernel modules from `/sysusr/usr` but it needs some wrangling to have them available at the right time…

boot_partition.txt
```diff
@@ -0,0 +1,812 @@
...
+[     470624]  ./xen/pvboot-x86_64.elf*
+[     470624]  ./flatcar/grub/x86_64-xen/core.elf*
```
jepio, line 8: Same file, and i'm not convinced pvboot elf works. Hasn't been tested in years/ever.

initrd.txt
```diff
@@ -0,0 +1,1397 @@
+[   12539904]  ./rootfs-0/kernel/x86/microcode/GenuineIntel.bin
```
pothos, line 1: The microcode updates can be delayed by defining a service that does `echo 1 > /sys/devices/system/cpu/microcode/reload` once we would symlink/copy them to `/lib/firmware/{intel-ucode,amd-ucode}` in the initrd from `/sysusr`.

initrd.txt
```diff
@@ -0,0 +1,1397 @@
...
+[    2058480]  ./rootfs-1/usr/lib/systemd/libsystemd-core-252.so*
+[    1933576]  ./rootfs-1/usr/lib64/libc.so.6*
+[    1845305]  ./rootfs-1/usr/lib/firmware/phanfw.bin
+[    1649076]  ./rootfs-1/usr/lib/firmware/mellanox/mlxsw_spectrum-13.2010.1006.mfa2
+[    1649016]  ./rootfs-1/usr/lib/systemd/systemd-networkd*
+[    1359116]  ./rootfs-1/usr/lib/firmware/mellanox/mlxsw_spectrum2-29.2010.1006.mfa2
+[    1167440]  ./rootfs-1/usr/sbin/nvme*
+[    1145040]  ./rootfs-1/usr/lib64/libgcrypt.so.20.4.3*
+[    1127648]  ./rootfs-1/usr/lib/firmware/mellanox/mlxsw_spectrum3-30.2010.1006.mfa2
+[    1038064]  ./rootfs-1/usr/sbin/btrfs*
```
pothos, line 14: That binary could be symlinked to `/sysusr/`, same for xfs* and mkfs.* stuff and even `ip` and `mdadm`

initrd.txt
```diff
@@ -0,0 +1,1397 @@
...
+[     309744]  ./rootfs-1/usr/lib/modules/6.6.17-flatcar/kernel/drivers/net/ethernet/qlogic/qed/qed.ko.xz
+[     303768]  ./rootfs-1/usr/lib/modules/6.6.17-flatcar/kernel/drivers/net/ethernet/mellanox/mlxsw/mlxsw_spectrum.ko.xz
+[     275128]  ./rootfs-1/usr/lib/firmware/ql2500_fw.bin
+[     272792]  ./rootfs-1/usr/bin/systemctl*
+[     271288]  ./rootfs-1/usr/lib64/libtinfow.so.6.4*
+[     268616]  ./rootfs-1/usr/lib/modules/6.6.17-flatcar/kernel/drivers/net/ethernet/chelsio/cxgb4/cxgb4.ko.xz
+[     267160]  ./rootfs-1/usr/lib64/libtinfo.so.6.4*
+[     266400]  ./rootfs-1/usr/lib64/libsmartcols.so.1.1.0*
+[     265300]  ./rootfs-1/usr/lib/firmware/ql2400_fw.bin
+[     254499]  ./rootfs-1/usr/share/ca-certificates/ca-certificates.crt
```
pothos, line 74: Small win, but that file could be symlinked.

initrd.txt
jepio, line 8: Spectrum is a network switch, i think we could get rid of these firmware files.

initrd.txt
```diff
@@ -0,0 +1,1397 @@
...
+[     971680]  ./rootfs-1/usr/sbin/ldconfig*
+[     968312]  ./rootfs-1/usr/lib64/libssl.so.3*
+[     967632]  ./rootfs-1/usr/lib64/libsystemd.so.0.35.0*
+[     943352]  ./rootfs-1/usr/lib/firmware/qed/qed_init_values_zipped-8.59.1.0.bin
+[     911472]  ./rootfs-1/usr/lib64/libm.so.6*
+[     893472]  ./rootfs-1/usr/bin/bash*
+[     814776]  ./rootfs-1/usr/lib64/libsepol.so.2*
+[     812616]  ./rootfs-1/usr/sbin/xfs_db*
+[     785272]  ./rootfs-1/usr/lib64/libzstd.so.1.5.5*
+[     745872]  ./rootfs-1/usr/sbin/xfs_repair*
```
jepio, line 24: i dont think we have anything that would call xfs_repair from the initrd

pothos:
My impression is that we should start with the big stuff: find a way to load network firmware and kernel modules, and CPU microcode update from `/sysusr/usr/`.

ader1990:
> Thanks, we can also discuss the kernel config: https://alpha.release.flatcar-linux.net/amd64-usr/current/flatcar_production_image_kernel_config.txt

Added this file for discussion.

flatcar_production_image_kernel_config.txt
```diff
@@ -0,0 +1,6342 @@
...
+CONFIG_MLX5_CORE_EN_DCB=y
+# CONFIG_MLX5_CORE_IPOIB is not set
+CONFIG_MLX5_SW_STEERING=y
+# CONFIG_MLX5_SF is not set
+CONFIG_MLXSW_CORE=m
+CONFIG_MLXSW_CORE_HWMON=y
+CONFIG_MLXSW_CORE_THERMAL=y
+CONFIG_MLXSW_PCI=m
+CONFIG_MLXSW_I2C=m
+CONFIG_MLXSW_SPECTRUM=m
```
ader1990, line 2732: @jepio if this two lines can be removed https://cateee.net/lkddb/web-lkddb/MLXSW_SPECTRUM.html, I think it would make a big difference ~2MB, offering a small runway for a few patches. Should I start a PR with the scope of only removing / cleaning up modules/fw?

initrd.txt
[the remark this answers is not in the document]
ader1990: this means if the files we d like to move/remove have a great compression rate, it does not make too much sense to remove them?

initrd.txt
ader1990: ` 51M Mar  6 11:52 flatcar_production_image.vmlinuz` with CONFIG_MLXSW_SPECTRUM and CONFIG_MLXSW_SPECTRUM_DCB removed vs 55M the HEAD one.

initrd.txt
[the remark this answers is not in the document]
jepio: the initrd is compressed with some algorithm (zstd i think), and so we would want to focus on files that are big when compressed so that removing them actually has the desired impact. Otherwise we might remove a file that looks big but doesn't give much benefit because it compresses to something really small.

flatcar_production_image_kernel_config.txt
jepio: Maybe. Is it also possible to keep it enabled but not include these drivers/files in the initrd?

jepio: If it proves to be too much work then yes, the whole MLXSW_CORE can go.

initrd.txt
ader1990: ```/dev/vda1          129039      59962      69078  47% /boot``` with the removal on AMD64.

ader1990: Created PR https://github.com/flatcar/scripts/pull/1734 as a first version of the effort to decrease the initrd size.

ader1990:
I will keep in this updated comment a log of the AMD64 /boot partition size history, to have a clear picture on the size increase over time:

```
# 2024-03-18, Flatcar ALPHA 3913.0.0 https://alpha.release.flatcar-linux.net/amd64-usr/3913.0.0/
TBD
# 2024-03-25, Flatcar HEAD https://github.com/flatcar/scripts/commit/6ebffa7a4fde3a1b2c8d362b065b35b4ce3a43e4
/dev/sda1         129039 62503     66536  49% /boot
```

chewi:
In the past year, I have already made savings wherever possible with GRUB, changed how we compress the kernel image, and trimmed parts of the initrd in the face of unavoidable increases elsewhere. It's very tight now, and trying to squeeze out more savings is just prolonging the inevitable. I've had a plan since the start of the year (which I have discussed openly) to move the kernel image out of /boot into /usr. Updating Dracut was just the start of that, and now I'm moving onto the main effort. Just bear with me.

chewi:
Small update. Jeremi has requested that I write up the details of what I'm proposing. That's half done now, and I'll finish it soon, but I was also on the verge of a major milestone, so I was keen to go that little bit further. Flatcar now boots with the kernel under /usr, the kernel parameters and verity root hash GPG signed for verification by GRUB, and the verity root hash also S/MIME signed for verification by the kernel. :slightly_smiling_face:

jepio:
> This was initially done so that the old GRUB could boot new releases, but given that the microcode update will immediately push everyone over the edge, a new GRUB config will be needed regardless.

What's the link between the microcode update and needing a new GRUB config? I don't think we have a shared understanding on the team in this area.

Has somebody looked into why an incremental microcode update would lead to such a jump in size? We really should be more conservative with what we allow into the kernel/initrd with regards to code size - we can't just keep growing without any control. And yes - this does require solid engineering effort.

chewi:
Being able to boot new releases with the old GRUB config assumes that at least some users will have enough space left in /boot to install the new kernels there for a while before they really do become too large. This microcode update inflates the image by about 3MB, which is roughly how much space we have left now **in the best case**. All users would therefore have no choice but to at least take a new GRUB config, if not a new GRUB image.

Krzesimir observed this size increase in CI. As yet, I have only looked at the size of the original tarball. This is also about 3MB larger than the previous one, so that tallies. Krzesimir said that the increase is due to a new Intel CPU family. We can probably postpone that for a short time, but if anything, it's the older CPU families we should be pruning. I imagine the updates for the really old ones might be quite small though.

Having worked closely with Dracut, there are only small gains to be had now. I already had to work hard to avoid going over the limit with the recent update, e.g. preventing the inclusion of libgcrypt. I would prefer to have more space to enable the features users would like to have than to break our backs keeping the size down when it will inevitably not be enough.

chewi:
In case you missed it on Matrix, I'm now working on an entirely new idea. That's not to say the old idea was bad, this one is just better, and most importantly, much simpler. :grin:

While comparing what we do with Bottlerocket, I was intrigued by the fact that they do not use an initrd at all. That got me thinking about why Flatcar _does_ use one, and the only reason seems to be that our image is for /usr, not /. Isn't it also because we need to be able to wipe the ROOT before mounting USR? No, because we already mount USR at /sysusr to get Ignition before that happens. So, why don't we just mount it at /usr in the first place? How much of an initrd do you need to do that? Hardly anything, it turns out.

After copying some bits from our existing Dracut-based initrd to USR like the Ignition systemd units, I have more or less managed to boot Flatcar normally using an initrd that has just:

* A few empty directories and symlinks.
* A sysroot.mount unit for ROOT.
* A sysroot-usr.mount unit to bind mount from /usr.
* A static busybox.
* This shell script:

```sh
#!/busybox ash

set -e
/busybox mount -t devtmpfs devtmpfs /dev
/busybox mount -t proc proc /proc
/busybox mount -o ro /dev/vda3 /usr
exec /lib/systemd/systemd "${@}"
```

Okay, I've hardcoded /dev/vda3 and there's no verity here, but this is just a prototype. I know from Bottlerocket that adding verity wouldn't be hard. Don't like busybox? That could be changed, but I think it's fine.

Any drivers needed to access the disk will admittedly need to change from being modules to built-ins, but I don't believe that will hurt much. With this approach, it would be worth me going through the kernel config see whether anything currently built-in could become a module.

The upshot? Excluding microcode, instead of an 86MB uncompressed initrd, we now have a 2.4MB one. No kernel move or GRUB config changes required. It also allows users to use tools in early boot that are currently not available until after switching root.

Suffice to say I'm, going to divert onto this plan. Suggestions welcome.

PR #1733, opened 2024-03-06
Drop unused net-libs/libssh2 package

krnowak:
It's not used anywhere. Checked build logs of SDK and packages and net-libs/libssh2 shows up nowhere.

Fixes https://github.com/flatcar/Flatcar/issues/1311.


dongsupark (approved):
Good catch. Looks like it is really not used.

PR #1736, opened 2024-03-07
Upgrade Linux Kernel for main from 6.6.17 to 6.6.21

flatcar-infra:
Subject says it all.


dongsupark:
CI fails at build_image when running sbsign only with amd64, added `on-hold`.

pothos:
With the revert the build seems to work now

pothos (approved):

PR #1742, opened 2024-03-11
Build ZFS sysext with each release

jepio:
# Build ZFS sysext with each release

Add a ZFS sysext that is built with every Flatcar release and can be enabled through the `/etc/flatcar/enabled-sysext.conf` mechanism that @pothos added. Since it provides a kernel module, it is tightly coupled to the Flatcar release. To solve issues with etc configs and udev rules needing to be run after the sysext is loaded, I created a tmpfiles.d that sets up the required symlinks and ordering between sysext.service and udev.service.

Following @pothos advice, i've added a step for building the sysexts after the image build. This DEFAULT_SYSEXTS can be extended to add more sysexts.

There are also some other improvements that I've added:
- sysext building can happen in parallel
- SOURCE_DATE_EPOCH is set for sysext build. I've taken the timestamp of the /usr/lib/os-release file as the source of the timestamp.
- coreos-kernel now provides a better module build environment in /lib/modules/.../build so that the ZFS mod can be built and which should simplify building nvidia-drivers.
- app-alternatives/awk is now pulled into the image instead of gawk.
- `update_ebuild --portage=rsync` now works for eclasses

Pulls in this:
- [ ] https://github.com/flatcar/bootengine/pull/89
## How to use

(still need to try the end to end flow once this is on bincache)
```
./build_image
./build_image sysext
scp flatcar-zfs.raw flatcar:/etc/extensions/
```

## Testing done


build_library/prod_image_util.sh
```diff
@@ -247,3 +247,42 @@ create_prod_tar() {
   sudo losetup --detach "${lodev}"
   upload_image "${container}"
 }
+
+DEFAULT_SYSEXTS=(
```
pothos, line 251: Can we call this EXTRA_SYSEXTS as these won't be installed by default? I imagine that we could have `DEFAULT_SYSEXTS` in the future for things that are pre-enabled.

build_library/prod_image_util.sh
```diff
@@ -247,3 +247,42 @@ create_prod_tar() {
...
+  flatcar-zfs:sys-fs/zfs
```
pothos, line 252:
```suggestion
  zfs:sys-fs/zfs
```
How about prepending `flatcar-` in the function below because this is not free form but part of the contract to identify the payloads.

build_library/sysext_mangle_flatcar-zfs
```diff
@@ -0,0 +1,43 @@
...
+pushd "${rootfs}"
+
+rm -rf ./usr/{lib/debug/,lib64/cmake/,include/}
+rm -rf ./usr/lib/dracut/
+rm -rf ./usr/share/initramfs-tools
+rm -rf ./usr/src
+
+mkdir -p ./usr/share/flatcar/etc
+rm -rf ./etc/{csh.env,environment.d/,profile.env}
+cp -a ./etc/. ./usr/share/flatcar/etc/
```
pothos, line 15: I fear that this may be confusing - how about a different path?
The use of `/usr/lib/systemd/system/systemd-udevd.service.d/10-zfs.conf` is ok because this would ideally work once we set up the overlay early enough but putting files under `/usr/share/flatcar/etc/` won't make them show up.
```suggestion
cp -a ./etc/. ./usr/share/zfs/etc/
```

build_library/prod_image_util.sh
jepio: will do

build_library/sysext_mangle_flatcar-zfs
jepio: make sense

build_library/prod_image_util.sh
krnowak: I think this listing should be in a separate file (see `ci-automation/base-sysexts.sh`), so in future an automation could obtain a list of extra sysexts (for generating image changes or deciding if a package update affects some sysext). With this done, I think that the list of extra sysexts should be passed directly to `create_prod_sysexts` function then.

build_library/prod_image_util.sh
jepio: I'll do this, but i started to question: why the need for the prefix?

pothos: The idea is to distinguish Flatcar-provided extensions from user-provided extensions, e.g., in the output of `systemd-sysext status`

pothos:
Overall this looks very nice :)
For release publishing we already have the required logic in download_payloads and generate_payload so I hope that it works smooth and we don't run into bugs while we exercise this path for the first time.

build_library/prod_image_util.sh
jepio: In a different file for sourcing? That could make sense, but then it shouldn't be in `ci-automation` and instead somewhere in `build_library`, so that it can be shared with `build_image`. Which sysexts we build shouldn't be a CLI parameter.

krnowak: The file in ci-automation was an example. It's fine if it's under `build_library`.

jepio: moved to `./build_library/extra_sysexts.sh`

build_library/prod_image_util.sh
jepio: done

jepio:
Had to force push because i forgot to add `zfs` to board-packages.

jepio:
Needed to fix the arm64 build and the layout on bincache.

jepio:
@pothos want to have this in the nearest alpha?

pothos:
Would be good because the bootengine PR is already merged

jepio:
@pothos pushed a changelog, if you have a suggestion you can commit it directly.

@sayanchowdhury lets merge this before building alpha.

changelog/changes/2024-03-13-zfs.md
```diff
@@ -0,0 +1 @@
+- Provide a zfs-2.2.2 sysext with the release. Write 'zfs' to /etc/flatcar/enabled-sysext.conf through ignition and the sysext will be installed during provisioning. ZFS support is experimental and ZFS is not supported for the root partition. ([flatcar/scripts#1742](https://github.com/flatcar/scripts/pull/1742))
```
pothos, line 1:
```suggestion
- Provided a ZFS-2.2.2 Flatcar extension as optional systemd-sysext image with the release. Write 'zfs' to `/etc/flatcar/enabled-sysext.conf` through Ignition and the sysext will be installed during provisioning. ZFS support is experimental and ZFS is not supported for the root partition. ([flatcar/scripts#1742](https://github.com/flatcar/scripts/pull/1742))
```

abuisine:
Cool to see this upstreamed !

jepio:
> Cool to see this upstreamed !

hey @abuisine, I was planning to write you to see if it works for your usecase. Still need to add docs, but mantle contains the test case we now have in our CI infra with butane examples, see here: 
https://github.com/flatcar/mantle/blob/flatcar-master/kola/tests/sysext/zfs.go#L19-L116

abuisine:
We'll make a bunch of tests next week for sure, will keep you posted

coreyjewett:
Thanks @jepio! I was previously using the [enix modules](https://github.com/flatcar/Flatcar/issues/1110#issuecomment-1700688879), but never got them to work right at boot. I switched to the `alpha` channel and 3913.0.0 loaded my pool as expected right after reboot. 🎉

PR #1756, opened 2024-03-15
profiles: Enable TPM2 support in systemd

pothos:
We could use systemd-cryptenroll and cryptsetup with a TPM device but so far the support was not compiled in.
Enable the use flags for TPM2 support in systemd.

## Testing done

systemd-cryptenroll works with a tpm but so far I didn't manage to make the systemd-cryptsetup generator unlock with a TPM


sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.use
```diff
@@ -29,7 +29,7 @@ app-admin/sudo -sendmail
 
 # disable hybrid cgroup as we use the unified mode now
 # use lzma which is the default on non-gentoo systems, enable selinux,
-sys-apps/systemd -cgroup-hybrid curl idn lzma selinux
+sys-apps/systemd -cgroup-hybrid curl idn lzma selinux tpm
```
ader1990, line 32: maybe we can also add the pkcs11 flag too? https://github.com/flatcar/Flatcar/issues/1394

ader1990:
There is a cyclical dep between the app-crypt/tpm-tss and virtual/tmpfiles and systemd. Need to see how to break it somehow:

```

  * Error: circular dependencies:
 
 (app-crypt/tpm2-tss-4.0.1:0/4::portage-stable, ebuild scheduled for merge to '/build/amd64-usr/') depends on
  (virtual/tmpfiles-0-r5:0/0::portage-stable, ebuild scheduled for merge to '/build/amd64-usr/') (runtime)
   (sys-apps/systemd-255.3:0/2::coreos, ebuild scheduled for merge to '/build/amd64-usr/') (runtime)
    (app-crypt/tpm2-tss-4.0.1:0/4::portage-stable, ebuild scheduled for merge to '/build/amd64-usr/') (buildtime_slot_op)
```

pothos:
One way could be using `TMPFILES_OPTIONAL=1` in `tpm2-tss`, the other could be to add an entry in the `break_dep_loop` list in `build_packages`

ader1990:
As Flatcar always uses systemd, we can make this package purely virtual to just flag the usage of the tmpfiles eclass. This can be done by removing the systemd dependency. Might be useful in the future to not have this kind of dependency breaking hacks for other packages too.

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.use
pothos: Can we do this in a follow up? I would first like to test this in combination with the clevis TPM support through the new Flatcar-helper service from the next Flatcar nightly.

ader1990 (approved):

PR #1768, opened 2024-03-18
Update ca-certificates in main from 3.98 to 3.99

flatcar-infra:
Subject says it all.


dongsupark (approved):
CI passed except for unrelated arm64 TPM failures.

PR #1771, opened 2024-03-18
sys-fs/multipath-tools: Disable realtime scheduling for multipathd

jepio:
# sys-fs/multipath-tools: Disable realtime scheduling for multipathd

Add a dropin that prevents the service from controlling its own scheduling. On newer Flatcar releases, service ordering seems to have changed and multipathd is able to configure realtime scheduling before systemd tries (and fails) to enable the cpu controller.  The dropin is meant to enforce sane behavior, and matches older Flatcar releases where cpu controller was enabled before multipathd tries (and fails) to configure realtime scheduling.

## How to use

TODO: mantle test case
TODO: changelog

## Testing done

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/multipath-tools
+++ b/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/multipath-tools
@@ -1,0 +1,13 @@
+cros_post_src_install_add_dropin() {
+	mkdir -p "${D}$(systemd_get_systemunitdir)/multipathd.service.d"
+	cat <<EOF >"${D}$(systemd_get_systemunitdir)/multipathd.service.d/flatcar.conf"
+[Service]
+# Multipathd sets itself to sched_rr with highest priority.
+# Cgroups2 doesn't support realtime processes outside the root cgroup,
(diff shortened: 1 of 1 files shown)


sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/multipath-tools
```diff
@@ -0,0 +1,13 @@
+cros_post_src_install_add_dropin() {
+	mkdir -p "${D}$(systemd_get_systemunitdir)/multipathd.service.d"
+	cat <<EOF >"${D}$(systemd_get_systemunitdir)/multipathd.service.d/flatcar.conf"
+[Service]
+# Multipathd sets itself to sched_rr with highest priority.
+# Cgroups2 doesn't support realtime processes outside the root cgroup,
+# if any such process exists then cpu controller can't be enabled.
+# This poses a bit of a dilemma.
+# Block realtime control for the process, but give it highest non-rt priority.
+RestrictRealtime=yes
```
ader1990, line 10: LGTM, worked on my local setup as well.

jepio:
Added a changelog.

jepio:
Not doing a mantle test case for now as this is unlikely to regress.

PR #1790, opened 2024-03-25
Update coreos-cloudinit to latest main branch

gabriel-samfira:
# Update coreos-cloudinit to the latest flatcar-master commit

This change disables user-configdrive.service on OpenStack, as coreos-cloudinit.service already runs on OpenStack when the system is not configured via ignition.

## How to use

Deploy a virtual machine on OpenStack with config drive, verify that the unit is disabled.

## Testing done

Built a fresh flatcar image and deployed it on a local OpenStack. Checked that `user-configdrive.service` was not started.

/build-image


pothos (approved):

pothos (approved):

pothos:
Can you squash before merging? The two arm64 tpm failures are expected and currently worked on, all good.

PR #1816, opened 2024-04-02
app-arch/xz-utils: Sync with Gentoo (revert to known-good)

pothos:
The 5.6 release contained a backdoor for SSH. The 5.6 release wasn't
    used in Flatcar and so far it seems that the backdoor wouldn't even be
    compiled for Gentoo. However, we so far don't know whether the other
    patches are malicious.
    
Revert to 5.4.2 as last known-good release (like Gentoo did).
    Note that the Flatcar main branch had a copy of the 5.6 ebuild but was
    not using it. Flatcar Alpha was on 5.4.6-r1, so before the backdoor but
    the malicious contributor did other changes of unclear impact part of
    this release. Similarly, Beta is on 5.4.5 and Stable is on 5.4.3. These
    should get downgraded, too.

## How to use

Backport to Stable, just to be sure


pothos:
> https://gitweb.gentoo.org/repo/gentoo.git/log/app-arch/xz-utils -> I think a portage stable update will catch these two commits upstream and there will be no need for a separate commit.

The sync works on a file base for now at least and we sync the whole directories but not commits.

ader1990:
> > https://gitweb.gentoo.org/repo/gentoo.git/log/app-arch/xz-utils -> I think a portage stable update will catch these two commits upstream and there will be no need for a separate commit.
> 
> The sync works on a file base for now at least and we sync the whole directories but not commits.

I was referring to the fact that the portage stable sync will contain the relevant file fixes.

krnowak:
I did similar change in my weekly updates branch. We downgrade to 5.4.2 which is the last version released by the previous maintainer.

pothos:
I want to backport this to Alpha

pothos:
> I want to backport this to Alpha

Updated now that we can backport to Stable and downgrade everything to 5.4.2 as last known-good release like Gentoo did.

krnowak (approved):

pothos:
Should we also rebuild the SDK? Doing so would need a note for the next person tagging the releases and starting the builds.

ader1990:
FYI ->Gentoo upstream blacklisted xz-utils 5.4.3 this way-> https://github.com/flatcar/scripts/pull/1788/files#diff-0e9c768a2a09eafc2d45cecaf99e8f68735e67877e7986ee387dab561be0f86dR25.

krnowak:
> FYI ->Gentoo upstream blacklisted xz-utils 5.4.3 this way-> https://github.com/flatcar/scripts/pull/1788/files#diff-0e9c768a2a09eafc2d45cecaf99e8f68735e67877e7986ee387dab561be0f86dR25.

I did the similar in the weekly updates PR: https://github.com/flatcar/scripts/pull/1788/files#diff-0e9c768a2a09eafc2d45cecaf99e8f68735e67877e7986ee387dab561be0f86d

This will be dropped though, on the follow-up weekly updates, when we inherit the mask from Gentoo.

PR #1827, opened 2024-04-03
qemu_template.sh: Add support for attaching a software TPM

pothos:
- qemu_template.sh: Add support for attaching a software TPM
    
    For testing TPM2-backed rootfs encryption it is handy to have a software
    TPM option for the qemu script.
    Add a flag for a software TPM with swtpm like kola also does. The user
    has to specify a folder for the secret state and this won't be removed
    because the same store should be able to be passed when booting the VM
    again after shutdown.
- vm_image_util.sh: Bump default VM memory to 2 GB
    
    While Flatcar itself runs fine with 1 GB, many workloads do not and
    having to debug this is time consuming when one forgets to bump the VM
    memory, e.g., in the Qemu script.
    Default to 2 GB as known-good setting for things like Kubernetes or
    setting up LUKS devices.
- qemu_template.sh: Allow parameters for VM image and memory
    
    When testing multiple images one always has to copy them to the
    expected file name, and when trying to run two VMs this means one needs
    to either use separate directories or modify the qemu script. One also
    needs to modify the qemu script to bump the memory for K8s or for LUKS.
    
    Support parameters for the VM image name and the VM memory.

(Note that the kola tests for qemu and vmware all use 2 GB or slightly more. Thus it makes sense to default to this as well in the release artifacts we provide to the users.)

## How to use

```
./flatcar_production_qemu.sh -T myswtpmdir
```

## Testing done

The above and verified that the swtpm process gets cleaned up even when qemu doesn't start, e.g., `./flatcar_production_qemu.sh -T swtpm-dir -- -unknown-arg`

Downloaded rendered template and checked that memory and image location settings work.

### DIFF
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -26,6 +26,8 @@ Options:
     -c FILE     Config drive as an iso or fat filesystem image.
     -a FILE     SSH public keys for login access. [~/.ssh/id_{dsa,rsa}.pub]
     -p PORT     The port on localhost to map to the VM's sshd. [2222]
+    -I FILE     Set a custom image file.
+    -M MB       Set VM memory in MBs.
     -s          Safe settings: single simple cpu and no KVM.
     -h          this ;-)
 
@@ -76,6 +78,12 @@ while [ $# -ge 1 ]; do
         -s|-safe)
             SAFE_ARGS=1
             shift ;;
+        -I|-image-file)
+            VM_IMAGE="$2"
+            shift 2 ;;
+        -M|-memory)
+            VM_MEMORY="$2"
+            shift 2 ;;
         -v|-verbose)
             set -x
             shift ;;
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -35,6 +35,8 @@ Options:
                 (see https://github.com/stefanberger/swtpm/wiki/Certificates-created-by-swtpm_setup).
     -R FILE     Set up pflash ro content, e.g., for UEFI (with -W).
     -W FILE     Set up pflash rw content, e.g., for UEFI (with -R).
+    -K FILE     Set kernel for direct boot used to simulate a PXE boot (with -R).
+    -R FILE     Set initrd for direct boot used to simulate a PXE boot (with -K).
     -s          Safe settings: single simple cpu and no KVM.
     -h          this ;-)
 
@@ -99,6 +101,12 @@ while [ $# -ge 1 ]; do
             shift 2 ;;
         -W|-pflash-rw)
             VM_PFLASH_RW="$2"
+            shift 2 ;;
+        -K|-kernel-file)
+            VM_KERNEL="$2"
+            shift 2 ;;
+        -R|-initrd-file)
+            VM_INITRD="$2"
             shift 2 ;;
         -v|-verbose)
             set -x
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -31,6 +31,8 @@ Options:
     -M MB       Set VM memory in MBs.
     -T DIR      Add a software TPM2 device through swtpm which stores secrets
                 and the control socket to the given directory.
+    -R FILE     Set up pflash ro content, e.g., for UEFI (with -W).
+    -W FILE     Set up pflash rw content, e.g., for UEFI (with -R).
     -s          Safe settings: single simple cpu and no KVM.
     -h          this ;-)
 
@@ -89,6 +91,12 @@ while [ $# -ge 1 ]; do
             shift 2 ;;
         -T|-tpm)
             SWTPM_DIR="$2"
+            shift 2 ;;
+        -R|-pflash-ro)
+            VM_PFLASH_RO="$2"
+            shift 2 ;;
+        -W|-pflash-rw)
+            VM_PFLASH_RW="$2"
             shift 2 ;;
         -v|-verbose)
             set -x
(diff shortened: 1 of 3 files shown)


build_library/qemu_template.sh
```diff
@@ -109,6 +123,29 @@ write_ssh_keys() {
     sed -e 's/^/ - /'
 }
 
+if [ -n "${SWTPM_DIR}" ]; then
+    mkdir -p "${SWTPM_DIR}"
+    if ! command -v swtpm >/dev/null; then
+        echo "$0: swtpm command not found!" >&2
```
ader1990, line 129: On Ubuntu, I needed to install some software packages and also initialize the swtpm:

```
sudo apt install swtpm swtpm-tools -y
sudo swtpm_setup --tpmstate $SWTPM_DIR \
    --create-ek-cert --create-platform-cert --lock-nvram --tpm2
```

Also, on Ubuntu, the only allowed socket path was in /run/swtpm/.

See: https://github.com/flatcar/Flatcar/issues/593#issuecomment-2025241150

ader1990: Inside the sdk_container docker process, I also tried to map the socket, but could not. Is there a way to map the socket?

From the docs, this should have worked, but it doesn't (no /run/swtpm/sock is present in the docker container):
```
./run_sdk_container -t -m /run/swtpm/sock:/run/swtpm/sock
```

ader1990: If we can test the TPM inside the container created by `./run_sdk_container`, there would be no need to break out back in the host OS to do the TPM testing.

pothos: For Fedora it works directly, maybe we can find out when `swtpm_setup` needs to be run? Otherwise I would document this in the help output for the flag.

ader1990: > swtpm_setup --tpmstate

The procedure is documented here https://manpages.ubuntu.com/manpages/jammy/man8/swtpm_setup.8.html

ader1990: And here: https://github.com/stefanberger/swtpm/wiki/Certificates-created-by-swtpm_setup#tpm-2-certificates
I do not know how Fedora already created the tpm store, if you change the directory, it will work as well?

pothos: For me this works with an empty folder (swtpm 0.8.1)

Maybe the config makes a difference?
```
$ cat /etc/swtpm_setup.conf 
# Program invoked for creating certificates
create_certs_tool= /usr/bin/swtpm_localca
create_certs_tool_config = /etc/swtpm-localca.conf
create_certs_tool_options = /etc/swtpm-localca.options
# Comma-separated list (no spaces) of PCR banks to activate by default
active_pcr_banks = sha256
```

pothos: I've pushed a vague note in the help output. Since this may be different for the various distributions we probably can't cover all of them - some people might also want to run this on MacOS and I don't know how if the setup would be required there and how it would look.

pothos:
Pushed one more change to also allow setting the pflash contents through a flag to switch between firmwares more easily, e.g., BIOS, UEFI, and UEFI with Secure Boot.

build_library/qemu_template.sh
ader1990: I think we can safely add this documentation on the flatcar-website, how to test the TPM functionality in isolation from the helper on Fedora and Ubuntu (this way, most of the generic OSses are covered).

ader1990: > For me this works with an empty folder (swtpm 0.8.1)
> 
> Maybe the config makes a difference?
> 
> ```
> $ cat /etc/swtpm_setup.conf 
> # Program invoked for creating certificates
> create_certs_tool= /usr/bin/swtpm_localca
> create_certs_tool_config = /etc/swtpm-localca.conf
> create_certs_tool_options = /etc/swtpm-localca.options
> # Comma-separated list (no spaces) of PCR banks to activate by default
> active_pcr_banks = sha256
> ```

On Ubuntu 22.04, swtpm 0.6.3:
```
cat /etc/swtpm_setup.conf
# Program invoked for creating certificates
create_certs_tool= /usr/lib/x86_64-linux-gnu/swtpm/swtpm-localca
create_certs_tool_config = /etc/swtpm-localca.conf
create_certs_tool_options = /etc/swtpm-localca.options
```

build_library/qemu_template.sh
```diff
@@ -26,6 +27,15 @@ Options:
     -c FILE     Config drive as an iso or fat filesystem image.
     -a FILE     SSH public keys for login access. [~/.ssh/id_{dsa,rsa}.pub]
     -p PORT     The port on localhost to map to the VM's sshd. [2222]
+    -I FILE     Set a custom image file.
+    -M MB       Set VM memory in MBs.
+    -T DIR      Add a software TPM2 device through swtpm which stores secrets
+                and the control socket to the given directory. This may need
+                some configuration first with 'swtpm_setup --tpmstate DIR ...'.
```
ader1990, line 34: Maybe add here:

```
                Add a software TPM2 device through swtpm which stores secrets
                and the control socket to the given directory. 
                See: https://github.com/stefanberger/swtpm/wiki/Certificates-created-by-swtpm_setup#tpm-2-certificates.'

```

pothos: Thanks for the link, that wiki is the best place to document more complex setups and how distros vary

ader1990 (approved):

PR #1861, opened 2024-04-09
build_library/grub.cfg: Enable TPM module by default

pothos:
For binding a secret to the OS we need TPM PCRs that measure the kernel and boot configuration. Used for:
https://github.com/flatcar/flatcar-website/pull/317

## How to use

Backport to Alpha

## Testing done

When booting with the UEFI script the first boot has PCR 8 and 9 set without any `grub.cfg` customization.

### DIFF
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -60,7 +60,7 @@ case "${FLAGS_target}" in
         CORE_NAME="core.img"
         ;;
     x86_64-efi)
-        CORE_MODULES+=( serial linux efi_gop efinet pgp http tftp )
+        CORE_MODULES+=( serial linux efi_gop efinet pgp http tftp tpm )
         CORE_NAME="core.efi"
         SBAT_ARG=( --sbat "${BOARD_ROOT}/usr/share/grub/sbat.csv" )
         ;;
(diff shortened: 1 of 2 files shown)


build_library/grub.cfg
```diff
@@ -6,6 +6,8 @@ set prefix=($root)/flatcar/grub
 # Load any and all video drivers.
 # Required under UEFI to boot Linux with a working console.
 insmod all_video
+# Load the TPM2 module to measure the boot code path and files into PCR 8+9
+insmod tpm
```
jepio, line 10: does it autoload if it's part of CORE_MODULES in `build_library/grub_install.sh`? we need it there for secure boot anyway

jepio:
```
GRUB will not measure its own @file{core.img} - it is expected that firmware
will carry this out. GRUB will also not perform any measurements until the
tpm module is loaded. As such it is recommended that the tpm module be built
into @file{core.img} in order to avoid a potential gap in measurement between
@file{core.img} being loaded and the tpm module being loaded.
```

pothos: Makes sense, if the module is loaded externally through the config file that's also external it's a chicken-egg problem for the measurement chain :D

pothos: Seems the config file is part of the efi binary as "memdisk", so that part would be measured. Anyway, added it to the CORE_MODULES and hope that it gets loaded without `insmod` then.

pothos: It works without `insmod`.

jepio (approved):

PR #1870, opened 2024-04-11
Upgrade Linux Kernel for flatcar-3913 from 6.6.21 to 6.6.26

flatcar-infra:
Subject says it all.


dongsupark:
See instead https://github.com/flatcar/scripts/pull/1932

PR #1906, opened 2024-04-15
.github: Disable debug output for image-changes

pothos:
The "set -x" bash debug output is very noisy and one doesn't even see the actual output in this noise.

### DIFF
--- a/.github/workflows/image_changes.sh
+++ b/.github/workflows/image_changes.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -x
+#set -x
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


krnowak (commented):
I'm not sure about this one. I don't even remember if we print anything meaningful to the image changes job's standard output. I think that the report you want should be available as a zipped artifact to download.

pothos:
> I'm not sure about this one. I don't even remember if we print anything meaningful to the image changes job's standard output. I think that the report you want should be available as a zipped artifact to download.

Checked this, yes, the actual output is just a few error about not finding certain OEM content reports because they weren't built. The report is archived and not printed.

PR #1916, opened 2024-04-17
dev-lang/rust: Drop alternative patch

pothos:
The patch https://github.com/rust-lang/rust/pull/119582 got merged some releases ago as alternative to the patch
https://github.com/rust-lang/rust/pull/119445 we included in the ebuild file. We currently use both because the unneeded patch got changed to apply again but it's not needed and may cause strange behavior. Therefore, drop the alternative patch.

## Testing done

On [Jenkins](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1424/cldsv/) (Packages and Images worked, VM failed unrelated) because the SDK update in GitHub Actions is not enough


dongsupark (approved):
LGTM if CI passed

PR #1937, opened 2024-04-18
coreos-base/coreos-init: drop daemon-reload command

tormath1:
With the recent systemd upgrade we do not need to reload the systemd daemon anymore.

This pulls:
- https://github.com/flatcar/init/pull/96
---
- [ ] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3877/cldsv/

### DIFF
--- a/build_sysext
+++ b/build_sysext
@@ -277,6 +277,7 @@ all_fields=(
   'ID=flatcar'
   "${version_field}"
   "ARCHITECTURE=${ARCH}"
+  "EXTENSION_RELOAD_MANAGER=1"
 )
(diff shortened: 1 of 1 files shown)


pothos:
Some more problems, even if the Flatcar extensions would set `` they still need to use `.upholds/` symlinks instead of `.wants`.

ZFS: `multi-user.target.wants/zfs.target` → `multi-user.target.upholds/zfs.target`
Containerd: `multi-user.target.wants/containerd.service`
AWS: `multi-user.target.wants/ntpd.service`
Azure: `multi-user.target.wants/chronyd.service`
GCE: `multi-user.target.wants/ntpd.service`

One way is to patch `WantedBy=` to be `UpheldBy` and then the enablement with `systemctl enable` would work.

PR #1948, opened 2024-04-22
Add CONFIG_PPS_CLIENT_GPIO=m

space55:
Closes [1430](https://github.com/flatcar/Flatcar/issues/1430)

# Add CONFIG_PPS_CLIENT_GPIO=m

This will enable the PPS GPIO kernel module to be used in arm64 builds. It can be done in the common config as well, however, I personally would anticipate that arm64 instances are much more likely to have a use for PPS GPIO than x86, and as such, shouldn't be needed in the common config. Please let me know if it would be desirable to have this setting elsewhere.

## How to use

System without this change
```bash
$ modprobe pps-gpio
modprobe: FATAL: Module pps-gpio not found in directory /lib/modules/6.1.85-flatcar
```

System with this change
```bash
$ modprobe pps-gpio
$ # Note: no output
```

Alternatively:
```bash
$ lsmod | grep gpio
pps_gpio               16384  1
```

## Testing done

Currently need an image of Flatcar to be built by CI for me to test


pothos:
The image is available in the arm64-generic-image.zip https://github.com/flatcar/scripts/actions/runs/8792407190?pr=1948
If you already have an install on an arm64 board, you can also extract the update payload from `
arm64-test-update.zip` and switch to it with `flatcar-update` with `-D -V 1.2.3 -P FILE`.

space55:
Confirmed that the image does, in fact, still boot, and the `pps-gpio` module is loaded when the `modprobe pps-gpio` command is executed.

pothos (approved):

PR #1950, opened 2024-04-22
azure updates

jepio:
# azure updates
* refresh waagent patch to fix error that happens when deprovisioning
* add compat symlink to old oem waagent.conf path, because some test system of the waagent team depends on this
* fully disable interface restarting. this just causes slowdowns and test flakiness on flatcar, but we configure hostname from initrd so we don't need this.
* add azure-nvme-utils to base OS, for the next instance generation with os and temp disk on NVMe. this can't be in a sysext because udev rules

## How to use

## Testing done

Testing here: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/22112/cldsv/


sdk_container/src/third_party/coreos-overlay/app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
```diff
@@ -419,6 +421,7 @@ index 00000000..d0d6f7c8
 +# /usr/share/flatcar/etc.
 +#
 +ExecStartPre=/bin/bash -c 'if [[ ! -e /etc/waagent.conf ]]; then ln -sf ../usr/share/waagent/waagent.conf /etc/waagent.conf; fi'
++ExecStartPre=/bin/bash -c 'if [[ ! -e /usr/share/oem/waagent.conf ]]; then ln -sf ../../../etc/waagent.conf /usr/share/oem/waagent.conf; fi'
```
pothos, line 424: That relative link is there on purpose because it also happens to work for `/oem/waagent.conf` or would an absolute link be better?

sdk_container/src/third_party/coreos-overlay/coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -81,6 +81,7 @@ RDEPEND="${RDEPEND}
 
 # sys-devel/gettext: it embeds 'envsubst' binary which is useful for simple file templating.
 RDEPEND="${RDEPEND}
+	app-admin/azure-nvme-utils
```
pothos, line 84: Maybe a changelog entry would be good under `changes/` to see when this is available in Stable and also understand how it's used/what the effects are.

sdk_container/src/third_party/coreos-overlay/app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
jepio: oh you're right, i hadn't thought of it. it's probably better to make this an absolute link

sdk_container/src/third_party/coreos-overlay/app-admin/azure-nvme-utils/azure-nvme-utils-0.0.0-r1.ebuild
```diff
@@ -0,0 +1,28 @@
+# Copyright 2024 Gentoo Authors
```
krnowak, line 1: That probably should say Flatcar maintainers instead.

sdk_container/src/third_party/coreos-overlay/app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
krnowak: Do we need to care about `/usr/share/oem`? I think that just `/oem` should work, since `/usr/share/oem` is a symlink to `/oem`.

sdk_container/src/third_party/coreos-overlay/coreos-base/coreos/coreos-0.0.1.ebuild
jepio: added an entry

sdk_container/src/third_party/coreos-overlay/app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
jepio: switched to an absolute symlink in /oem/waagent.conf

jepio:
@krnowak @pothos I don't have time to undo this, but I would revert this: https://github.com/flatcar/scripts/pull/1950/commits/9556c7f94b932ded3658738d616c2f666a364ba8
This would break the vm changing hostname, which is common on Azure.

pothos:
> This would break the vm changing hostname, which is common on Azure.

Do you mean live changing through the Azure API without a reboot? Sounds like previously this resulted in a transient hostname being set, i.e., `/etc/hostname` would still use the old one?

Not sure if it's good to revert because preventing the network restart was also meant as bug fix, or? What we could try is to see whether we can trigger a new DHCP request instead of the if-up-down hack. My naive thinking is that the DHCP request would keep the same IP address and be a no-op except that it transfers the hostname to systemd-networkd. I guess `networkctl renew IFNAME` or `networkctl forcerenew IFNAME` should work.

pothos:
> Do you mean live changing through the Azure API without a reboot? Sounds like previously this resulted in a transient hostname being set, i.e., /etc/hostname would still use the old one?

Looks like it's the other way round, changing internally and waagent noticed this and issues the if-up-down to propagate to the read-only VM properties.

> I guess `networkctl renew IFNAME` or `networkctl forcerenew IFNAME` should work.

Apparently they do not but `networkctl reconfigure IFNAME` worked. We should change the patch to use this.

pothos:
Opened a PR https://github.com/flatcar/scripts/pull/1978

PR #1951, opened 2024-04-23
coreos/config zfs-kmod: fix Kernel version issue for build_sysext

dongsupark:
In `coreos/config/env/sys-fs/zfs-kmod`, explicitly set `KV_{FULL,MAJOR,MINOR,PATCH}`, which are unfortunately not available from context of `build_sysext`.

Without setting them, build_sysext complains like below, which results in subsequent zfs CI tests failing due to missing files like `System.map` and module deps.

```
System.map for kernel 6.6.21-flatcar was not found, may be due to the
built kernel sources no longer being available and lacking the fallback:
```

Also in `profiles/coreos/base/profile.bashrc` add a hook for pre_pkg_postinst to explicitly set `KV_FULL` to a correct Kernel version, not the version of build host, but the version given by ebuilds. This is necessary for build_sysext to detect correct Kernel versions, when building Kernel modules like zfs-kmod.

Without that, pkg_postinst step results into running `depmod` a wrong Kernel version, so that subsequent zfs CI tests fail to run.

Note: this PR includes an update to Kernel 6.6.23 for now, to uncover the issue of CI failure. Will remove that Kernel update right before merging.

## Testing done

TBD


jepio:
Thanks for the analysis, this gave me some hints as to where to look. It looks like this is what causes the problems:
```bash
linux-info_pkg_setup() -> linux-info_get_any_version():

linux-info_get_any_version() {
        if ! use kernel_linux; then
                die "${FUNCNAME}() called on non-Linux system, please fix the ebuild"
        fi

        if [[ ${MERGE_TYPE} == binary && -z ${LINUX_INFO_BINARY_RESET} ]]; then
                unset KV_FULL _LINUX_CONFIG_EXISTS_DONE KV_OUT_DIR
                LINUX_INFO_BINARY_RESET=1
        fi

        if [[ ${MERGE_TYPE} != binary ]] && ! get_version; then
                ewarn "Unable to calculate Linux Kernel version for build, attempting to use running version"
        fi

        if [[ -z ${KV_FULL} ]] && ! get_running_version; then
                die "Unable to determine any Linux Kernel version, please report a bug"
        fi
}
```

MERGE_TYPE is binary when building the sysext.

So how about this minimal fix:
```patch
diff --git a/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/zfs-kmod b/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/zfs-kmod
index 55536d82a8..c5d909643a 100644
--- a/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/zfs-kmod
+++ b/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-fs/zfs-kmod
@@ -1,2 +1,7 @@
 : ${MODULES_ROOT:=$(echo ${SYSROOT}/lib/modules/*)}
 KERNEL_DIR="${MODULES_ROOT}/build"
+
+cros_pre_pkg_setup_kernel_version() {
+  LINUX_INFO_BINARY_RESET=1
+  get_version
+}
```

dongsupark:
Thanks.
Your change is not exactly what I wanted to achieve, but have simply replaced the PR with your suggestion, as I am also fan of minimal changes.
Let's see if CI passes.

dongsupark:
CI passed http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3916/cldsv/

changelog/bugfixes/2024-04-24-kernel-version-zfs-kmod.md
```diff
@@ -0,0 +1 @@
+- Fixed CI failure of zfs sysext with Kernel 6.6 by setting LINUX_INFO_BINARY_RESET=1 to prevent KV_FULL and KV_OUT_DIR from being unset. ([scripts#1951](https://github.com/flatcar/scripts/pull/1951))
```
pothos, line 1: We don't need this entry because it's not fixing a bug of the last release.

dongsupark: Done

dongsupark:
cherry-picked to flatcar-3913, flatcar-3941

PR #1952, opened 2024-04-24
Migrate release AMI gc to ci-automation

pothos:
The deletion of old release AMIs was still relying on the previous Jenkins setup. Move this to the ci-automation folder.

## Testing done

It worked on a manual run in Jenkins: `Pruning complete: {total:508 kept:42 skipped:420 recentlyUsed:4 softDeleted:42 deleted:0}` and similar showed up 8 times.

### DIFF
--- a/ci-automation/garbage_collect.sh
+++ b/ci-automation/garbage_collect.sh
@@ -256,6 +256,7 @@ function _garbage_collect_impl() {
     docker run --pull always --rm --net host \
       --env AZURE_AUTH_CREDENTIALS --env AZURE_PROFILE \
       --env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY \
+      --env AWS_CREDENTIALS \
       --env DIGITALOCEAN_TOKEN_JSON \
       --env EQUINIXMETAL_KEY --env EQUINIXMETAL_PROJECT \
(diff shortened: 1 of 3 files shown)


ci-automation/garbage_collect.sh
```diff
@@ -256,6 +256,7 @@ function _garbage_collect_impl() {
     docker run --pull always --rm --net host \
       --env AZURE_AUTH_CREDENTIALS --env AZURE_PROFILE \
       --env AWS_ACCESS_KEY_ID --env AWS_SECRET_ACCESS_KEY \
+      --env AWS_CREDENTIALS \
```
krnowak, line 259: I suppose that these are yet different stuff from `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`?

jenkins/README.md
krnowak, line 1: We probably could just drop this whole directory, no?

krnowak (approved):
Two questions, otherwise looks good.

jenkins/README.md
pothos: We would have to migrate the `jenkins/systemd-run-wrap.sh` script which is used by `jenkins-os/container/` but otherwise yes, possible.

PR #1953, opened 2024-04-24
vm_image_util: produce qcow2 images for Scaleway

tormath1:
For importing Scaleway images, extension needs to be '.qcow2'

See: https://www.scaleway.com/en/docs/compute/instances/how-to/snapshot-import-export-feature/
> Make sure that the QCOW / QCOW2 image file you want to import,
> uses the file extension .qcow or .qcow2 to avoid issues while importing the image.

## Testing done

- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/3917/cldsv/

To be backported on Alpha and Beta.

### DIFF
--- a/changelog/changes/2024-04-24-scaleway.md
+++ b/changelog/changes/2024-04-24-scaleway.md
@@ -1,0 +1,1 @@
+- Scaleway: images are now provided directly as `.qcow2` to ease the import on Scaleway ([scripts#1953](https://github.com/flatcar/scripts/pull/1953))


pothos (commented):
For qcow2 we don't need the bz2 wrapping when introducing a new image format. We can add a case with `COMPRESSION_FORMAT=none` for scaleway in `ci-automation/vms.sh`

The change looks good otherwise: As scaleway images are not in Stable yet I think we can switch the output file behavior.

build_library/vm_image_util.sh
```diff
@@ -421,6 +425,11 @@ _dst_path() {
 # Get the proper disk format extension.
 _disk_ext() {
     local disk_format=$(_get_vm_opt DISK_FORMAT)
+    local disk_extension=$(_get_vm_opt DISK_EXTENSION)
+    if [[ ! -z "${disk_extension}" ]]; then
```
krnowak, line 429: Quotes are not necessary in `[[` (as opposed to `[`), and `! -z` is basically `-n`:

```suggestion
    if [[ -n ${disk_extension} ]]; then
```

krnowak (approved):
One nitpick, otherwise looks good.

PR #1954, opened 2024-04-24
grub cfg tweaks

jepio:
# grub cfg tweaks

These changes are not user visible but slightly improve our grub configuration:
- pass a useful hint to the memdisk grub config so that it searches the most likely boot partition path first
- include btrfs grub module in CORE_MODULES, since btrfs is used by the oem partition

## How to use

## Testing done


pothos (approved):

build_library/grub_install.sh
```diff
@@ -144,7 +144,7 @@ info "Generating ${GRUB_DIR}/load.cfg"
 # The existing $root value is given as a hint so it is searched first.
```
krnowak, line 144: This comment became outdated with your change.

jepio: thanks, updated the comment

PR #1955, opened 2024-04-25
Shrink /boot by removing duplicate and unused grub binaries/modules

jepio:
# Shrink /boot

Remove grub modules from disk that are assembled into the grub executable.
Remove grub executables that are copies that are not used in the boot path.

This should have no user visible impact.

## How to use

## Testing done

### DIFF
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -183,6 +183,12 @@ sudo grub-mkimage \
     --output "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}" \
     "${CORE_MODULES[@]}"
 
+if [[ "${FLAGS_target}" != i386-pc ]]; then
+    for mod in "${CORE_MODULES[@]}"; do
+      sudo rm "${ESP_DIR}/${GRUB_DIR}/${mod}.mod"
+    done
+fi
+
 # Now target specific steps to make the system bootable
 case "${FLAGS_target}" in
     i386-pc)
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -204,8 +204,9 @@ case "${FLAGS_target}" in
             sudo sbsign --key /usr/share/sb_keys/shim.key \
                 --cert /usr/share/sb_keys/shim.pem \
                 "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}"
-            sudo cp "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}.signed" \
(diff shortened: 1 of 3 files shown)


build_library/grub_install.sh
```diff
@@ -183,6 +183,12 @@ sudo grub-mkimage \
     --output "${ESP_DIR}/${GRUB_DIR}/${CORE_NAME}" \
     "${CORE_MODULES[@]}"
 
+if [[ "${FLAGS_target}" != i386-pc ]]; then
```
pothos, line 186: Since they become part of the `core.img` it should be possible to remove this, or? I think we can try this out and either BIOS will boot or not.

build_library/grub_install.sh
```diff
@@ -183,6 +183,12 @@ sudo grub-mkimage \
...
+    for mod in "${CORE_MODULES[@]}"; do
+      rm "${ESP_DIR}/${GRUB_DIR}/${mod}.mod"
```
ader1990, line 188: `Command 'rm "${ESP_DIR}/${GRUB_DIR}/${mod}.mod"' exited with nonzero code: 1`
`rm: cannot remove '/tmp/tmp.WM8QCKnE6l/flatcar/grub/arm64-efi/normal.mod': Permission denied`
 -> looks that this commands needs also a `sudo`?

ader1990:
This is the current difference on AMD64:

```
this PR amd64 /boot 48% used:
used 61782
free 67258

current alpha amd64 /boot 50% used:
used 63483
free 65556
```

build_library/grub_install.sh
ader1990: Added a sudo, and the build now works.

build_library/grub_install.sh
ader1990: I will also try this out and see if BIOS works or not.

ader1990: With @pothos suggestion to remove also the amd64 parts:

```
used 61681
free 67358

versu before:

used 61782
free 67258
```

ader1990:
Hello @jepio, I have verified that the image boots on both Gen 1/Gen 2 for Hyper-V AMD64.

ader1990:
Verified on ARM64 and the free space looks good too:

```
Filesystem               1K-blocks Used Available Use% Mounted on
/dev/nvme0n1p1    129039   61556     67483  48% /boot
```

ader1990:
> @ader1990 keep in mind that existing nodes will still have these files so this PR is just a cleanup, and doesn't let us ship a larger kernel.

Definitely, this PR will help a little to not go beyond the 50% logical threshold in the near future with things like natural Linux kernel or other portage stable minor version updates.

PR #1962, opened 2024-04-29
image: add kubevirt image build

ader1990:
As Ignition supports KubeVirt, add a custom oem for it and also the required parts to be able to build an image in .qcow2 format and .gz compression.

Fixes: https://github.com/flatcar/Flatcar/issues/1358

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/coreos-metadata.service
+++ b/sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/coreos-metadata.service
@@ -21,6 +21,8 @@ ConditionKernelCommandLine=|flatcar.oem.id=scaleway
 
 ConditionKernelCommandLine=|flatcar.oem.id=hetzner
 
+ConditionKernelCommandLine=|flatcar.oem.id=kubevirt
+
 Description=Flatcar Metadata Agent
 
 [Service]
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -19,6 +19,7 @@ VALID_IMG_TYPES=(
     hyperv
     hyperv_vhdx
     iso
+    kubevirt
     openstack
     openstack_mini
     packet
@@ -51,6 +52,7 @@ VALID_OEM_PACKAGES=(
     gce
     hetzner
     hyperv
+    kubevirt
     openstack
     packet
     qemu
@@ -314,6 +316,14 @@ IMG_scaleway_OEM_PACKAGE=common-oem-files
 IMG_scaleway_OEM_USE=scaleway
 IMG_scaleway_OEM_SYSEXT=oem-scaleway
 IMG_scaleway_DISK_EXTENSION=qcow2
+
+## kubevirt
+IMG_kubevirt_DISK_FORMAT=qcow2
+IMG_kubevirt_DISK_LAYOUT=vm
+IMG_kubevirt_OEM_PACKAGE=common-oem-files
+IMG_kubevirt_OEM_USE=kubevirt
+IMG_kubevirt_OEM_SYSEXT=oem-kubevirt
+IMG_kubevirt_DISK_EXTENSION=qcow2
 
 ###########################################################
 
--- a/changelog/changes/2024-04-29-kubevirt-images.md
+++ b/changelog/changes/2024-04-29-kubevirt-images.md
@@ -1,0 +1,1 @@
+- Added KubeVirt qcow2 image for amd64/arm64 ([flatcar/scripts#1962](https://github.com/flatcar/scripts/pull/1962))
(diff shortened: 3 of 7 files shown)


ci-automation/vms.sh
```diff
@@ -140,6 +140,8 @@ function _vm_build_impl() {
             COMPRESSION_FORMAT="zip"
         elif [[ "${format}" =~ ^(scaleway)$ ]];then
             COMPRESSION_FORMAT="none"
+        elif [[ "${format}" =~ ^(kubevirt)$ ]];then
+            COMPRESSION_FORMAT="gz"
```
pothos, line 144: The qcow2 is already compressed, do we need the additional gz wrapper or can we use `none`?

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/coreos-metadata.service
```diff
@@ -21,6 +21,8 @@ ConditionKernelCommandLine=|flatcar.oem.id=scaleway
```
pothos, line 21: Besides the metadata afterburn also supports the hostname and the ssh keys for this provider. We should add this in `bootengine` (`flatcar-metadata-hostname.service`) and `init` (`sshkeys.service`)

ci-automation/vms.sh
ader1990: I would like to be visibile that the compression format has been applied, in case we add a raw image at some time too.
Also, it is an implementation detail that qcow2 can also support compression inside, but the naming is not implied.

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/coreos-metadata.service
ader1990: I added this in the case of fallback, but this image should be used with Ignition userdata format. The default flatcar production image with no oem works in the case of openstack metadata/userdata format on KubeVirt, as Kubevirt is just a config drive implementation of OpenStack metadata.

ader1990:
Tested locally on my Kubevirt Environment and the image works as expected.

Workflow:

1. Deply Kubernetes with CDI and KubeVirt (not the scope of this how to)
2. Use virtctl to create a PVC from the image:

```bash
virtctl image-upload pvc flatcar-kubevirt-pvc --size=10Gi --image-path flatcar_production_kubevirt_image.qcow2
```

3. Create KubeVirt VM

```yaml
apiVersion: kubevirt.io/v1
kind: VirtualMachine
metadata:
  generation: 1
  labels:
    kubevirt.io/os: linux
  name: vm-flatcar
spec:
  running: true
  template:
    metadata:
      creationTimestamp: null
      labels:
        kubevirt.io/domain: vm-flatcar
    spec:
      architecture: amd64
      nodeSelector:
        kubernetes.io/arch: amd64
      domain:
        cpu:
          cores: 2
        devices:
          disks:
          - disk:
              bus: virtio
            name: disk0
          - cdrom:
              bus: sata
              readonly: true
            name: cloudinitdisk
        machine:
          type: q35
        resources:
          requests:
            memory: 1024M
      volumes:
      - name: disk0
        persistentVolumeClaim:
          claimName: flatcar-kubevirt-pvc
      - cloudInitConfigDrive:
          userData: |
            {"ignition":{"version":"3.3.0"},"kernelArguments":{"shouldExist":["flatcar.autologin"]},"passwd":{"users":[{"name":"core","passwordHash":"$6$sn3ZSJJJln5JkAZb$VDTKzLpCyjlEe7Kh0DKjOnEawkkOoi0tOKVbcCv0FIWSf3u9Y1p1I5YdJJ5L8uDmmMvO2CBlmJZNdxFuekjjE1"}]}}
        name: cloudinitdisk

```

4. Connect to VM using core/foo password

ci-automation/vms.sh
ader1990: The CDI https://kubevirt.io/user-guide/operations/containerized_data_importer/ supports RAW/QCOW2 and GZ/XZ.
I ll need to add a how to on the website as well where to document that the compression format is supported.

sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/files/coreos-metadata.service
pothos: About coroes-cloudinit: The mount unit trigger will only run when Ignition didn't run. There is no explicit enablement because there is no `--oem kubevirt` flag for coreos-cloudinit.

pothos (approved):
Thanks

ader1990:
Tested locally on KubeVirt using Ignition and Bash userdata, and all works as expected.

pothos (approved):

PR #1964, opened 2024-04-29
Podman sysext

JeWe37:
# Podman sysext

Initial work on getting a podman sysext to work, see flatcar/Flatcar#112. For now just includes the basic ebuilds required to build the sysext.

A few minor changes to the infrastructure were made, it's now possible to pass the debug flag to emerge when building packages and a sysext can now be built based on multiple packages. The latter was useful as I wanted to include passt for pasta rootless networking in the podman sysext.

Next I intend to look into what a mangle script should include, particularly based on the prior work on podman sysexts based on the docker builder.

This is very much my first time working with this repo, as such I am not familiar with the context. The ebuilds were pulled directly from gentoo. For most, masking on amd64 was disabled and were otherwise kept as-is. For the podman ebuild in particular, gpgme was added as a BDEPEND, as otherwise building the binary package would fail. There appear to be scripts responsible for updating the ebuilds, how does one integrate with those?

## How to use

Inside the SDK container, on a host with a kernel that is not too recent(otherwise the ZFS sysext won't build):
```
./build_packages
./build_image sysext
```

## Testing done

Next to none. This is very much a draft.


pothos:
Thanks, for the config or service files we might need to have them be under `/usr` if that isn't already the case. If `/usr` is not usable for config files, as is the case for `/etc/subuid`, `/etc/subgid`, we would need to create them on load with a helper service. See `build_library/sysext_mangle_*`.

JeWe37:
Alright, latest draft takes care of the mangle script. I still want to do more some more testing but not today.

The mangle script in its current form sets up the UID/GID mapping via systemd-tmpfiles which seemed like the neatest solution, alongside the required config files. It also ignores podman* bridge networks, sets up dbus which podman wants for cgroup management, and cleans up extraneous files(only required for development etc.).

build_library/sysext_mangle_flatcar-podman
```diff
@@ -0,0 +1,54 @@
...
+
+mkdir -p ./usr/lib/systemd/network
+
+cat <<EOF >>./usr/lib/systemd/network/50-podman.network
+[Match]
+Type=bridge
+Name=podman*
+
+[Link]
+Unmanaged=yes
```
pothos, line 28: Is this really needed or a precaution? The `zz-default.network` should not apply DHCP on bridges.

build_library/sysext_mangle_flatcar-podman
```diff
@@ -0,0 +1,54 @@
...
+EOF
+
+cat <<EOF >>./usr/lib/systemd/user/dbus.service
```
pothos, line 31: The Podman extension isn't the best place for this, I think we should rather add this to the base image.

build_library/sysext_mangle_flatcar-podman
JeWe37: Apparently just a precaution-again me working off of the prior work on this, where this was redundant as well I suppose. Will be removed in the next revision.

build_library/sysext_mangle_flatcar-podman
pothos: Thanks for finding this. I think it's ok to enable it. The user session is also used by the systemd user instance.

JeWe37:
Regarding the BDEPEND issue for podman, that is arguably an upstream issue, filed that as a [bug](https://bugs.gentoo.org/930982). 

For this PR I used Gentoo unstable versions, is that acceptable? Or should I revert to using only the currently stable versions? I also may need to unmask arm64 separately, I can't test on that myself.

JeWe37:
Seems to all be working nicely now. I went through and tested what I can think of, running podman containers in rootful and rootless modes works, so do quadlet and pasta networking.

JeWe37:
My mistake it seems I committed the removal of gpgme from the BDEPEND from my testing. I'll fix that in a bit.

pothos:
Looks great, you can ignore the test failure here, fix in https://github.com/flatcar/mantle/pull/530 (The other two are failing due to the GitHub Action setup).

pothos (approved):
Great work!

Can you also create a file `changelog/changes/2024-05-03-podman.md` with
```
- Provided a Podman Flatcar extension as optional systemd-sysext image with the release. Write 'podman' to `/etc/flatcar/enabled-sysext.conf` through Ignition and the sysext will be installed during provisioning ([scripts#1964](https://github.com/flatcar/scripts/pull/1964))
```
and when done, please squash all commits into one.

After merging we should also add a kola tests and docs for this, if you want, you can try this but up to you :)

JeWe37:
Squashed and changelog added! Commit message is hopefully in the right format.

PR #1965, opened 2024-04-30
Upgrade Linux Kernel for flatcar-3815 from 6.1.87 to 6.1.89

flatcar-infra:
Subject says it all.


jepio:
@pothos @krnowak @dongsupark this requires reverting the same patch that Dongsu reverted in main.

dongsupark:
@jepio Thanks. Fix is in another PR https://github.com/flatcar/scripts/pull/1971.

dongsupark:
See instead https://github.com/flatcar/scripts/pull/1971

PR #1978, opened 2024-05-03
app-emulation/wa-linux-agent: Use networkctl to propagate hostname

pothos:
The if-up-down to trigger the DHCP request causes problems. It's better
    to directly ask systemd-networkd to issue the request. It seems that
    one needs to use "reconfigure" instead of "forcerenew", so I went with
    only that instead of somehow trying to see if "forcerenew" has an
    effect.

## How to use

For this comment https://github.com/flatcar/scripts/pull/1950#issuecomment-2092741939

## Testing done

Used `flatcar-update -P … -E …` to switch to the built image, and checked that waagent triggers the DHCP renewal:
```
May 03 14:06:56 newname systemd-hostnamed[2482]: Hostname set to <newname> (static)
May 03 14:06:56 newname systemd-resolved[1538]: System hostname changed to 'newname'.
May 03 14:06:56 newname sudo[2479]: pam_unix(sudo:session): session closed for user root
May 03 14:07:06 newname waagent[2401]: 2024-05-03T14:07:06.006518Z INFO EnvHandler ExtHandler EnvMonitor: Detected hostname change: hello2 -> newname
May 03 14:07:06 newname waagent[2401]: 2024-05-03T14:07:06.007770Z INFO EnvHandler ExtHandler Examine /proc/net/route for primary interface
May 03 14:07:06 newname waagent[2401]: 2024-05-03T14:07:06.007969Z INFO EnvHandler ExtHandler Primary interface is [eth0]
May 03 14:07:06 newname systemd-networkd[1736]: eth0: found matching network '/usr/lib/systemd/network/zz-default.network', based on potentially unpredictable interface name.
May 03 14:07:06 newname systemd-networkd[1736]: eth0: Configuring with /usr/lib/systemd/network/zz-default.network.
May 03 14:07:06 newname systemd-networkd[1736]: eth0: DHCP lease lost
May 03 14:07:06 newname systemd-networkd[1736]: eth0: DHCPv6 lease lost
May 03 14:07:06 newname systemd-networkd[1736]: eth0: DHCPv4 address 10.0.0.4/24, gateway 10.0.0.1 acquired from 168.63.129.16
```
There is no `Link DOWN`, `Lost carrier` anymore for `eth0` and  `enP52305s1`.

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
+++ b/sdk_container/src/third_party/coreos-overlay/app-emulation/wa-linux-agent/files/0001-flatcar-changes.patch
@@ -1,4 +1,4 @@
-From 7382c63bb2c90a1173393faf093002341f830a09 Mon Sep 17 00:00:00 2001
+From 948c6075656fde25703ba402f8cd94715feaa774 Mon Sep 17 00:00:00 2001
 From: Krzesimir Nowak <knowak@microsoft.com>
 Date: Mon, 27 Feb 2023 15:59:21 +0100
 Subject: [PATCH] flatcar changes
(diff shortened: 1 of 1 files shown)


jepio (approved):
love the approach

PR #2012, opened 2024-05-29
coreos-overlay/profiles: Add keywords for arm64 sdk

jepio:
# Add arm64 keywords for sdk packages
- fero-client was previously keyworded but a revbump due to a dependency change resulted in the keyword not matching any longer
- ccid is a new dep that is still marked ~arm64 in the Gentoo tree.

## How to use

## Testing done

Testing with bootstrap_sdk when starting from dev container tarball as seed. These missing keywords show up in stage4.

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
+++ b/sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
@@ -14,6 +14,9 @@
 # Seems to be the only available ebuild in portage-stable right now.
 =app-crypt/adcli-0.9.2 ~amd64 ~arm64
 
+# Needed by arm64-native SDK
+=app-crypt/ccid-1.5.4 ~arm64
+
 # The only available ebuild (from GURU) has ~amd64 and no keyword for arm64 yet.
 =app-crypt/clevis-19-r1 **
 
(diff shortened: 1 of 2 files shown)


sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```diff
@@ -27,7 +29,7 @@
 =app-shells/bash-5.2_p26 ~amd64 ~arm64
 
 # No keyword for arm64 yet.
-=coreos-devel/fero-client-0.1.1 **
+~coreos-devel/fero-client-0.1.1 **
```
chewi, line 32: Shouldn't we just add the keyword to fero-client instead?

krnowak: Yeah, I'd say so too, it's our ebuild. On a tangent - we probably we should fork fero-client to flatcar org, because upstream is archived.

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```diff
@@ -14,6 +14,8 @@
 # Seems to be the only available ebuild in portage-stable right now.
 =app-crypt/adcli-0.9.2 ~amd64 ~arm64
 
+=app-crypt/ccid-1.5.4 ~arm64
```
krnowak, line 17: Please add a comment above it like `# Needed by arm64-native SDK.`.

jepio: Done

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
jepio: Moved keyword to the ebuild and deleted this entry from the profile

krnowak (approved):

PR #2014, opened 2024-05-29
netperf: update to latest license

ader1990:
Use License from the github repo:
https://github.com/HewlettPackard/netperf/blob/master/COPYING

Fixes: https://github.com/flatcar/Flatcar/issues/1459

## How to use

## Testing done

### DIFF
--- a/sdk_container/src/third_party/portage-stable/licenses/netperf
+++ b/sdk_container/src/third_party/portage-stable/licenses/netperf
@@ -1,43 +1,21 @@
-
- 
-              Copyright (C) 1993 Hewlett-Packard Company
-                         ALL RIGHTS RESERVED.
- 
-  The enclosed software and documentation includes copyrighted works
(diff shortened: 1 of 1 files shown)


ader1990:
> I think if we're doing this it will be automatically reverted at the next `licenses` update from the package automation:
> 
> https://github.com/flatcar/scripts/blob/ec192a3708a9b784890dbe495bca1155c6e2da99/.github/workflows/portage-stable-packages-list#L398
> 
> 
> What about updating this directly on Gentoo overlay?

for some reason, I saw that file was 12y old and thought that the folder was not synced periodically. I ll open a PR on gentoo upstream in this case.

ader1990:
See: https://github.com/flatcar/Flatcar/issues/1459#issuecomment-2140076721

The License is correct, as Gentoo and Flatcar do not use the source code with the updated license.

PR #2015, opened 2024-05-31
Update netperf to MIT licensed version

jepio:
# Update netperf to MIT licensed version

Create a _preXXXX ebuild that references the latest commit in the github repo from netperf, which is MIT licensed.

## How to use

## Testing done

Build tested.


chewi:
I know we said we'd probably drop it, but I've submitted [an upstream Gentoo fix](https://github.com/gentoo/gentoo/pull/36917) for this anyway.

PR #2016, opened 2024-05-31
changelog: disable DHCP RapidCommit for CloudSigma OEM

daMupfel:
# Disable DHCP RapidCommit for CLoudSigma OEM

This is related to the following issue: https://github.com/flatcar/Flatcar/issues/1438

## How to use

Build OEM beta image and upload it to CloudSigma. Create a host with said image. 

## Testing done

I tested this by manually applying the OEM config to the /oem/cloud-config.yml file and rebooting.

Note: This is my idea how to fix this issue, i am unsure if this is the desired solution and if this is the right branch to merge into as it currently only is an issue on beta builds.


changelog/bugfixes/2024-05-31-disable-dhcp-rapid-commit-for-cloudsigma.md
```diff
@@ -0,0 +1 @@
+- Disabled the new DHCP RapidCommit feature which is enabled by default since systemd 255. CloudSigma provides an incompatible implementation which results in cloud-init not being applied as no IP is issued. ([Flatcar#1438](https://github.com/flatcar/Flatcar/issues/1438))
```
ader1990, line 1: Hello, the link in the changelog should be to this PR and the commit message can contain something like:

Fixes: <issue link>
or
See: <issue link>

ader1990 (commented):
Regarding the implementation, as @pothos mentioned, the OEM sysext load might be done too late in the process, thus having systemd already running with RapidCommit=True already.

Can you please try to build the image with locally and make sure the fix works with the built image?

I will try locally with maybe another OEM like qemu that is easier to use too.

Thanks.

dosomder:
Nice work, just what I needed.. :+1:

changelog/bugfixes/2024-05-31-disable-dhcp-rapid-commit-for-cloudsigma.md
daMupfel: Thanks, i adapted it. Hopefully the way you meant :).

daMupfel:
Thanks for checking in to this, i will try to build it locally. Have never done it so we will see how it goes :).

daMupfel:
So i was finally able to test this.

While it does work, it only works after a reboot since the network configuration is already loaded at that point.

I was not able to figure out how to restart the network service. I tried:
```
runcmd:
  - sudo systemctl restart systemd-networkd
```

But found out later that runcmd is not supported by the coreos-cloudinit implementation.

Do you have a suggestion on how i could do that?

Regards,

David

daMupfel:
By the way, what is the plan of moving the next beta to stable? Since it would break for us at that point :)

daMupfel:
Okay, so after a bit of looking around in the [coreos-cloud](https://github.com/flatcar/coreos-cloudinit) i found out that one can also provide network units to cores.units.

I built a custom image based on the latest beta and applied this change. This does now seem to work as intended.

Not sure if i require `runtime: True` i don't think it matters as it will write the file on every boot anyway?

Regards,

David

daMupfel:
@ader1990 Hi :),

Thank You for the initial review! Is there anything else i can provide to help this PR?

daMupfel:
This seems to have missed the stable now. I would like to ask again what there is we can provide to integrate this. We currently require an own image which we would like to stop again.

daMupfel:
Hi @tormath1, i split it now up into two commits :).

daMupfel:
@tormath1 Thank you very much for taking care of my PR!

We don't have any affiliation other than being customers of CloudSigma. At some point, CloudSigma made an effort to support CoreOS, which is where this comes from. At that time, one could select CoreOS as a base image for new VMs. CoreOS then disappeared at some point from the UI.
What we do nowadays is convert the OEM image to a raw image and then upload it as a custom base image to CloudSigma.

Please take this with a grain of salt since I am only an external observer.

PR #2027, opened 2024-06-12
netperf: use the MIT licensed version

ader1990:
Fixes: https://github.com/flatcar/Flatcar/issues/1459

Needs first a PR that syncs the latest Gentoo repo files for netperf.


ader1990:
@t-lo @miao0miao would it be advisable to add the ebuild in this commit and not use github actions sync for the future versions?

t-lo:
@ader1990 We should add the new ebuild here and also add net-analyzer/netperf to https://github.com/flatcar/scripts/blob/main/.github/workflows/portage-stable-packages-list here in this PR, too. Only acct-user/netperf and acct-group/netperf are there currently, which seems like a miss.

ader1990:
> @ader1990 We should add the new ebuild here and also add net-analyzer/netperf to https://github.com/flatcar/scripts/blob/main/.github/workflows/portage-stable-packages-list here in this PR, too. Only acct-user/netperf and acct-group/netperf are there currently, which seems like a miss.

Added a commit to sync with Gentoo upstream and then enable the version with MIT license. Also added the package to the github workflow that does the sync PR.

chewi (approved):

PR #2028, opened 2024-06-12
Sync cargo.eclass and drop coreos-cargo.eclass

chewi:
# Sync cargo.eclass and drop coreos-cargo.eclass

coreos-cargo.eclass has some hacks to cross-compile with Cargo. I have now added cross-compiling to the upstream Gentoo cargo.eclass using a lot less code, removing the need for awkward tweaks to enable coreos-cargo.eclass. This syncs the upstream changes and drops our eclass entirely.

The `TARGET_*` wrappers that coreos-cargo.eclass set up no longer seem necessary. These worked around an issue with Cargo that was reported many years ago, but I was not able to reproduce it now.

## How to use

```
CFLAGS="-O2 -pipe -march=armv8-a" CXXFLAGS="-O2 -pipe -march=armv8-a" SYSROOT=/build/arm64-usr CROSS_CMD=ebuild emerge-aarch64-cros-linux-gnu ~/trunk/src/scripts/sdk_container/src/third_party/coreos-overlay/coreos-base/afterburn/afterburn-5.5.1-r1.ebuild clean install
```

## Testing done

I did the above, and also with ue-rs. fero-client doesn't seem to build at the moment for an unrelated reason (Rust too new?), but it was already broken by the coreos-cargo.eclass hacks anyway.

- [ ] Changelog entries added in the respective `changelog/` directory - **Not user-facing**.


chewi:
I'll look at fero-client since that's breaking CI. I see there's still an issue with thin-provisioning-tools, as it has a custom `src_install`. I'll address that in Gentoo.

sdk_container/src/third_party/portage-stable/eclass/rust-toolchain.eclass
krnowak, line 1: Could you please add `eclass/rust-toolchain.eclass` to `.github/workflows/portage-stable-packages-list`?

krnowak (commented):
Looks good, thanks! About `fero-client` - could you maybe ask on Flatcar channel if we are even using it in SDK? It looks like it would be a part of the release process. Maybe instead of fixing it, we could just drop it. Because otherwise we probably would need to fork it into the `flatcar` org as it's archived by upstream.

sdk_container/src/third_party/portage-stable/eclass/rust-toolchain.eclass
chewi: Dropped and done.

chewi:
I've fixed up thin-provisioning-tools, as well as aadvark-dns and netavark, which were creating amd64 binaries in the arm64 podman sysext. The fixes aren't actually upstream yet, as the eclass change needs a few days for feedback, and I need to run the other changes by their respective maintainers. They may insist on a revbump, meaning the changes won't be in the stable versions, but hopefully not.

krnowak (approved):
Cool. We will need to remember not to clobber your modifications during weekly updates if they are not upstreamed by the time the weekly updates happen.

chewi:
Thanks for approving. No need to rush the merge though, I'll do it early next week, which should be enough time for the upstream merges.

PR #2048, opened 2024-06-20
ci-automation/release: add support for Managed idenities

sayanchowdhury:
Add the support to run the pre-release and the release jobs via managed idenities

to be merged with https://github.com/flatcar/mantle/pull/535

### DIFF
--- a/ci-automation/release.sh
+++ b/ci-automation/release.sh
@@ -108,8 +108,7 @@ function _inside_mantle() {
           --debug \
           --platform="${platform}" \
           --aws-credentials="${aws_credentials_config_file}" \
-          --azure-profile="${azure_profile_config_file}" \
-          --azure-auth="${azure_auth_config_file}" \
+          --azure-identity \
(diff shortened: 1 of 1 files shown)


ci-automation/release.sh
```diff
@@ -110,6 +110,7 @@ function _inside_mantle() {
           --aws-credentials="${aws_credentials_config_file}" \
           --azure-profile="${azure_profile_config_file}" \
           --azure-auth="${azure_auth_config_file}" \
```
jepio, line 112: can these be dropped now?

ci-automation/release.sh
```diff
@@ -140,6 +141,7 @@ function _inside_mantle() {
         --product-ids="${pid}" \
         --azure-profile="${azure_profile_config_file}" \
         --azure-auth="${azure_auth_config_file}" \
```
jepio, line 143: can these be dropped now?

PR #2068, opened 2024-06-28
Finish merged-usr process for generic images, move SDK over to it too

krnowak:
Current situation:
- SDK:
  - `/bin`, `/sbin`, `/lib`, `/lib64`, `/usr/bin`, `/usr/sbin`, `/usr/lib`, `/usr/lib64` are all directories
- generic:
  - `/usr/bin`, `/usr/sbin`, `/usr/lib`, `/usr/lib64` are all directories
  - `/bin`, `/sbin`, `/lib`, `/lib64` are symlinks to their counterparts in `/usr`.

New situation for both (essentially known as `merged-usr` in Gentoo):
- `/usr/bin`, `/usr/lib`, `/usr/lib64` are all directories
- `/sbin` and `/usr/sbin` are effectively symlinks to `/usr/bin`
- `/bin`, `/lib`, `/lib64` are symlinks to their counterparts in `/usr`.

Directories in `/usr/local` are left as is, so `/usr/local/sbin` is still a directory, not a symlink to `bin`.

There is quite a sizeable refactor of the baselayout ebuild, so please have a look at the commit messages.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1714/cldsv/

Needs https://github.com/flatcar/baselayout/pull/34 and https://github.com/flatcar/mantle/pull/551.


krnowak:
Filed a fix for Gentoo: https://github.com/gentoo/gentoo/pull/37397

sdk_container/src/third_party/coreos-overlay/sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -1,138 +1,69 @@
...
 
 LICENSE="GPL-2"
 SLOT="0"
-IUSE="cros_host symlink-usr"
+IUSE="cros_host"
 
 # This version of baselayout replaces coreos-base
-DEPEND="sys-apps/systemd
+DEPEND="
+	sys-apps/systemd
```
chewi, line 28: This systemd dependency can cause issues, so I'd appreciate it if you could change it to `RDEPEND` only.

krnowak: What about dropping it? The `net-dns/libidn2` dependency also looks weird here. The rest of deps are conflicts to some old versions, which could be dropped too, so maybe I'll delete the DEPEND variable entirely?

chewi: Technically, systemd is needed so that systemd-tmpfiles can process these files at runtime. You could argue that we only ever use this package with systemd-based profiles and therefore systemd is part of the system set.

I don't know what libidn2 would have been used for.

jepio: > This systemd dependency can cause issues, so I'd appreciate it if you could change it to RDEPEND only.

What issues?

chewi: It should have been `BDEPEND` rather than `DEPEND`. IIRC, it caused cyclic dependency issues when I cross-compiled stage3 for the arm64 SDK.

chewi (approved):

PR #2079, opened 2024-07-01
Replace nmap netcat with openbsd netcat

krnowak:
This PR replaces net-analyzer/nmap with net-analyzer/openbsd-netcat. It also installs the `ncat` symlink for compatibility.

Fixes https://github.com/flatcar/Flatcar/issues/1460

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/4414/cldsv/

There are the following PRs in mantle and init that this PR is pulling in:
- mantle: https://github.com/flatcar/mantle/pull/539
- init: https://github.com/flatcar/init/pull/124

--


krnowak:
> Unfortunately I think this might break the `flatcar-update` script for now: https://github.com/flatcar/init/blob/ed0ce5ab08d77ff8fa775975961987ae2704fb34/bin/flatcar-update#L272

Well spotted. Thanks! There is also a kola test that uses ncat (`docker.network` I think), so it will also need an update.

krnowak:
CI passed.

chewi (approved):

PR #2080, opened 2024-07-02
Fixes for flatcar-workon

krnowak:
The `update_chroot` calls `flatcar_workon` to get a list of workon packages to rebuild on SDK. But not passing `--board` flag to `flatcar_workon` means using default board (`amd64-usr`), so it really gets a list of packages from the board. But the script failed for me during the sdk_container job, because no board was set up that point.

So this PR adds back the `--host` flag to `flatcar_workon` (which used to be here in `cros_workon`) and makes `update_chroot` pass the new flag to the script.

Changes tested as a part of [sdk-container job for merged-usr changes](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk_container/1148/consoleFull).

There are also [logs for the failing job](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk_container/1146/consoleText), search for `amd64-usr has not been setup yet` there.

### DIFF
--- a/flatcar_workon
+++ b/flatcar_workon
@@ -6,6 +6,8 @@
 
 DEFINE_string board "${DEFAULT_BOARD}" \
   "The board to set package keywords for."
+DEFINE_boolean host "${FLAGS_FALSE}" \
+  "Uses the host instead of board"
 
 FLAGS_HELP="usage: $0 <command> [flags] <atom>
 commands:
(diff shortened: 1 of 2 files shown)


flatcar_workon
```diff
@@ -17,6 +19,13 @@ eval set -- "${FLAGS_ARGV}"
 
 set -euo pipefail
 
+# If both board and host are specified, just use host, because board
+# does not have to be specified and may come from default, in which
+# case there's no way to override.
+if [[ -n ${FLAGS_board} ]] && [[ ${FLAGS_host} = "${FLAGS_TRUE}" ]]; then
```
chewi, line 25:
```suggestion
if [[ -n ${FLAGS_board} && ${FLAGS_host} = "${FLAGS_TRUE}" ]]; then
```

chewi (approved):
Ah! I intended for no `--board` argument to imply `--host` but I got that wrong. I did notice this late on Friday while doing my own build but hadn't looked into it yet. Thanks for fixing.

PR #2089, opened 2024-07-03
sys-apps/systemd: updated to minor version 255.8

ader1990:
Synced with Gentoo and bumped sys-apps/systemd version from 255.4 to 255.8

## Testing done


sdk_container/src/third_party/coreos-overlay/sys-apps/systemd/systemd-255.8.ebuild
```diff
@@ -258,7 +255,7 @@ src_unpack() {
 
 src_prepare() {
 	local PATCHES=(
-		"${FILESDIR}"/255-install-format-overflow.patch
+		"${FILESDIR}/systemd-test-process-util.patch"
```
ader1990, line 258: This patch `systemd-test-process-util.patch` is not per se required for Flatcar, maybe we can remove it too?

sayanchowdhury: Since this patch is pulled from upstream, I would suggest to keep this patch around to keep the diff as minimal as possible.

ader1990: I thought so too, just wanted to get another opinion.

ader1990:
> Thanks, this can be merged once the image changes has been checked (to see if any important files or lib have been removed / added)

both change files for amd64/arm64 look as expected. The udev db/hwdb changed the most by +234 KB on amd64.

PR #2092, opened 2024-07-05
ci-automation/vendor-testing/azure: Allow passing kola vnet

jepio:
# ci/azure: Allow passing --azure-kola-vnet

This adds support for providing a value for the newly introduced --azure-kola-vnet kola parameter through the environment. This parameter is meant to indicate that kola is running inside of a vnet in Azure and the kola created storage account will be restricted to being accessed from that vnet. This lets us disable public access to storage accounts.

Needs a corresponding change to jenkins jobs, because we have no way of determining what vnet a worker node is connected to programmatically. So it needs to be defined by the job.

## How to use

Run kola in Azure, in the same subscription that kola has access to and pass `--azure-kola-vnet`.

## Testing done

### DIFF
--- a/ci-automation/vendor-testing/azure.sh
+++ b/ci-automation/vendor-testing/azure.sh
@@ -56,6 +56,7 @@ run_kola_tests() {
       --azure-size="${instance_type}" \
       --azure-hyper-v-generation="${hyperv_gen}" \
       ${AZURE_USE_GALLERY} \
+      ${AZURE_KOLA_VNET:+--azure-kola-vnet=${AZURE_KOLA_VNET}} \
       ${azure_vnet_subnet_name:+--azure-vnet-subnet-name=${azure_vnet_subnet_name}} \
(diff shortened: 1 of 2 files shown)


ci-automation/vendor-testing/azure.sh
```diff
@@ -56,6 +56,7 @@ run_kola_tests() {
       --azure-size="${instance_type}" \
       --azure-hyper-v-generation="${hyperv_gen}" \
       ${AZURE_USE_GALLERY} \
+      ${AZURE_KOLA_VNET:+--azure-kola-vnet=${AZURE_KOLA_VNET}} \
```
krnowak, line 59: Please also add the (empty) definition of AZURE_KOLA_VNET to `ci-automation/ci-config.env`, so it'll be obvious that it's something to be provided in env.

krnowak (approved):
One thing left, but otherwise looks good.

krnowak (approved):

PR #2115, opened 2024-07-11
Upgrade to Catalyst 4 and don't use repo snapshots for stage1

chewi:
# Upgrade to Catalyst 4 and don't use repo snapshots for stage1

Split out from #2093.

Although Catalyst 4 still only has release candidates, Gentoo has been using it for its releases for years. Catalyst 3 was masked a while ago, so we should not linger on this version.

Catalyst 4 has totally changed the way repositories are handled. It only works when the name of the directory containing the repository matches the configured name of that repository. This was not the case for us, with the coreos repository residing in the coreos-overlay directory. We wanted to move and rename our repositories anyway, but this is a big change, so we'll do separately. For now, this just renames coreos to coreos-overlay.

This also stops using repo snapshots for stage1, which avoids the need to fix up the coreos repo name in the seed with a hook script and also renders all the existing hooks scripts obsolete. This approach is in line with what Gentoo does.

## How to use

Simply enter a recent SDK and run `sudo ./bootstrap_sdk`. Note that this will automatically modify the SDK container to rename the coreos repo and upgrade Catalyst, so you will need to recreate the container to go back to an older branch.

## Testing done

I have manually rebuilt the SDK and am currently also rebuilding it in Jenkins using 4012.0.0 as the seed.


chewi:
I have made associated documentation changes in flatcar/flatcar-website#351.

chewi:
I ran a two-phase build in Jenkins, and it looks good. It doesn't show the changes in the SDK, but the production image is barely changed besides packages changing from ::coreos to ::coreos-overlay.

Good to go then? I don't think it needs a changelog entry as its not user-facing.

ader1990:
As it affects the end user image package full name, maybe it is worth a changes changelog? `packages changing from ::coreos to ::coreos-overlay.`

chewi:
> As it affects the end user image package full name, maybe it is worth a changes changelog? `packages changing from ::coreos to ::coreos-overlay.`

It's going to change to ::flatcar-overlay in the PR after this one, but I guess I should still add this in the meantime and update it later.

ader1990 (approved):
LGTM, thanks!

build_library/catalyst.sh
```diff
@@ -3,6 +3,11 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
+# Before doing anything, ensure we have at least Catalyst 4.
+if catalyst --version | grep -q "Catalyst [0-3]\."; then
+    sudo emerge -v1 ">=dev-util/catalyst-4" || exit 1
```
jepio, line 8: could use a `--jobs ${NUM_JOBS}`.

catalyst.sh is a library file and meant to be sourced, executing a build command from such a file is a bad idea. this would fit better in catalyst_init() or in the caller.

build_library/catalyst.sh
```diff
@@ -207,58 +198,50 @@ catalyst_init() {
     # so far so good, expand path to work with weird comparison code below
     FLAGS_seed_tarball=$(readlink -f "$FLAGS_seed_tarball")
 
-    if [[ ! "$FLAGS_seed_tarball" =~ .*\.tar\.bz2 ]]; then
-        die_notrace "Seed tarball doesn't end in .tar.bz2 :-/"
+    if [[ ! "$FLAGS_seed_tarball" =~ .\.tar\.(bz2|xz) ]]; then
+        die_notrace "Seed tarball doesn't end in .tar.bz2 or .tar.xz :-/"
```
jepio, line 202: why - is the compression configurable? on bincache i see we're still producing a bz2 file.

build_library/catalyst.sh
```diff
@@ -207,58 +198,50 @@ catalyst_init() {
...
-    info "    portage/repos.conf/coreos.conf"
-    repos_conf > "$TEMPDIR/portage/repos.conf/coreos.conf"
     info "    stage1.spec"
     catalyst_stage1 > "$TEMPDIR/stage1.spec"
-    info "    stage2.spec"
-    catalyst_stage2 > "$TEMPDIR/stage2.spec"
-    info "    stage3.spec"
-    catalyst_stage3 > "$TEMPDIR/stage3.spec"
-    info "    stage4.spec"
-    catalyst_stage4 > "$TEMPDIR/stage4.spec"
```
jepio, line 246: why did you move the spec generation from `write_configs()` to `catalyst_build()`?

build_library/portage/profile/package.use.force/releng/alternatives
jepio, line 1: i can't find any reference to this file/directory? are you sure it's being consumed.

sdk_container/src/third_party/portage-stable/app-arch/pixz/pixz-1.0.7-r1.ebuild
jepio, line 1: add to auto-update list

sdk_container/src/third_party/portage-stable/sys-fs/squashfs-tools-ng/squashfs-tools-ng-1.3.0.ebuild
jepio, line 1: add to auto-update list

build_library/catalyst.sh
chewi: Both good points. I was originally going to put it in `catalyst_init` but then thought I needed the upgrade to happen earlier. I was wrong though, so I'll do that.

build_library/catalyst.sh
chewi: Gentoo switched to xz a while back, and this is left over from when I was trying a vanilla Gentoo seed. That turned out to not be viable, but I also used xz when making my own seed. It doesn't hurt. bzip2 only wins at compression speed anyway, which is why Gentoo switched.

build_library/catalyst.sh
chewi: You used to be able to set `source_subpath` via a command line option, but you can't anymore. The `SEED` variable changes depending on which stages were built earlier, so now it's easiest to write the spec when it's actually needed.

build_library/portage/profile/package.use.force/releng/alternatives
chewi: It's used by this line in `write_configs()`:

```
cp -r "${BUILD_LIBRARY_DIR}"/portage/ "${TEMPDIR}/"
```

The file itself is applied to the profile to temporarily force certain USE flags while building.

sdk_container/src/third_party/portage-stable/app-arch/pixz/pixz-1.0.7-r1.ebuild
chewi: Done.

sdk_container/src/third_party/portage-stable/sys-fs/squashfs-tools-ng/squashfs-tools-ng-1.3.0.ebuild
chewi: Done.

build_library/portage/profile/package.use.force/releng/alternatives
jepio: But these are already present in the base profile: https://github.com/flatcar/scripts/blob/main/sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.use.force#L4-L15

So we don't need this file and can also remove the `cp -r "${BUILD_LIBRARY_DIR}"/portage/ "${TEMPDIR}/"` step - right?

chewi: Ah, I missed that. You normally only want to force these temporarily so that they can be changed by the user later, but that doesn't apply in our case. I'll drop that file, but I'll keep the config mechanism for the QEMU bits. It might also be useful for something else later.

jepio: Which "config mechanism for the QEMU bits"?

chewi: Ah yes, those don't exist yet. :smile: In my forthcoming arm64 SDK branch, some other config is added under build_library/portage to allow building under QEMU with Catalyst.

PR #2121, opened 2024-07-15
Move and rename package repos

chewi:
# Move and rename package repos to repos/{gentoo-subset,flatcar-overlay}

Catalyst 4 has totally changed the way repositories are handled. It only works when the name of the directory containing the repository matches the configured name of that repository. We already renamed coreos to coreos-overlay accordingly, but we actually want entirely different names and more convenient locations too.

The repositories are now known as gentoo-subset and flatcar-overlay, and they live under scripts/repos. Using the same name as upstream Gentoo would have been problematic, and just "flatcar" would have looked awkward in documentation.

## How to use

Simply enter a recent SDK and run `sudo ./bootstrap_sdk`. Note that this will automatically update the SDK container's Portage configuration, so you will need to recreate the container to go back to an older branch.

## Testing done

I have built an SDK manually. I also now trigger a Jenkins build.


chewi:
The two-phase Jenkins job against 4012.0.0 passed. I know there are conflicts, but the nature of this change means they will frequently appear, so I'll deal with them once this is reviewed.

chewi:
I just rebased this following the big package update. That really wasn't fun, so I would appreciate eyes on this soon.

jepio:
It’s a good idea to get everyone to consider what the possible side effects or breakage due to this change could be before we merge it.

I can think of one that existing tests unfortunately didn’t catch: the emerge-gitclone helper used in the devcontainer does not handle the new repo paths - but the tests don’t exercise that sufficiently well (only binpkgs are tested). Could you check that? A new test case would be good for this.
________________________________

ader1990:
@chewi would it be possible to have two commits? one commit with the file renames and one commit with the actual logic needed to be reviewed? my browser bails on me after I try to scroll a little. And this way would be cleaner if we d need to point to the logic change.

jepio:
> It’s a good idea to get everyone to consider what the possible side effects or breakage due to this change could be before we merge it.

Thought of another one - the workflows that are being touched in this PR are used to file PRs against branches for all channels. The updated repo paths in the workflows will only work for main, unless this gets cherry-picked to all branches, which is risky.

chewi:
Good idea with splitting the commit. That might make rebasing easier too. Pushed that now.

I was aware of emerge-gitclone, so I'll take a look at that. I was also looking at the flatcar-build-scripts. I hadn't realised the workflow scripts were shared though.

I had been trying to avoid creating symlinks, as I prefer to work through the breakage rather than stay in a semi-migrated state for ages. For these cases though, symlinks may be the simplest answer. We can remove them once we this has fully rotated through all the release channels. What about LTS releases though, are they also affected?

chewi:
See flatcar/flatcar-build-scripts#160 for fixes to the build scripts. These should ideally be merged first. They are backwards-compatible, so they don't need to wait for this.

build_library/build_image_util.sh
```diff
@@ -365,7 +365,7 @@ get_metadata() {
             local mirror="$(echo "${v}" | grep mirror:// | cut -d '/' -f 3)"
             if [ -n "${mirror}" ]; then
                 # Take only first mirror, those not working should be removed
-                local location="$(grep "^${mirror}"$'\t' /mnt/host/source/src/third_party/portage-stable/profiles/thirdpartymirrors | cut -d $'\t' -f 2- | cut -d ' ' -f 1 | tr -d $'\t')"
+                local location="$(grep "^${mirror}"$'\t' /var/gentoo/repos/gentoo-subset/profiles/thirdpartymirrors | cut -d $'\t' -f 2- | cut -d ' ' -f 1 | tr -d $'\t')"
```
jepio, line 368: why `/var/gentoo/repos`? there is nothing at that location in my sdk container

build_library/catalyst.sh
```diff
@@ -231,7 +231,7 @@ write_configs() {
     info "Configuring Portage..."
     cp -r "${BUILD_LIBRARY_DIR}"/portage/ "${TEMPDIR}/"
 
-    ln -sfT '/mnt/host/source/src/third_party/coreos-overlay/coreos/user-patches' \
+    ln -sfT '/var/gentoo/repos/flatcar-overlay/coreos/user-patches' \
```
jepio, line 234: same question about `/var/gentoo/repos` as above - can we stick to the same paths as used elsewhere?

sdk_container/src/third_party/coreos-overlay
jepio, line 1: these symlinks are dangling when referenced from `/mnt/host/source/src/third_party/` inside the SDK

sdk_lib/sdk_entry.sh
```diff
@@ -10,9 +10,29 @@ fi
 
 chown -R sdk:sdk /home/sdk
 
-# Fix up SDK repo configuration to use the new coreos-overlay name.
-sed -i -r 's/^\[coreos\]/[coreos-overlay]/' /etc/portage/repos.conf/coreos.conf 2>/dev/null
-sed -i -r '/^masters =/s/\bcoreos(\s|$)/coreos-overlay\1/g' /usr/local/portage/crossdev/metadata/layout.conf 2>/dev/null
+# Fix up SDK repo configuration to use the new layout.
```
jepio, line 13: This entry block has now gotten scary.

We used to rely on `update_chroot --setuponly` for setting up the sdk, how about calling it from here. It has the same definitions.

jepio (commented):
first pass

update_chroot
```diff
@@ -103,17 +103,17 @@ EMERGE_DEFAULT_OPTS="--verbose --oneshot"
 source "/etc/portage/make.conf.user"
 EOF
 
-sudo_clobber "/etc/portage/repos.conf/portage-stable.conf" <<EOF
+sudo_clobber "/etc/portage/repos.conf/gentoo-subset.conf" <<EOF
```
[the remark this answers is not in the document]
chewi: Catalyst 4 sets them up this way, so it's easiest to follow suit.

setup_board
```diff
@@ -233,9 +233,10 @@ fi
 
 info "Configuring portage in ${BOARD_ROOT}"
 sudo mkdir -p "${BOARD_ETC}/portage/"{profile,repos.conf}
-sudo ln -sfT "$(portageq get_repo_path / coreos-overlay)/coreos/user-patches" \
+sudo ln -sfT "$(portageq get_repo_path / flatcar-overlay)/coreos/user-patches" \
    "${BOARD_ETC}/portage/patches"
 sudo cp /etc/portage/repos.conf/* "${BOARD_ETC}"/portage/repos.conf/
+sudo rm -f "${BOARD_ETC}"/portage/repos.conf/coreos.conf
```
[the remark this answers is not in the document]
chewi: The SDK's own copy does, the board's copy does not.

sdk_lib/sdk_entry.sh
```diff
@@ -10,9 +10,29 @@ fi
...
+main-repo = gentoo-subset
+
+[gentoo-subset]
+location = /mnt/host/source/src/scripts/repos/gentoo-subset
+EOF
+cat > "/etc/portage/repos.conf/flatcar-overlay.conf" <<EOF
+[flatcar-overlay]
+location = /mnt/host/source/src/scripts/repos/flatcar-overlay
+EOF
+rm -f /etc/portage/repos.conf/coreos.conf
```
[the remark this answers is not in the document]
chewi: Good point. I was doing this against an SDK that hadn't been upgraded to Catalyst 4 yet.

build_library/build_image_util.sh
chewi: `/var/gentoo` is what it was before my Catalyst 4 work. I must have messed this up when splitting into separate commits.

update_metadata
```diff
@@ -16,7 +16,7 @@ eval set -- "${FLAGS_ARGV}"
 switch_to_strict_mode
 
 if [[ $# -eq 0 ]]; then
-    eval set -- portage-stable coreos-overlay
+    eval set -- gentoo-subset coreos
```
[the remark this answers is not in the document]
chewi: Correct. I don't think this script is currently used, but I'll fix it anyway.

update_distfiles
```diff
@@ -26,7 +26,7 @@ eval set -- "${FLAGS_ARGV}"
 switch_to_strict_mode
 
 if [[ $# -eq 0 ]]; then
-    eval set -- portage-stable coreos-overlay
+    eval set -- gentoo-subset coreos
```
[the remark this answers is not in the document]
chewi: As above.

sdk_container/src/third_party/coreos-overlay
chewi: Making them absolute would break outside the SDK. Would a `/mnt/host/repos` symlink work for you? I'm open to better suggestions.

jepio: Not sure what you're suggesting. When would someone access this outside the SDK?

setup_board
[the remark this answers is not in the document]
jepio: Yes, but the boards copies should be getting created in line 238.

Sounds like you may want something more along these lines:
```
sudo find "${BOARD_ETC}"/portage/repos.conf/ -mindepth 1 -delete
sudo cp /etc/portage/repos.conf/* "${BOARD_ETC}"/portage/repos.conf/
```

update_distfiles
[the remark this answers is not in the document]
jepio: > I don't think this script is currently used, but I'll fix it anyway.

Can you check that: https://github.com/search?q=update_distfiles+org%3Aflatcar&type=code

(i'm referring to the usage within `flatcar-linux-infra`).

sdk_container/src/third_party/coreos-overlay
chewi: I meant that `sudo ln -s source/src/scripts/repos /mnt/host/repos` makes those symlinks resolve. I sometimes work with the scripts repo outside of the SDK, so I think they would ideally not be broken in that situation.

jepio: Seems like we're going down the wrong path here - what/who are these symlinks for?

jepio: The background of my question is that `/mnt/host` is a ChromeOS legacy thing, so I wouldn't add new paths there.

The `sdk_container/` dir was meant for paths that are accessed from within the SDK container, for compatibility with the layout we had before (before the SDK container and git-submodules), where we had an SDK chroot with separate repos cloned  into `/mnt/host/source/src/...`.

update_chroot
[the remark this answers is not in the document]
chewi: We're currently not deleting all the existing *.conf files though. We could do, but I think that would be impolite, especially given that we've been talking about custom overlays recently.

setup_board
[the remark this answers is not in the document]
chewi: If we always just want to mirror what's in /etc/portate/repos.conf, I may as well just use rsync. It's part of Gentoo's `@system`, so we can assume it is available.

sdk_container/src/third_party/coreos-overlay
chewi: I added the symlinks after you mentioned that the GitHub workflows would break. I can see that .github/workflows/common.sh is used against other release channels. If we only need them to work for that, then I guess they can be absolute, even if they appear broken outside of the SDK.

chewi: Scratch that, looking at common.sh again, I think it needs to work both inside and outside the SDK. The relative symlink should work following this change I was already making though.

```diff
--- a/.github/workflows/common.sh
+++ b/.github/workflows/common.sh
@@ -13,9 +13,10 @@ if [[ ! -d "${WORK_SCRIPTS_DIR:-}" ]]; then
   fail "WORK_SCRIPTS_DIR env var does not point to a directory. It should point to the scripts repo which will be updated."
 fi
 
+# Update these to use the new repo location once LTS has migrated.
 readonly SDK_OUTER_TOPDIR="${WORK_SCRIPTS_DIR}"
 readonly SDK_OUTER_OVERLAY="${SDK_OUTER_TOPDIR}/sdk_container/src/third_party/coreos-overlay"
-readonly SDK_INNER_SRCDIR="/mnt/host/source/src"
+readonly SDK_INNER_SRCDIR="/mnt/host/source/src/scripts/sdk_container/src"
 readonly SDK_INNER_OVERLAY="${SDK_INNER_SRCDIR}/third_party/coreos-overlay"
 
 readonly BUILDBOT_USERNAME="Flatcar Buildbot"
```

build_library/catalyst.sh
chewi: Fixed.

sdk_lib/sdk_entry.sh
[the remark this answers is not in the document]
chewi: Fixed, assuming we take a different approach as per the other thread.

sdk_lib/sdk_entry.sh
chewi: ...except calling update_chroot breaks the version check that is then used to decide whether to update the board or not. :confused: Any ideas?

chewi: I guess I could move that update_chroot call inside that block because we'd also only want to run that when:

```
if [ "${version}" != "${DISTRIB_RELEASE}" ] ; then
```

update_distfiles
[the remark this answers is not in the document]
chewi: The script seems to upload to CoreOS servers, but perhaps I'm misunderstanding it.

chewi:
flatcar/flatcar-dev-util#19 will need to be merged first so that I can get a new version of emerge-gitclone into this PR.

update_distfiles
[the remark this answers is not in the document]
jepio: That script runs on the host behind https://mirror.release.flatcar-linux.net/ and syncs our distfile mirror. This is part of our infra, I just realized that it would also stop working with these changes :/

update_chroot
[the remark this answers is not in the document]
jepio: With the churn happening in this area, maybe deleting these conf files is the sane approach. /etc is part of the container image (same with /board/<xxx>/etc) so user modifications get lost by design.

Custom overlays will require a proper design and integration with scripts - perhaps the user will add an overlay in `scripts/repos/....` and then this will be taken into account when generating the files, but either way - the configs are managed by scripts.

build_library/catalyst.sh
```diff
@@ -62,8 +62,8 @@ storedir="$CATALYST_ROOT"
 distdir="$DISTDIR"
 envscript="$TEMPDIR/catalystrc"
 port_logdir="$CATALYST_ROOT/log"
-repo_basedir="/mnt/host/source/src/third_party"
```
ader1990, line 65: was this value wrong before or it is just a mapping? it should have been /mnt/host/source/src/scripts/third_party?

ader1990 (commented):
I see at least 12 references for /mnt/host/source/src/scripts/repos -> maybe this can be added as a configuration path once in one file in a future patch.

build_library/catalyst.sh
chewi: It wasn't wrong before, the `third_party` directory was and still is mounted separately there. It just doesn't contain these repositories anymore.

Yeah, maybe a variable wouldn't be a bad idea, although there are several references to the location in other git repos too.

chewi:
Adding flatcar/flatcar-linux-infra#280 to the pile.

update_distfiles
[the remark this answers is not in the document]
chewi: Okay. I've opened a PR for that.

chewi:
I'm putting this (and the associated PRs) on the ice for the time being.

PR #2143, opened 2024-07-22
Drop Catalyst package.use.force for alternatives

chewi:
# Drop Catalyst package.use.force for alternatives

These flags normally need to be temporarily forced during stage1, but we already force them permanently in our profiles.

Removing this appears to make build_library/portage redundant, but it will later be used to allow building under QEMU with Catalyst, and it could have other uses too.

## How to use

Build the SDK as usual.

## Testing done

I'll run CI now.


jepio:
Sorry for messing with the PR - i thought the update-sdk workflow might be able to help here, but its doesnt do a bootstrap necessary to test this.

chewi:
Ah, I thought the same, but I see what you mean. I can run it through Jenkins then.

ader1990 (approved):

PR #2153, opened 2024-07-24
Skip Catalyst stage2 as recommended by upstream Gentoo

chewi:
# Skip Catalyst stage2 as recommended by upstream Gentoo

From https://wiki.gentoo.org/wiki/Catalyst/Stage_Creation#Build_Stage3:

> It is not necessary to build stage2 in order to build stage3. Gentoo
> release engineering does not build stage2, and you should not need to
> unless you're intentionally building a stage2 as your goal.

We can now sync portage-stable/scripts with upstream because bootstrap.sh is only used during stage2, and the changes we had are no longer relevant. It seems likely the changes were already redundant anyway.

Apart from making things simpler, judging by a glance at Jenkins, it knocks about half an hour off the build time.

## How to use

Build the SDK with `./bootstrap_sdk` as usual.

## Testing done

A two-phase Jenkins build was successful. No notable changes were reported.

### DIFF
--- a/.github/workflows/portage-stable-packages-list
+++ b/.github/workflows/portage-stable-packages-list
@@ -474,10 +474,7 @@ net-vpn/wireguard-tools
 perl-core/File-Temp
 
 profiles
-
-# The bootstrap script has some modifications, so we can't sync scripts directory yet.
-#
-# scripts
+scripts
 
 sec-keys/openpgp-keys-gentoo-release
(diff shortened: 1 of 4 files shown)


krnowak (approved):
The docs will need a rewrite/clarifications, because they are confusing me in some places. But can be done as a follow-up PR.

PR #2158, opened 2024-07-26
Enable Landlock

l0kod:
# Enable Landlock

Landlock is a feature to create security sandboxes thanks to 3 dedicated system calls.  They are designed to be safe to used by any processes, which can only drop their privileges, similarly to seccomp.

The new Landlock LSM is build in the kernel (CONFIG_SECURITY_LANDLOCK=y) but it is not enough to make it usable by default.  As a stackable LSM, it is required to enable it at boot time with the CONFIG_LSM list.  See https://docs.kernel.org/userspace-api/landlock.html#kernel-support

As for other stackable LSMs, prepending Landlock to the default LSM list enables users to potentially get more protection by default by letting programs sandbox themselves.

Landlock is already enabled by default on Ubuntu 22.04 LTS, Fedora 35, Arch Linux, Alpine Linux, Gentoo, Debian, chromeOS, Azure Linux, WSL2, and more.

## How to use

The [full test suite](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/landlock?h=v6.6) is available in the kernel source code.

Alternatively, the [sandboxer](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/samples/landlock?h=v6.6) can be used to easily test Landlock.

Because the goal of sandboxing is to be seamlessly integrated in applications/services and work in a best-effort mode, no specific user space configuration should be needed to leverage Landlock.

## Backport

After some time, it would be good to backport this configuration down to the 5.15 kernel to protect users as much as possible.

## Testing done

I ran the `sort_config.sh` script and tested the configuration with a vanilla kernel (not a Flatcar system). It should be tested by someone with knowledge of the Flatcar SDK though.


jepio (approved):

jepio:
Test failures look expected right now.
these fail for PRs from forks
```
devcontainer.docker
devcontainer.systemd-nspawn
```
this is a bug in the test runner right now:
```
cl.locksmith.cluster
```

changelog/changes/2024-07-26-landlock.md
```diff
@@ -0,0 +1 @@
+- Landlock is now enabled for programs to sandbox themselves.
```
ader1990, line 1: can you please add a PR ref here similar to https://github.com/flatcar/scripts/commit/7d4917d67cb76e50191eeb6395b083f224b02b78#diff-b1db1840bbe6435ffd261eeff937e5e0ade3e082a99ed9dff6affc044b92a22dR1 ? Also, I would suggest adding some context `- Linux kernel security module 'landlock' is now enabled for programs to be able to sandbox themselves  ([flatcar/scripts#2158](https://github.com/flatcar/scripts/pull/2158))`

ader1990:
I verified the image changes and it seems that the make config + defconfig also enabled `CONFIG_SECURITY_PATH=y`, worth a note here for tracking the changes.

l0kod:
> I verified the image changes and it seems that the make config + defconfig also enabled `CONFIG_SECURITY_PATH=y`, worth a note here for tracking the changes.

That's indeed an automatic dependency. I'll update the commit message.

ader1990 (approved):

l0kod:
Thanks!

PR #2207, opened 2024-08-08
Drop sysroot-wrappers and *-config script wrappers

chewi:
# Drop sysroot-wrappers and *-config script wrappers

Supersedes #2177.

Gentoo will soon start patching gcc to respect the `ESYSROOT` environment variable so that cross-compiling into a different directory works out of the box. This removes the need for our own sysroot wrappers. This approach does not have the downside of sysroot paths leaking into the final build.

While working on this, I unintentionally dropped the generation of our *-config script wrappers but found that this only broke one package, so I fixed that up. I then flushed out the few remaining instances of this issue that were being masked by presence of these scripts on the build host. The Gentoo fixes for those can be taken in the round as it does not actually affect our build.

The included eclass changes have already been merged into Gentoo.

## How to use

Do a full Flatcar build, preferably in two phases.

## Testing done

I have previously done several two-phase builds in Jenkins. I will now perform a fresh one.


sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-devel/gcc/gcc-esysroot.patch
krnowak, line 1: Could you please add a README.md in the same directory that says when the patch could be dropped? Like "gcc-esysroot.patch" can be dropped when updating to gcc >=15.1.0." or something like that? My automation displays those files, so it's easy to know to drop it when the relevant update comes.

krnowak (approved):

PR #2216, opened 2024-08-08
app-emulation/open-vm-tools: Drop unnecessary kernel version patch

chewi:
# Drop unnecessary open-vm-tools patch

The project hasn't supported building Linux kernel modules for [years](https://github.com/vmware/open-vm-tools/commit/131e4a60fa2d2db0c30920d826d23aec44d8ab09), so we don't need to correct the detected kernel version.

## How to use

Check the package still builds on older kernels without producing any modules.

## Testing done

I've tried forcing kernel modules to build. It literally won't let you.

```
configure: error: Building kernel modules for Linux is no longer supported.
```

I'll run CI next.


krnowak (approved):
Cool, less stuff to maintain, thanks!

PR #2223, opened 2024-08-12
Upgrade Linux Kernel for main from 6.6.43 to 6.6.45

flatcar-infra:
Subject says it all.


ader1990:
I would suggest to hold the kernel upgrades 6.6.y if there are no CVEs or urgent issues until the next minor version 6.6.46, version that will contain the fix for the QEMU-KVM Flatcar guests: https://git.kernel.org/pub/scm/linux/kernel/git/stable/stable-queue.git/commit/?id=3e713b73c01fac163a5c8cb0953d1e300407a773 see queue-6.6/net-drop-bad-gso-csum_start-and-offset-in-virtio_net_hdr.patch .

Or we can add this patch to Linux Kernel 6.6.45 and then make sure we drop it on 6.6.46 -> commit already done by @tormath1 https://github.com/flatcar/scripts/pull/2194/commits/33259937abe19f612faac255706d5a509666fbc9.

PR #2230, opened 2024-08-14
Drop python tests from GCE and Azure OEM images and python sysext

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/4491/cldsv/

We used to drop them for GCE, but after converting to sysext, this change was dropped.

### DIFF
--- a/build_library/sysext_mangle_flatcar-python
+++ b/build_library/sysext_mangle_flatcar-python
@@ -7,4 +7,14 @@ pushd "${rootfs}"
 
 rm -rf ./usr/{lib/debug,share,include,lib64/pkgconfig}
 
+# Remove test stuff from python - it's quite large.
+for p in ./usr/lib/python*; do
+    if [[ ! -d ${p} ]]; then
+        continue
+    fi
+    # find directories named tests or test and remove them (-prune
+    # avoids searching below those directories)
+    find "${p}" \( -name tests -o -name test \) -type d -prune -exec rm -rf '{}' '+'
+done
+
 popd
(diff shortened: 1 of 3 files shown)


krnowak:
@chewi mentioned in https://github.com/flatcar/scripts/pull/2227#issuecomment-2289103017 that Gentoo installs the python tests even on installation media, because apparently some libraries expect those files to exist and there's no way to tell which files are those.

I suppose we could try removing them and maybe bringing them back if it breaks something?

chewi:
Looking at [Gentoo bug #531648](https://bugs.gentoo.org/531648), it looks like these are only used by other test suites, including some from third-party libraries. We probably don't want to support end users running any Python test suites on Flatcar, even third-party ones.

chewi (approved):
I suspect those files compress very well.

krnowak:
> Also noticed that reports are not handling the other sysexts we generate (python, zfs, podman and stuff).

Addressed by https://github.com/flatcar/flatcar-build-scripts/pull/162 and https://github.com/flatcar/scripts/pull/2236.

chewi:
Just an FYI, Gentoo has now split out these tests into a separate package so that they only get pulled in by the packages that actually need them.

krnowak:
> Just an FYI, Gentoo has now split out these tests into a separate package so that they only get pulled in by the packages that actually need them.

Thanks for letting me know. I can see that it was done for the dev-lang/python. But do other python packages still install their tests?

chewi:
I hadn't considered other packages. I can see even Portage does that. My bad, I could have improved that when I migrated it to Meson. :sweat_smile: At least the space taken up by these is relatively small compared to what Python itself installed.

PR #2229, opened 2024-08-13
Upgrade dev-lang/rust and virtual/rust in main from 1.80.0 to 1.80.1

flatcar-infra:
Subject says it all.

PR #2237, opened 2024-08-15
dev-lang/rust: Drop our custom package (1.80.0) in favour of upstream Gentoo's (1.80.1)

chewi:
# Switch to Gentoo Rust package and upgrade to 1.80.1

It is not clear why this was forked originally. One reason was to avoid the sys-apps/lsb-release dependency, but it probably wasn't just that. It seems likely that the upstream package did not support cross targets at the time. Now it does.

The package builds significantly faster now that LTO is no longer enabled, especially given that it was the expensive "fat" LTO rather than "thin" LTO. The installed size is still kept down by masking rustdoc.

Cross targets are now handled via the admittedly experimental `RUST_CROSS_TARGETS` support. This has been in place for a while, and I think it is fairly widely used now. If it does disappear, it would almost certainly be for something even better.

## How to use

Nothing much to see. Grab the new SDK and build some Rust.

## Testing done

This has been through many Jenkins runs. Everything looks good.


chewi:
This would supersede #2229.

build_library/toolchain_util.sh
```diff
@@ -471,22 +463,22 @@ install_cross_libs() {
...
+    echo "Installing dev-lang/rust with (potentially outdated) cross targets."
+    "${sudo[@]}" emerge "${emerge_flags[@]}" --binpkg-respect-use=y --update dev-lang/rust
+
+    [[
+       -d /usr/lib/rustlib/x86_64-unknown-linux-gnu &&
+       -d /usr/lib/rustlib/aarch64-unknown-linux-gnu
+    ]] && return
+
+    echo "Rebuilding dev-lang/rust with updated cross targets."
+    "${sudo[@]}" emerge "${emerge_flags[@]}" --usepkg=n dev-lang/rust
```
krnowak, line 481: There seem to be a possible scenario where we build rust twice?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.unmask
```diff
@@ -1,4 +1,4 @@
 # Overwrite portage-stable mask. We haven't seen LLVM related problems
 # with rust so far, so keep using 1.70.0.
-~dev-lang/rust-1.80.0
-~virtual/rust-1.80.0
+~dev-lang/rust-1.80.1
+~virtual/rust-1.80.1
```
krnowak, line 4: Heh, is this needed? I can see the comment saying about overwriting mask for rust 1.70 because of some LLVM problems. I think this could just go away, unless portage-stable profile is masking it.

krnowak (commented):
I think you'll need to also drop `rust-apply-patch.sh` and `rust-release-main.yaml` from `.github/workflows` - our package automation is going to take over updating rust.

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.unmask
krnowak: I don't see a problem with not being on the latest and freshest rust release. We have been so far only because the github action worked that way. :)

build_library/toolchain_util.sh
chewi: Yes, even before, it would install Rust from a binary package and then rebuild it if it didn't have the necessary cross targets. The previous logic was broken though, as it would do this once per cross target instead of just once. It's not like crossdev, the single package covers all the targets.

krnowak (approved):

PR #2238, opened 2024-08-15
sys-apps/baselayout: Don't fetch non-live version from git

chewi:
# sys-apps/baselayout: Don't fetch non-live version from git

Use a regular tarball from GitHub. baselayout is a special case because Catalyst initially installs it with `--nodeps`. We currently use stage 4 as a seed, which includes git, so this is fine right now, but we would like to switch to using stage 3 instead. You also might use a cross-compiled stage 3 as a seed when porting the SDK to a new architecture.

For the switch to stage 3, I had initially wanted to publish the tarball with a different filename, but this has proved super awkward due to the digest check. baselayout is the only thing preventing this change from being backwards-compatible. Rather than change the filename, I am getting this baselayout change in well ahead of the switch to avoid breaking current development branches. Only the SDK build would break though, which isn't something we typically do outside CI, and simply rebasing will fix it.

## How to use

Nothing to use here.

## Testing done

I've done a two-phase Jenkins run. There were no notable changes.


sdk_container/src/third_party/coreos-overlay/sys-apps/baselayout/baselayout-9999.ebuild
```diff
@@ -5,18 +5,20 @@ EAPI=7
 EGIT_REPO_URI="https://github.com/flatcar/baselayout.git"
 
 if [[ "${PV}" == 9999 ]]; then
+	inherit git-r3
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
 	EGIT_COMMIT="937a45faef0f7fa88d3d2c3f7ba60a7f3e2e82f7" # flatcar-master
+	SRC_URI="https://github.com/flatcar/baselayout/archive/${EGIT_COMMIT}.tar.gz -> flatcar-${PN}-${EGIT_COMMIT}.tar.gz"
```
krnowak, line 12: I remember that we used to have issues with the tarballs generated from a commit that they sometimes changed the hash, because one byte in the tarball has changed or something like that.

Not going to block the PR on the issue - I think we can just react to it when it happens.

krnowak (approved):

sdk_container/src/third_party/coreos-overlay/sys-apps/baselayout/baselayout-9999.ebuild
chewi: It has happened on extremely rare occasions, but I know GitHub try very hard to avoid it for precisely this reason. Nearly 9000 Gentoo ebuilds rely on this, so I think we can handle just one.

PR #2241, opened 2024-08-19
Upgrade Linux Kernel for flatcar-3975 from 6.6.43 to 6.6.47

flatcar-infra:
Subject says it all.


ader1990:
The failed test case seems to be unrelated to the change: `extra-test.[first_dual].cl.update.oem`, maybe someone with more in-depth knowledge of how Mantle OEM update tests work can take a look?

PR #2248, opened 2024-08-20
Drop eclass tests, move obsolete eclasses to overlay

krnowak:
We weren't syncing the eclass tests for ages and we weren't running them, so their existence in portage-stable is pointless at the moment. Also, three eclasses were removed from Gentoo, so we move them to our overlay as there is still a package or two that are using them.

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1682/cldsv/


krnowak:
CI passed (I didn't run any tests - no package has changed, so I was only checking if stuff still builds.)

chewi:
`epatch` is only used in one place and you simply need to replace the command with `eapply`.

chewi:
> `epatch` is only used in one place and you simply need to replace the command with `eapply`.

Okay, I see it is inherited by eutils. You could just remove that entry though.

chewi:
To be honest, I'd rather just drop the few remaining instances of EAPI 6. Gentoo has literally just purged almost all of them this week. It won't take me long at all.

chewi:
As discussed elsewhere, I have dropped all instances of EAPI 6, so we can just remove these eclasses now. I will open a new PR to succeed this one.

PR #2249, opened 2024-08-22
sys-kernel/bootengine: Bump to fix slow PXE/ISO boots with udev rule

chewi:
# sys-kernel/bootengine: Bump to fix slow boots with udev rule

See: flatcar/bootengine#98
See: https://github.com/flatcar/Flatcar/issues/1514

## How to use

See: flatcar/bootengine#98

## Testing done

Lots of manual testing and a packages_all_arches Jenkins job, as mentioned in the other PR. I'll do a GitHub run for this PR.


sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
```diff
@@ -7,7 +7,7 @@ EGIT_REPO_URI="https://github.com/flatcar/bootengine.git"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	EGIT_COMMIT="fb2631ce5e6a21d044c8dca73f59db01f9d5abcf" # flatcar-master
+	EGIT_COMMIT="5d4e7312cadaa428af4bc9b32de07e96720656be" # flatcar-master
```
jepio, line 10: needs to be d2e22122e8e106a21f64c048992290edc7053384

jepio (changes requested):
tip: I usually avoid using `Closes:` tags in commit messages, and prefer cross-referencing with `See:`. The issue has now been closed even though we haven't merged the scripts PR yet. Even after this PR is merged, we'd still need to backport to stable-3975 (because that's where the issue was spotted) before we can consider the issue closed. Sometimes it even makes sense to wait until a release has been done.

krnowak (approved):
If this needs backporting to other channels, make sure you don't backport additional changes in bootengine by accident. If other channels were using different commit for bootengine, we will need to create backport branches in baselayout that contain your fix only.

chewi:
Oh, sorry about that. Still getting my head around this. I thought alpha just tracked main.

PR #2258, opened 2024-08-27
net-firewall/conntrack-tools: remove obsolete 1.4.5

t-lo:
This change removes an outdated conntrack-tools ebuild - version 1.4.5 - from maintenance branch flatcar-4054. The branch uses 1.4.6-rc1 anyways so the legacy ebuild can safely be deleted.

This ebuild uses EAPI 6 and will cause a transient error message during build. It imports the linux-info eclass which was recently updated to only support EAPI 7 and higher. This will cause emerge to print an error message when considering the 1.4.5 ebuild. While that error message has no impact on the build because 1.4.6-rc1 is more recent and will hence be used, it can be misleading when investigating unrelated issues (which it did to me, hence this patch).

## How to use

- Check out this branch
  ```shell
  cd scripts
  git checkout t-lo/flatcar-4054-remove-conntrack-1.4.5
  ```
- Build it
  ```
  ./run_sdk_container -t
         ./build_packages
         ./build_image
  ```

## Testing done

Built the image (see above), built qemu uefi image, ran tests locally.


chewi:
Please update the Manifest too. I know we don't run QA tools that would check that but...

t-lo:
> Please update the Manifest too. I know we don't run QA tools that would check that but...

You're right! Good catch, will update. We should make use of these QA tools one day...

chewi (approved):

t-lo:
Tests are green except for `extra-test.[first_dual].cl.update.oem` which is currently broken; tracked in https://github.com/flatcar/Flatcar/issues/1518. Very likely unrelated. Merging.

PR #2259, opened 2024-08-28
build_sysext: Detect dev rebuild of release tag, fix build ID issue

t-lo:
build_sysext uses a base squashfs (basically a full snapshot of the Flatcar OS image) to build custom sysexts on top. Before building it ensures the base image actualy matches the OS version in the repository root.

The version string includes a BUILD_ID which might be auto-generated (by including common.sh) if it is not present in the version file - e.g. when the version is an official release (tag). This build ID auto-generation causes issues with the version check when image build and sysext build scripts run independently - each will generate its own build ID, and this will cause build_sysext's version check to fail.

build_sysext will now use the build id from the base squashfs when it is not set in the source tree's version.txt to work around that issue. This is a more general solution than 361eda220b368c3c3a959357c54db10d4c2f1d1a (which this patch reverts) as it directly addresses the issue in build_sysext instead of working around it in sysext_prod_builder.

## How to use

- Check out this branch
  ```shell
  cd scripts
  git checkout t-lo/fix-sysext-build-id-issue-in-image-to-vm
  ```
- Build it
  ```
  ./run_sdk_container -t
         ./build_packages
         ./build_image
         ./image_to_vm.sh --from ../build/images/amd64-usr/latest/ --format' 'qemu_uefi'
  ```

## Testing done

Ran the above builds.

## Cherry-pick / backport

To all active mainenance branches except LTS (which doesn't use sysext), i.e. `flatcar-3975` `flatcar-4012` `flatcar-4054`

### DIFF
--- a/build_sysext
+++ b/build_sysext
@@ -7,6 +7,7 @@
 # Script to generate sysext. See systemd-sysext(8). Prerequisite is
 # that you've run build_packages and build_image.
 
+
 SCRIPT_ROOT=$(dirname "$(readlink -f "$0")")
 . "${SCRIPT_ROOT}/common.sh" || exit 1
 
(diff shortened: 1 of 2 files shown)


build_sysext
```diff
@@ -183,7 +184,18 @@ mkdir "${BUILD_DIR}/install-root"
 mkdir "${BUILD_DIR}/workdir"
 mount -t overlay overlay -o lowerdir="${BUILD_DIR}/fs-root${pkginfo_lowerdirs}",upperdir="${BUILD_DIR}/install-root",workdir="${BUILD_DIR}/workdir" "${BUILD_DIR}/install-root"
 
-VERSION_BOARD=$(grep "^VERSION=" ${BUILD_DIR}/fs-root/usr/lib/os-release | cut -d = -f 2-)
+REPO_BUILD_ID="$(source "${REPO_MANIFESTS_DIR}/version.txt"; echo "$FLATCAR_BUILD_ID")"
+REPO_FLATCAR_VERSION="$(source "${REPO_MANIFESTS_DIR}/version.txt"; echo "$FLATCAR_VERSION")"
```
krnowak, line 188: The quoting here can be dropped when assigning - it's not necessary and gives a bit better experience in syntax coloring in code editors too. :)

```suggestion
REPO_BUILD_ID=$(source "${REPO_MANIFESTS_DIR}/version.txt"; echo "$FLATCAR_BUILD_ID")
REPO_FLATCAR_VERSION=$(source "${REPO_MANIFESTS_DIR}/version.txt"; echo "$FLATCAR_VERSION")
```

build_sysext
```diff
@@ -183,7 +184,18 @@ mkdir "${BUILD_DIR}/install-root"
...
+VERSION_BOARD=$(source "${BUILD_DIR}/fs-root/usr/lib/os-release" && echo "$VERSION")
+
+if [ -z "$REPO_BUILD_ID" ] ; then
```
krnowak, line 191: Using double brackets allows us to drop some more quoting.

```suggestion
if [[ -z $REPO_BUILD_ID ]] ; then
```

build_sysext
```diff
@@ -183,7 +184,18 @@ mkdir "${BUILD_DIR}/install-root"
...
+    BASE_SQUASHFS_BUILD_ID=$(source "${BUILD_DIR}/fs-root/usr/lib/os-release" && echo -n "$BUILD_ID")
+    info "This is a dev rebuild of an official release tag: No BUILD ID set in '${REPO_MANIFESTS_DIR}/version.txt'.  Will use base squasfs BUILD ID for version check."
+    info "Repo root FLATCAR_VERSION is '$REPO_FLATCAR_VERSION', squashfs build ID is '$BASE_SQUASHFS_BUILD_ID'"
+    FLATCAR_VERSION="${REPO_FLATCAR_VERSION}+${BASE_SQUASHFS_BUILD_ID}"
```
krnowak, line 195: Is `BASE_SQUASHFS_BUILD_ID` guaranteed to be not empty? If not, it's possible we will end up with flatcar version like `4054.0.0+`. Thus:
```suggestion
    FLATCAR_VERSION="${REPO_FLATCAR_VERSION}${BASE_SQUASHFS_BUILD_ID:++}${BASE_SQUASHFS_BUILD_ID}"
```

krnowak (approved):

t-lo:
Tests are green and PR is approved, but main has moved on. Will force rebase and merge since it's a small change and well understood.

t-lo:
Merged and cherry-picked for flatcar-3975, flatcar-4012, flatcar-4054.

PR #2261, opened 2024-08-29
sys-block/open-iscsi: Migrate coreos-stable to portage-stable, 2.1.10

chewi:
# open-iscsi: Migrate to portage-stable, 2.1.10 bump

The old version 2.1.4 does not cross-compile without help from QEMU that we cannot rely on going forwards. 2.1.10 is Meson-based and handles this much better.

Rather than update the package in-place, migrate it to portage-stable and cover the differences with a small patch and env script. Upstream now carries the systemd files, so we do not need to add these.

## How to use

1. Boot up Flatcar under QEMU.
2. `sudo systemctl start iscsid`
3. Check the name in `/etc/iscsi/initiatorname.iscsi`.
4. Install `targetcli` on the host.
5. Set up a test target:

```
backstores/fileio create test /tmp/test.img 100m
iscsi/ create iqn.2006-04.com.example:test-target
cd iscsi/iqn.2006-04.com.example:test-target/tpg1/
luns/ create /backstores/fileio/test
set attribute generate_node_acls=1
acls/ create <initiator name>
exit
```

6. Return to Flatcar.
7. `sudo iscsiadm -m discovery -t sendtargets -p <host LAN IP>`
8. `sudo iscsiadm -m node -l -T <target from above> -p <host LAN IP>`
9. Play around with /dev/sda.
10. `sudo systemctl enable iscsi`
11. Reboot and check /dev/sda is still there.

## Testing done

I did the above. I also ran a Jenkins job against qemu_uefi. It was suggested that I add a Kola test. The usual way to write a test like this would be to spin up two instances as the client and server. I'd like to do this, but `targetcli`, which manages the Linux server, is written in Python, which Flatcar doesn't have.


sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
krnowak, line 1: I suppose this is a good first step - it basically preserves our modifications. One thing I wanted to note is that we do not require moving stuff out from /etc anymore. What we could do here instead is:
- keep `/etc/iscsi/iscsid.conf`
- make `/usr/share/iscsi/iscsid.conf` a symlink to `/usr/share/flatcar/etc/iscsi/iscsid.conf` (for backward compatibility only)
(have a look at coreos-base/misc-files package which already does it for a couple of packages)

On the other hand, handling `/var` stuff and the dropin probably still requires that hook you added.

krnowak (commented):
An entry for `sys-block/open-iscsi` in `.github/workflows/portage-stable-packages-list` is missing.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
krnowak: I'm a bit torn now whether this `/usr/share/iscsi/iscsid.conf -> /usr/share/flatcar/etc/iscsi/iscsid.conf` creation should be a part of `coreos-base/misc-files` (which was the place for things like that) or in the hook you wrote. The hook solution didn't come to my mind when I created the misc files package.

chewi: Hah, I had been quietly thinking "Why isn't /etc done with overlayfs?" :laughing:

I think I'd rather keep the logic for this together with the package it relates to.

chewi: Is there a reason why at least some of /var isn't handled like this?

krnowak: Dunno, really. Possibly because there is not `/var` variant of systemd-sysext. :) (Which is what we would eventually want to use for `/etc`).

But anyway, it's Friday, I'm sleepy, I may not think clearly. What I remember is that we can turn `/var` into tmpfiles when finishing up the image, see https://github.com/flatcar/scripts/blob/872ea6d14db3c2e5f651f5e8c0f67b2cbebaf061/build_library/build_image_util.sh#L671-L689

I think there are two tmpfiles generated for `/var` contents. Maybe the iscsi stuff would end up there too, so this modification won't be needed either.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
```diff
@@ -0,0 +1,22 @@
+cros_post_src_install_mv_files() {
+	mv "${ED}"/var/lib/iscsi/ "${ED}"/usr/share/ || die
+	rmdir "${ED}"/var/lib || die
+
+	dosym ../flatcar/etc/iscsi/iscsid.conf /usr/share/iscsi/iscsid.conf
```
jepio, line 5: `/usr/share/flatcar/etc` is the lowerdir for the `/etc/ overlay, so `/usr/share/iscsi/iscsid.conf` should just point at `/etc/iscsi/iscsid.conf`? or is it needed at all?

chewi: The others like `/usr/share/ssh/sshd_config` all point to `../flatcar/etc`. I don't know why, as it should still work like you say, and you wouldn't need that empty file. I was just being consistent.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
chewi: Interesting, thanks. This will work for `/var/db/iscsi`, which has a `keepdir` call. It won't work for `/var/lib/iscsi` though because it doesn't. This directory sticks because it contains `ifaces/iface.example`. If we don't care about the example file, I could simply call `keepdir /var/lib/iscsi/ifaces` and let gen_tmpfiles.py handle the rest. We generally don't install documentation, so dropping the example file feels like the right thing to do.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
krnowak: It's needed for backward compatibility, if some user already has a symlink at `/etc/iscsi/iscsid.conf` that points to `/usr/share/iscsi/iscsid.conf`. So if `/usr/share/iscsi/iscsid.conf` points to `/etc/iscsi/iscsid.conf`, then we have symlinks pointing at each other. That's why we should point `/usr/share/iscsi/iscsid.conf` to `/usr/share/flatcar/etc/iscsi/iscsid.conf`.

krnowak: If someone has /etc from old old versions of Flatcar, then their `/etc/iscsi/iscsid.conf` will be a symlink, that overwrites whatever the lower layer in `/usr/share/flatcar/etc` has.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
krnowak: Sounds good, please check if it does the trick. You can check the contents of the `base_image_var.conf` and `base_image_var_late.conf` files in the generated image.

chewi:
Took a few goes to get it right, but it's looking good now. gen_tmpfiles.py takes care of the tmpfiles stuff now, so the env script is shorter. 2.1.10 got stabilised in the meantime.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
```diff
@@ -0,0 +1,18 @@
+cros_post_src_install_tmpfiles() {
+	keepdir /var/lib/iscsi/ifaces
+
+	dosym ../flatcar/etc/iscsi/iscsid.conf /usr/share/iscsi/iscsid.conf
+	insinto /usr/share/flatcar/etc/iscsi
+	newins - iscsid.conf
```
krnowak, line 6: Is this a canonical way to install an empty file? TIL.

krnowak (approved):

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-block/open-iscsi
chewi: The `-` is used to install a file from stdin. I just didn't pass anything to stdin here.

krnowak: Right. I've read that in the Gentoo PMS. I remember that when I wanted to install an empty file, I was doing `touch "${T}/empty-file"` and then do `newins "${T}/empty-file /whatever/path`. But this looks simpler. The only wrinkle is "In this case, it is an error if standard input is a terminal." - I guess it doesn't happen here.

chewi: That didn't happen, even when I ran `emerge` locally. I think Portage just prevents it.

PR #2268, opened 2024-09-01
Monthly GLSA metadata 2024-09-01

flatcar-infra:
Updated GLSA metadata


dongsupark (approved):
I thought https://security.gentoo.org/glsa/202408-07 might be a blocker, as Flatcar only has Go 1.21.
However it was not the case, because GLSA check is performed only under `/build/{amd64,arm64}-usr`, while Go is installed only in the SDK.
CI is expected to pass.

PR #2291, opened 2024-09-06
Publish stage 3 SDK tarball to use as a seed instead of stage 4

chewi:
# Publish stage 3 SDK tarball to use as a seed instead of stage 4

We currently use a stage 4 tarball as a seed for building the SDK, but stage 3 is smaller and less likely to cause issues.

This doesn't change the name of the tarball because supporting both stage3 and stage4 at the same time just isn't feasible with the way the digest check works.

The change is almost backwards compatible except that Catalyst will fail to build older baselayout versions due to stage3 lacking git. A recent change to the baselayout package sources it from a tarball instead of git, and this has now made it into a release. This issue only affects building the SDK, and it can be resolved by simply rebasing.

## How to use

You can't really try this change specifically, but you can build the SDK up to stage 3 and then manually use that tarball as a seed to build another SDK.

## Testing done

I have tested this manually as much as I can, but the change itself cannot be tested in the usual way because it involves publishing the tarball. The change is at least relatively small.

- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc. -- **N/A**

### DIFF
--- a/bootstrap_sdk
+++ b/bootstrap_sdk
@@ -89,9 +89,11 @@ cp "${BUILD_LIBRARY_DIR}/toolchain_util.sh" "${ROOT_OVERLAY}/tmp"
 
 catalyst_build
 
-if [[ "$STAGES" =~ stage4 ]]; then
+def_upload_path="${UPLOAD_ROOT}/sdk/${ARCH}/${FLAGS_version}"
+
+if [[ $STAGES =~ stage3 ]]; then
     info "Build complete! Changing output name to something more sensible."
-    build_name="stage4-${ARCH}-${FLAGS_version}.tar.bz2"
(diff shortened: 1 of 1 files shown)


krnowak:
The sdk tarball is also used for building SDK docker images. Did you test if the stage3 is usable here too? See `ci-automation/sdk_container.sh` and `build_sdk_container_image` scripts.

chewi:
Ah damn, I think that torpedoes the whole thing. I was previously told it was only used for the seed. If it used the "stage4" file directly while that was still around, that would be okay, but it looks like it can at least potentially fetch it from the bincache, so we'd have to publish both.

PR #2299, opened 2024-09-10
secure boot: lockdown, grub fallback, CI

jepio:
# secure boot: lockdown, grub fallback, CI
* import secureboot lockdown kernel patches. These are required by the upstream shim community. The alternative is to either always force lockdown (bad - no unsigned kernel modules) or do so through a kernel cmdline parameter in grub (frowned upon and harder to fix later). I took and rebased the patches from debian: https://sources.debian.org/src/linux/6.6.13-1~bpo12%2B1/debian/patches/features/all/lockdown/.
* enable grub fallback. this can be triggered if the kernel signature fails verification, and will go through gptprio selection and eventually fallback to the previous slot. fallback only supports numerical menuentry ids (source: https://github.com/coreos/grub/blob/2.02-coreos/grub-core/normal/menu.c#L350).
* CI: changes from @sayanchowdhury for use with https://github.com/flatcar/mantle/pull/556

## How to use

## Testing done

Tested here:
amd64: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/26834/cldsv/
arm64: http://jenkins.infra.kinvolk.io:8080/job/container/job/test/26780/cldsv/

### DIFF
--- a/ci-automation/vendor-testing/qemu.sh
+++ b/ci-automation/vendor-testing/qemu.sh
@@ -10,6 +10,9 @@ set -euo pipefail
 
 source ci-automation/vendor_test.sh
 
+SECUREBOOT=""
+ovmf_vars=""
+
 # ARM64 qemu tests only supported on UEFI
 if [ "${CIA_ARCH}" = "arm64" ] && [ "${CIA_TESTSCRIPT}" != "qemu_uefi.sh" ] ; then
     echo "1..1" > "${CIA_TAPFILE}"
@@ -21,7 +24,7 @@ if [ "${CIA_ARCH}" = "arm64" ] && [ "${CIA_TESTSCRIPT}" != "qemu_uefi.sh" ] ; th
     exit 1
 fi
 
-# Fetch image and BIOS if not present
+# Fetch image and firmware if not present
 if [ -f "${QEMU_IMAGE_NAME}" ] ; then
     echo "++++ ${CIA_TESTSCRIPT}: Using existing ${QEMU_IMAGE_NAME} for testing ${CIA_VERNUM} (${CIA_ARCH}) ++++"
 else
@@ -31,21 +34,30 @@ else
     lbunzip2 "${QEMU_IMAGE_NAME}.bz2"
 fi
 
-bios="${QEMU_BIOS}"
+firmware="${QEMU_FIRMWARE}"
 if [ "${CIA_TESTSCRIPT}" = "qemu_uefi.sh" ] ; then
-    bios="${QEMU_UEFI_BIOS}"
+    firmware="${QEMU_UEFI_FIRMWARE}"
+    ovmf_vars="${QEMU_UEFI_OVMF_VARS}"
 fi
 
 if [ "${CIA_TESTSCRIPT}" = "qemu_uefi_secure.sh" ] ; then
-    bios="${QEMU_UEFI_SECURE_BIOS}"
+    firmware="${QEMU_UEFI_SECURE_FIRMWARE}"
+    ovmf_vars="${QEMU_UEFI_SECURE_OVMF_VARS}"
+    SECUREBOOT=1
 fi
 
 if [ "${CIA_TESTSCRIPT}" = "qemu_uefi.sh" ] || [ "${CIA_TESTSCRIPT}" = "qemu_uefi_secure.sh" ] ; then
-    if [ -f "${bios}" ] ; then
-        echo "++++ ${CIA_TESTSCRIPT}: Using existing ${bios} ++++"
+    if [ -f "${firmware}" ] ; then
+        echo "++++ ${CIA_TESTSCRIPT}: Using existing ${firmware} ++++"
     else
-        echo "++++ ${CIA_TESTSCRIPT}: downloading ${bios} for ${CIA_VERNUM} (${CIA_ARCH}) ++++"
-        copy_from_buildcache "images/${CIA_ARCH}/${CIA_VERNUM}/${bios}" .
+        echo "++++ ${CIA_TESTSCRIPT}: downloading ${firmware} for ${CIA_VERNUM} (${CIA_ARCH}) ++++"
+        copy_from_buildcache "images/${CIA_ARCH}/${CIA_VERNUM}/${firmware}" .
+    fi
+    if [ -f "${ovmf_vars}" ] ; then
+        echo "++++ ${CIA_TESTSCRIPT}: Using existing ${ovmf_vars} ++++"
+    else
+        echo "++++ ${CIA_TESTSCRIPT}: downloading ${ovmf_vars} for ${CIA_VERNUM} (${CIA_ARCH}) ++++"
+        copy_from_buildcache "images/${CIA_ARCH}/${CIA_VERNUM}/${ovmf_vars}" .
     fi
 fi
 
@@ -68,11 +80,13 @@ kola run \
     --board="${CIA_ARCH}-usr" \
     --parallel="${QEMU_PARALLEL}" \
     --platform=qemu \
-    --qemu-bios="${bios}" \
+    --qemu-firmware="${firmware}" \
     --qemu-image="${QEMU_IMAGE_NAME}" \
     --tapfile="${CIA_TAPFILE}" \
+    "${ovmf_vars:+--qemu-ovmf-vars=${ovmf_vars}}" \
     ${QEMU_KOLA_SKIP_MANGLE:+--qemu-skip-mangle} \
     "${devcontainer_opts[@]}" \
+    ${SECUREBOOT:+--enable-secureboot} \
     "${@}"
 
 set +x
--- a/sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/files/6.6/z0004-efi-Add-an-EFI_SECURE_BOOT-flag-to-indicate-secure-b.patch
+++ b/sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/files/6.6/z0004-efi-Add-an-EFI_SECURE_BOOT-flag-to-indicate-secure-b.patch
@@ -1,0 +1,166 @@
+From 1e2ffbec195c89d887bc088691ebb19c9173ecad Mon Sep 17 00:00:00 2001
+From: David Howells <dhowells@redhat.com>
+Date: Mon, 18 Feb 2019 12:45:03 +0000
+Subject: [PATCH 1/4] efi: Add an EFI_SECURE_BOOT flag to indicate secure boot
+ mode
+
+UEFI machines can be booted in Secure Boot mode.  Add an EFI_SECURE_BOOT
+flag that can be passed to efi_enabled() to find out whether secure boot is
+enabled.
+
+Move the switch-statement in x86's setup_arch() that inteprets the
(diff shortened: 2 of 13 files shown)


sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/files/6.6/z0004-efi-Add-an-EFI_SECURE_BOOT-flag-to-indicate-secure-b.patch
```diff
@@ -0,0 +1,166 @@
+From 1e2ffbec195c89d887bc088691ebb19c9173ecad Mon Sep 17 00:00:00 2001
```
ader1990, line 1: These patches will be a pain to maintain, given that there is an agressive backport policy for the stable branches happening lately. But I do not see a way out either, other than trying to push for / advocating to have these patches merged in Linux master.

For example, the small pahole patch that we use has to updated at least a few times a year, these core patches probably more often and with way more care.

There are also great security implications for these patches and my suggestion is to have them peer reviewed from the broader Linux community by sending them upstream first.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/coreos-sources-6.6.48.ebuild
```diff
@@ -38,4 +38,8 @@ UNIPATCH_LIST="
 	${PATCH_DIR}/z0001-kbuild-derive-relative-path-for-srctree-from-CURDIR.patch \
 	${PATCH_DIR}/z0002-revert-pahole-flags.patch \
 	${PATCH_DIR}/z0003-Revert-x86-boot-Remove-the-bugger-off-message.patch \
+	${PATCH_DIR}/z0004-efi-Add-an-EFI_SECURE_BOOT-flag-to-indicate-secure-b.patch \
```
ader1990, line 41: If you can add please the initial source of the patches in the commit message too, so that we can have a better tracking of those (https://sources.debian.org/src/linux/6.1.94-1/debian/patches/features/all/lockdown/). Question: are there any newer sources for the 6.6 and 6.10 that we can leverage, or maybe an lkml thread?

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/files/6.6/z0004-efi-Add-an-EFI_SECURE_BOOT-flag-to-indicate-secure-b.patch
jepio: These patches have been posted and rejected upstream by the kernel community. Still, they are required to participate in the community shim signing process and all distros carry them. I agree that it they will occasionally break and require us to manually fix them up.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/coreos-sources-6.6.48.ebuild
ader1990: > I'll update the commit message. No newer lkml submissions, there is a 6.6 branch from debian https://sources.debian.org/src/linux/6.6.13-1~bpo12%2B1/debian/patches/features/all/lockdown/ but i think i took the 6.1.94 version, manually applied to 6.6.47 and then exported them.

It's great we can rely on a newer version and hopefully things won't break that often :).

jepio:
> LGTM besides concerns already raised. I'm surprised these patches aren't mainline already.

They will never be mainline.You can see the full discussion here: https://lore.kernel.org/lkml/4136.1522452584@warthog.procyon.org.uk/.

> As there is a work in progress and if the desire is to move to a new kernel soon-ish (https://github.com/flatcar/scripts/pull/2300), I would suggest to have these patches only for the 6.10 or the new LTS 6.12, and to not have to maintain them for 6.6 too (less work if we maintain only for very up-to-date Linux versions). And maybe in the meantime find a way to upstream / gather some support for mainline.

The aim is to submit a shim for signing in october so we can't risk it by coupling the required changes to other ones like the kernel upgrade. If we have to maintain patches on two kernel branches then so be it.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/files/6.6/z0004-efi-Add-an-EFI_SECURE_BOOT-flag-to-indicate-secure-b.patch
ader1990: My main concern is the following - once a week, there is a new patch level release for the linux kernel versions that we use (currently just 1 now - 6.6, might be in the future 6.12 or even faster 6.10) - There is a chance that the patches for secureboot have conflicts and inevitably, this will lead to two things: fast-forward conflict resolution or less often than once a week kernel upgrades. The first part is the most important - because these patches affect core kernel pathways, the changes need to be carefully crafted and reviewed so that there is less chance of a critical vulnerability.

The discussion I would like to start is - this feature is going to be used right away on all the clouds? or just on a subset and if there is just a subset, we can just silo it to that subset and in the worst case scenario, there is less chance to have something affecting all the images.

tl;dr if these patches are not properly reviewed and maintained, there is a chance we have less security instead of more.

ci-automation/vendor-testing/qemu.sh
```diff
@@ -31,21 +34,30 @@ else
...
+  ovmf_vars="${QEMU_UEFI_SECURE_OVMF_VARS}"
+  SECUREBOOT=1
 fi
 
 if [ "${CIA_TESTSCRIPT}" = "qemu_uefi.sh" ] || [ "${CIA_TESTSCRIPT}" = "qemu_uefi_secure.sh" ] ; then
-    if [ -f "${bios}" ] ; then
-        echo "++++ ${CIA_TESTSCRIPT}: Using existing ${bios} ++++"
+  if [ -f "${firmware}" ] ; then
+        echo "++++ ${CIA_TESTSCRIPT}: Using existing ${firmware} ++++"
+    else
```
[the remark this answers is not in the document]
jepio: you're right, fixed it

sayanchowdhury (approved):
LGTM

jepio:
@ader1990 are you OK with merging?

ader1990:
> @ader1990 are you OK with merging?

I suppose I am okay with merging, but I still think it is not an optimal way forward.

Because having this unreviewed (and soon rebased, conflicts need to be taken care of, etc) patches affect Flatcar images that will never be used in a secure boot environment, thus adding more possible security vulnerabilities. I suggested to have a more one-step at a time approach, activate this patch just for the OEMs that we know need the secure boot implementation now, see how it goes for a while, and then activate the patches for all the images.

I am saying this because some of the code from the patches executes independently of the kernel config option is set or not, which opens the possibility for unwanted behaviours.
If the patches were just an `if check_kernel_config: do something`, it would have been great. But it is not the case.

jepio:
> > @ader1990 are you OK with merging?
> 
> I suppose I am okay with merging, but I still think it is not an optimal way forward.
> 
> Because having this unreviewed (and soon rebased, conflicts need to be taken care of, etc) patches affect Flatcar images that will never be used in a secure boot environment, thus adding more possible security vulnerabilities. I suggested to have a more one-step at a time approach, activate this patch just for the OEMs that we know need the secure boot implementation now, see how it goes for a while, and then activate the patches for all the images.
> 
> I am saying this because some of the code from the patches executes independently of the kernel config option is set or not, which opens the possibility for unwanted behaviours. If the patches were just an `if check_kernel_config: do something`, it would have been great. But it is not the case.

The patches don't do anything unless secure boot is enabled at runtime:
```
 		case efi_secureboot_mode_enabled:
 			set_bit(EFI_SECURE_BOOT, &efi.flags);
+#ifdef CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT
+			lock_kernel_down("EFI Secure Boot",
+					 LOCKDOWN_INTEGRITY_MAX);
+#endif
```
So lockdown is only triggered in a secure boot environment.

We can't ship a different kernel configuration for different OEMs because that would require preparing separate update payloads and nebraska can't handle that. These patches are part of standard Ubuntu, Debian, Fedora, RHEL, SuSe kernels so they have been well tested.

I would also rather not have them.

ader1990:
> > > @ader1990 are you OK with merging?
> > 
> > 
> > I suppose I am okay with merging, but I still think it is not an optimal way forward.
> > Because having this unreviewed (and soon rebased, conflicts need to be taken care of, etc) patches affect Flatcar images that will never be used in a secure boot environment, thus adding more possible security vulnerabilities. I suggested to have a more one-step at a time approach, activate this patch just for the OEMs that we know need the secure boot implementation now, see how it goes for a while, and then activate the patches for all the images.
> > I am saying this because some of the code from the patches executes independently of the kernel config option is set or not, which opens the possibility for unwanted behaviours. If the patches were just an `if check_kernel_config: do something`, it would have been great. But it is not the case.
> 
> The patches don't do anything unless secure boot is enabled at runtime (the switch checks whether firmware set SECUREBOOT=1)
> 
> ```
>  		case efi_secureboot_mode_enabled:
>  			set_bit(EFI_SECURE_BOOT, &efi.flags);
> +#ifdef CONFIG_LOCK_DOWN_IN_EFI_SECURE_BOOT
> +			lock_kernel_down("EFI Secure Boot",
> +					 LOCKDOWN_INTEGRITY_MAX);
> +#endif
> ```
> 
> So lockdown is only triggered in a secure boot environment.
> 
> We can't ship a different kernel configuration for different OEMs because that would require preparing separate update payloads and nebraska can't handle that. These patches are part of standard Ubuntu, Debian, Fedora, RHEL, SuSe kernels so they have been well tested.
> 
> I would also rather not have them.

[efi_get_fdt_params](https://github.com/flatcar/scripts/pull/2299/commits/8393a4cf4bd8c2b813870d9ae6dff3c8a0512e8e#diff-f01a8ceae996f2c99bea7f89a5daba891d4b8764dd2ac61ad37edf72452b4a32R53) is changed, also some small other places in the process. 

If other operating systems use the same byte-wise patches, I suppose it is a smaller risk. Let's hope for the best.

I see, for example, that 6.10 also has the same level of support for the patches: https://sources.debian.org/data/main/l/linux/6.10.6-1~bpo12+1/debian/patches/features/all/lockdown/, which is great, as we can copy those over byte-wise. Maybe we could add a small .github ci automation to get these patches automatically on a kernel change, like we do with the gentoo upstream - to have the same source of truth Flatcar and Debian.

ader1990 (approved):

ader1990:
Leaving this here for future usage and reference:

https://sources.debian.org/data/main/l/linux/6.6.13-1~bpo12%2B1/debian/patches/features/all/lockdown/efi-lock-down-the-kernel-if-booted-in-secure-boot-mo.patch vs https://sources.debian.org/data/main/l/linux/6.10.6-1~bpo12+1/debian/patches/features/all/lockdown/efi-lock-down-the-kernel-if-booted-in-secure-boot-mo.patch

Seems that the rebasing work is already done, although it lags on a few patch releases.

For: https://github.com/flatcar/scripts/pull/2300

PR #2301, opened 2024-09-10
sys-boot/grub: Move to portage-stable, bump to 2.12, add Red Hat patches

chewi:
# grub: Move to portage-stable, 2.12 bump, add RH patches

We can now use Gentoo's upstream ebuild, save for a few small overrides in a separate env file.

This bumps GRUB from 2.06 to 2.12, but with Red Hat's large patch set applied, which is needed for Secure Boot. The existing two Flatcar patches have been rebased. Gentoo's patches are discarded because they conflict and are not relevant to Flatcar.

## How to use

Build an image and try it out. Alternatively, take an existing image, replace grubx64.efi in the first partition, and test with UEFI. You can build grubx64.efi by installing sys-boot/grub to the host and the board root (`--nodeps` should be okay), tweaking the grub_install.sh script, and running the following command.

```diff
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -113,23 +113,9 @@ cleanup() {
 trap cleanup EXIT
 
 info "Installing GRUB ${FLAGS_target} in ${FLAGS_disk_image##*/}"
-LOOP_DEV=$(sudo losetup --find --show --partscan "${FLAGS_disk_image}")
 ESP_DIR=$(mktemp --directory)
 MOUNTED=
 
-for (( i=0; i<5; ++i )); do
-    if sudo mount -t vfat "${LOOP_DEV}p1" "${ESP_DIR}"; then
-        MOUNTED=x
-        break
-    fi
-    warn "loopback device node ${LOOP_DEV}p1 still missing, reprobing..."
-    sudo blockdev --rereadpt "${LOOP_DEV}"
-    # sleep for 0.5, then 1, then 2, then 4, then 8 seconds.
-    sleep "$(bc <<<"scale=1; (2.0 ^ ${i}) / 2.0")"
-done
-if [[ -z ${MOUNTED} ]]; then
-    failboat "${LOOP_DEV}p1 where art thou? udev has forsaken us!"
-fi
 sudo mkdir -p "${ESP_DIR}/${GRUB_DIR}"
 
 info "Compressing modules in ${GRUB_DIR}"
@@ -143,7 +129,7 @@ info "Generating ${GRUB_DIR}/load.cfg"
 # by filesystem ID in case the platform doesn't provide the boot disk.
 # $root points to memdisk here so instead use hd0,gpt1 as a hint so it is
 # searched first.
-ESP_FSID=$(sudo grub-probe -t fs_uuid -d "${LOOP_DEV}p1")
+ESP_FSID=XXXX-XXXX # Get this from guestfish with blkid /dev/sda1
 sudo_clobber "${ESP_DIR}/${GRUB_DIR}/load.cfg" <<EOF
 search.fs_uuid ${ESP_FSID} root hd0,gpt1
 set prefix=(memdisk)
```

```
sudo ./build_library/grub_install.sh --target x86_64-efi --verity --copy_efi_grub .
```

Instructions on how to update the Red Hat patches are included in the README.

## Testing done

A Jenkins run passes and local testing with QEMU amd64 works. The amd64 image is about 621KB larger, with most of the increases in /boot, which is very concerning, but not entirely surprising.


changelog/security/2024-09-06-grub-2.12.md
jepio, line 1: all of these are already listed in:
- changelog/security/2023-08-24-grub.md
- changelog/security/2023-10-12-grub-update.md

jepio:
Do you have local changes to `build_library/grub.cfg` - something equivalent to what was there before commit abb6bdb682:
```
# UEFI uses linuxefi/initrdefi instead of linux/initrd except for arm64
if [ "$grub_platform" = efi ]; then
  if [ "$grub_cpu" != arm64 ]; then
    set suf="efi"
  fi
fi
```

chewi:
> Do you have local changes to `build_library/grub.cfg` - something equivalent to what was there before commit [abb6bdb](https://github.com/flatcar/scripts/commit/abb6bdb682078ee0d6306c0d5176effeba0719aa)?

Can you expand on this? I haven't changed the config yet, but I was vaguely aware that there's `linuxefi` now as well as `linux`. I don't really understand the difference given that it supports EFI already.

chewi:
Looking at the code, I think `linuxefi` is just an alias for `linux` that ensures you don't accidentally boot without EFI when that module can't be loaded.

```c
  cmd_linux =
    grub_register_command ("linux", grub_cmd_linux,
                           0, N_("Load Linux."));
  cmd_linuxefi =
    grub_register_command ("linuxefi", grub_cmd_linux,
                           0, N_("Load Linux."));
```

chewi:
We're discussing whether the Red Hat patches are actually needed in flatcar/Flatcar#630, but FWIW, Jenkins has actually passed now.

chewi:
The amd64 image is now 621KB larger, with most of the increases in /boot, which is very concerning. Dropping the Red Hat patches may help, but I suspect there will still be an increase. Coupled with potential increases to the initrd (which I'm trying to lessen), this could be a real issue.

chewi:
Closing in favour of #2318.

PR #2311, opened 2024-09-12
app-admin/etcd-wrapper: bump etcd version

tormath1:
required for Kubernetes 1.31 external etcd 

Noticed here: https://github.com/flatcar/mantle/pull/557
```
[ERROR ExternalEtcdVersion]: this version of kubeadm only supports external etcd version >= 3.5.11-0. Current version: 3.5.0
```

NOTE: This does not mean that Kubernetes 1.31 does not work on Flatcar, it means that "out of the box" you can't use shipped etcd from Flatcar as external etcd node.   

---

- [x] CI: http://localhost:8080/job/container/job/packages_all_arches/4681/cldsv/


changelog/updates/2024-09-12-etcd.md
```diff
@@ -0,0 +1 @@
+- etcd ([3.5.16](https://github.com/etcd-io/etcd/releases/tag/v3.5.16))
```
krnowak, line 1: Etcd has some wacky changelog situation - the tagged commit has no changelog, the release branches have no changelog. The changelogs for all the release branches are in the main branch, so let's point there instead.

```suggestion
- etcd ([3.5.16](https://github.com/etcd-io/etcd/blob/96e17d38e244ef18756a55d33066f2893402794b/CHANGELOG/CHANGELOG-3.5.md#v3516-2024-09-10))
```

krnowak (approved):
Looks good, but let's use a different link to changelog.

PR #2319, opened 2024-09-13
board-packages: add hyperv hv-daemons

t-lo:
This change adds the Azure and HyperV OEM "hv-daemons" to board-packages so build_packages.sh will actually build these. This un-breaks a build issue with the Azure and HyperV images.

It's a one-line change to un-break the nightly build and Azure testing.

Needs to be cherry-picked to flatcar-3975, flatcar-4054, flatcar-4081.


t-lo:
Merged & cherry-picked for  flatcar-3975, flatcar-4054, flatcar-4081..

krnowak:
The only thing I'd like to point out (for the future) is to bump the package revision (so it should go from r15 to r16).

PR #2326, opened 2024-09-16
.github: Follow-up fix for missing app-emulation/hv-daemons

krnowak:
One more place where we were mentioning app-emulation/hv-daemons unconditionally.


krnowak:
Now it works.

PR #2333, opened 2024-09-17
sys-boot/grub: Adjust version string to include flatcar1 suffix

chewi:
# sys-boot/grub: Adjust version to include flatcar1 suffix

We will bump this every time we or Gentoo change patches that modify parts of GRUB that are installed to the boot partition. We will reset the version back to 1 when the upstream GRUB version changes.

Without this, we are bound by Gentoo's PVR string, which we cannot change when we need to make changes to our own patches. The shim review wants to know the full version number of our GRUB build, and it would look bad to make such changes without changing the version.

This suffix is also applied to the Flatcar entry in the SBAT, which is especially important for the shim review.

The published binary package will still be labelled with the Gentoo PVR, but that seems less important given that end users cannot update individual packages on Flatcar installations.

## How to use

Boot the image from CI and see what GRUB version it shows when booting.

## Testing done

I've tested the image built in CI.

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
+++ b/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
@@ -1,10 +1,20 @@
+# Bump the flatcar version stated here every time we or Gentoo change patches
+# that modify parts of GRUB that are installed to the boot partition. Reset the
+# version back to 1 when the upstream GRUB version changes.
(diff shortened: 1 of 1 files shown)


chewi:
I've now tested an image. This is what it looks like.

image

chewi:
Hmmm, what happens with binary packages if they have the same PVR? Do they get overwritten? If so, that's probably okay. If not, that's a problem.

sayanchowdhury (approved):

PR #2342, opened 2024-09-27
pkg-auto: Add package automation scripts

krnowak:
This adds some scripts I have been using for over a year to deal with the weekly package updates.

It comes with a `README.md` which describes a workflow similar to my own.

The `sync_packages.sh` and `update_packages.sh` scripts are currently not used anywhere. The idea behind them was to use them for Github Action, but that will come as a follow-up PR.


pkg_auto/impl/util.sh
```diff
@@ -0,0 +1,253 @@
...
+    echo
+}
+
+# Prints help. Help is taken from the lines prefixed with double
+# hashes in the top sourcer of this file.
+function print_help() {
+    if [[ ${THIS} != "${BASH}" ]]; then
+        grep '^##' "${THIS}" | sed -e 's/##[[:space:]]*//'
+    fi
+}
```
chewi, line 146: I guessed you must have done this before I saw the function. "Oh you didn't... oh you did." :rofl: It's fine. :grin:

chewi (changes requested):
I haven't gone over this in detail as you've already walked us through it, and it's already proven itself over months. Your Bash code is also some of the best I've seen. Just need to fix the README and I think you're good to go.

krnowak:
Dropped some lines from README and added a paragraph about security fixes at the end of the `Processing the updates (the manual part)` part.

chewi (approved):

PR #2349, opened 2024-09-30
sys-kernel/coreos-modules: build Intel iGPU module

ader1990:
See: https://github.com/flatcar/Flatcar/issues/1306

This PR adds support for the Intel built-in GPU (CONFIG_DRM_I915=m)
## Testing done


sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.6
```diff
@@ -11,6 +11,8 @@ CONFIG_ACPI_HOTPLUG_MEMORY=y
 CONFIG_ACPI_IPMI=m
 CONFIG_ACPI_PCI_SLOT=y
 CONFIG_ACPI_PROCESSOR_AGGREGATOR=y
+CONFIG_AGP=y
+CONFIG_AGP_INTEL=y
```
jepio, line 15: AGP is a bus that predates PCI express, and hasn't been in use for 20 years. I would not enable it in our config.

jepio (commented):
comment about AGP

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.6
ader1990: I did enable it 'just in case',  because I saw this documentation here: https://github.com/torvalds/linux/blob/master/drivers/char/agp/Kconfig#L74

Sure, will disable the config if the module is not used or useful. Also saw that Ubuntu and other distros have the same configs enabled, but probably for legacy reasons.

ader1990:
The resulting image needs testing on a real environment, currently I do not have one to test on.

Also, the passthrough implementation using qemu or another hypervisor will most likely need extra kernel config options options enabled, extra kernel parameteres enabled and possibly initrd changes.

Found these two documents that might be of help:
https://3os.org/infrastructure/proxmox/gpu-passthrough/gpu-passthrough-to-vm/#proxmox-configuration-for-gpu-passthrough
https://3os.org/infrastructure/proxmox/gpu-passthrough/igpu-split-passthrough/

Thanks.

ader1990:
The size increase in /usr is around 13MB, and the initrd increase is almost zero, as the driver is built as module and not loaded in the initrd stage.

The increase in /usr is not negligible, and consists of firmware binaries in /usr/lib/firmware/i915.

The AMD64 compressed image has increased to around 590MB from 585MB.

Any thoughts or opinions on the size increase? Another option would be a systemd-sysext, similar to the zfs one.

savely-krasovsky:
@jepio are there any news on merging that PR? Just completed my migration from CoreOS and was quite shocked to see that Flatcar doesn't ship i915 driver. It's crucial for homelab setups with hardware encoding (Plex, Jellyfin, etc).

jepio (approved):

savely-krasovsky:
Thanks!

PR #2351, opened 2024-10-01
Upgrade Linux Kernel for main from 6.6.52 to 6.6.53

flatcar-infra:
Subject says it all.


changelog/updates/2024-10-01-linux-6.6.53-update.md
```diff
@@ -0,0 +1 @@
+- Linux ([6.6.53](https://lwn.net/Articles/989982))
```
dongsupark, line 1: That link is completely wrong, not sure why. It should be actually https://lwn.net/Articles/992296/.

PR #2355, opened 2024-10-02
coreos-modules: Enable kernel-module for mpi3mr

sfudeus:
# Enable kernel module mpi3mr for MPI3MR I/O & RAID controllers

This enables the kernel module mpi3mr, which is required e.g. on modern DELL hardware (PERC gen12, e.g. H965 raid controllers). 
The driver is part of the linux kernel since 5.2 from 2019.

## Testing done

Currently building and testing on affected Dell bare-metal servers. 


sfudeus:
The failed checks at least partially seem to complain about not finding my commitid, do the checks maybe not support forks? And in which test outputs would I find infos about size changes? `Generate reports against last ...` does not really contain relevant output AFAICS.

jepio:
The failed checks are expected, don't worry bout them. There is an image-size-change report in the CI logs that someone will review and then this is good to go (we'll also want to add a short changelog so that its easy to spot which release has this change).

ader1990:
> Hello @sfudeus, I will rerun the image build, as yesterday there was a rather big decrease in the initrd size because of another PR, and we neede a new build so that the comparison can be made correctly.

The new results are available:

```
./boot/flatcar/vmlinuz-a by 114688 bytes (112 kbytes) from 57391608 bytes (56046 kbytes, 54 mbytes) to 57506296 bytes (56158 kbytes, 54 mbytes)
```

There is a 112 KB increase of the /boot partition because of the expected increase in vmlinuz-a.

sfudeus:
> ```
> ./boot/flatcar/vmlinuz-a by 114688 bytes (112 kbytes) from 57391608 bytes (56046 kbytes, 54 mbytes) to 57506296 bytes (56158 kbytes, 54 mbytes)
> ```
> 
> There is a 112 KB increase of the /boot partition because of the expected increase in vmlinuz-a.

Thanks for looking into this and the data. 
Sounds reasonable and acceptable, doesn't it?

How do we go from here, ideally even getting this then backported to the current stable line (3975) or at least beta (4054)?

ader1990 (approved):

PR #2359, opened 2024-10-04
sys-kernel: enable kernel virtio vnc output

ader1990:
Fix for: https://github.com/flatcar/Flatcar/issues/1553

Tested locally with qemu-kvm instances on ARM64 and the logs show up early during the boot, as expected.


chewi:
Not that it's a bad idea, but why do you need SIMPLEDRM? I found that the existing kernel works with QEMU's `-display virtio-gpu-pci`, but you just don't see anything between GRUB and pivot root, presumably because the module is not in the initrd.

chewi:
I also wonder why CONFIG_EFI_FB isn't enough on arm64. That's what amd64 normally uses given that we don't pass `-device virtio-vga` there.

ader1990:
> I also wonder why CONFIG_EFI_FB isn't enough on arm64. That's what amd64 normally uses given that we don't pass `-device virtio-vga` there.

made some tests and verified the minimal patch, and it seems that DRM and VIRTIO_GPU is needed at the same time (makes sense as they are in the same config tree).

chewi (approved):

PR #2370, opened 2024-10-11
Upgrade Linux Kernel for main from 6.6.54 to 6.6.56

flatcar-infra:
Subject says it all.


dongsupark (approved):

ader1990 (approved):
Let's merge it, and then check if https://github.com/flatcar/Flatcar/issues/1565 might be solved by the 6.6.55 nfs patches.

PR #2373, opened 2024-10-11
Add eclass/usr-ldscript.eclass to package automation

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1781/cldsv/

--

It couldn't be added before to automation, because the eclass in Gentoo would introduce have unwanted side-effects into our built images. But the Gentoo version of the eclass is essentially a no-op when USE=split-usr is disabled. We have recently moved to use Gentoo profiles that disable USE=split-usr altogether, and with this move, we can now safely put the eclass into automation.

This marks the point where the entirety of the portage-stable is under automation.

Closes https://github.com/flatcar/Flatcar/issues/645.
Closes https://github.com/flatcar/Flatcar/issues/646.

--

### DIFF
--- a/.github/workflows/portage-stable-packages-list
+++ b/.github/workflows/portage-stable-packages-list
@@ -461,9 +461,7 @@ eclass/tree-sitter-grammar.eclass
 eclass/udev.eclass
 eclass/unpacker.eclass
 eclass/user-info.eclass
-# This file is modified by us to be an empty file, so can't be synced for now.
-#
-# eclass/usr-ldscript.eclass
+eclass/usr-ldscript.eclass
 eclass/vcs-clean.eclass
 eclass/vcs-snapshot.eclass
 eclass/verify-sig.eclass
--- a/sdk_container/src/third_party/portage-stable/eclass/usr-ldscript.eclass
+++ b/sdk_container/src/third_party/portage-stable/eclass/usr-ldscript.eclass
@@ -1,1 +1,172 @@
-# Empty file
+# Copyright 2019-2023 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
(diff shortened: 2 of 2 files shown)


krnowak:
This does not seem to be exactly a no-op change, though. But the changes seem to be minor - previously some packages were installing `/usr/lib64/libfoo.so.1.2.3` (being a proper dynamic library), `/usr/lib64/libfoo.so.1` (being a symlink to the `libfoo.so.1.2.3`), and `/usr/lib64/libfoo.so` (being some linker script). Now `/usr/lib64/libfoo.so` is also a symlink to `libfoo.so.1.2.3` instead.

The linker script was fixing some linking issues that stemmed from the fact that some dynamic libraries were stored in `/lib`, whereas static libraries - in `/usr/lib`. This fix is clearly irrelevant, because we only have `/usr/lib`, and `/lib` is a symlink to the former. This means that the fix was for profiles that enabled `USE=split-usr`. We are not using such profiles anymore.

The reason that this change occurred now is because some ebuilds are calling `gen_usr_ldscript` function (which created those linker scripts, clobbering the `libfoo.so -> libfoo.so.1.2.3` symlink) and this PR added the `usr-ldscript.eclass` which brings in the second definition of the function (and it clobbers the first one), and this function became essentially a no-op because it does an early return if `USE=split-usr` is disabled. The first definition came from `toolchain-funcs.eclass', it's documented as deprecated and doesn't have the shortcut for `USE=split-usr` being disabled. So earlier, `gen_usr_ldscript` from `toolchain-funcs.eclass` was in use and it was generating linker scripts, now `gen_usr_ldscript` from `usr-ldscript.eclass` is in use and it does nothing.

krnowak:
CI passed.

dongsupark (approved):

PR #2431, opened 2024-11-06
sys-boot/grub: Apply Red Hat's large patch set (and drop Gentoo's)

chewi:
# Apply and fix Red Hat's GRUB patches, drop Gentoo's patches

We initially thought we would need Red Hat's patch set. Then it looked like we wouldn't because the TPM Event Log appeared to work without it. We later discovered that on amd64, it only works with Secure Boot disabled. The patch set also fixes Secure Boot on arm64, which would have otherwise needed a couple of patches from Canonical.

Red Hat's patches break the fallback mechanism, so apply additional patches to fix that. See rhboot/grub2#195.

We have to drop Gentoo's patches because they conflict, but they don't affect Flatcar anyway.

## How to use

Check that the various images successfully boot.

## Testing done

This [Jenkins SDK run](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1816/cldsv/) passed successfully. /boot on amd64 has grown by 450KB, which isn't ideal, but not unexpected after the last time we tried adding these patches. We're still just under 50% usage. I've just noticed that we're installing test GRUB modules for some reason, so I'll see whether we can drop those.


changelog/changes/2024-11-06-grub-2.12-flatcar3.md
```diff
@@ -0,0 +1 @@
+- grub 2.12-flatcar3: GRUB now includes many patches from Red Hat to support Secure Boot, as well as Flatcar's own patches. The version string includes a numbered "flatcar" suffix to track changes to these additional patches. This string can be seen in the GRUB menu.
```
jepio, line 1: link to PR from the changelog

sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-boot/grub/README.md
```diff
@@ -15,6 +15,29 @@ to use a separate repo was scrapped, and two patch files were created. The patch
...
+be merged soon.
+
+## How to import the Red Hat patches
+
+Red Hat maintains a fork of GRUB on GitHub with branches for each Fedora release. Generate a diff between the latest upstream release and the latest Fedora branch.
+
+```
+git clone https://github.com/rhboot/grub2.git grub
+cd grub
+git diff grub-<VERSION>..fedora-<VERSION> -- . ':(exclude).gitignore' ':(exclude)bootstrap.conf' > grub-2.12-00-redhat.patch
```
ader1990, line 38: Future work: maybe add some automatation in the github actions to update this patch on a regular basis

sayanchowdhury (approved):

ader1990 (approved):
Ran a k8s full deployment locally, works as expected.

sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-boot/grub/README.md
chewi: I'm hoping it'll get mainlined eventually, but we'll see.

sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-boot/grub/grub-2.12-01-execute-return-code.patch
```diff
@@ -0,0 +1,112 @@
...
+--- a/grub-core/script/execute.c
++++ b/grub-core/script/execute.c
+@@ -935,7 +935,10 @@ grub_script_execute_sourcecode (const char *source)
+ 	  break;
+ 	}
+ 
+-      ret = grub_script_execute (parsed_script);
++      /* Don't let trailing blank lines determine the return code.  */
++      if (parsed_script->cmd)
++	ret = grub_script_execute (parsed_script);
```
dongsupark, line 40:
```suggestion
+        ret = grub_script_execute (parsed_script);
```
It should be spaces instead of tab.

chewi: I was following the existing style, as weird as it is. I've submitted this to Red Hat, so they can flag this up if needs be.

PR #2434, opened 2024-11-08
Switch from raw to QCOW2 and 2MB to 4MB firmware, adjust firmware variables creation

chewi:
# Switch QCOW2 to 4MB firmware, adjust variables creation

The new arm64 firmware supporting Secure Boot is in QCOW2 format only, avoiding the extra space taken up by the 64MB padding. Supporting both raw and QCOW2 images would be messy, so switch entirely to QCOW2.
    
Only the 4MB images are in QCOW2 format on amd64, so also switch away from the 2MB images. 4MB images are now the default for most distributions as they are needed to apply certain Windows updates.

Rather than starting with a blank variables image, reuse the image that already has the Microsoft certificates and the latest DBX revocation list applied. Gentoo also applies the Red Hat certificates, which we don't need, but this is okay.

Drop the PK and KEK certificates as these are only needed when you are going to ship DB updates to existing systems, which we are not going to do.

## How to use

Check that the various images still boot under QEMU, including amd64 with Secure Boot enabled.

## Testing done

The [Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1820/) passed, including qemu_uefi_secure.


build_library/vm_image_util.sh
```diff
@@ -857,20 +852,29 @@ _write_qemu_uefi_conf() {
...
-    cp "/usr/share/edk2-ovmf/OVMF_CODE.secboot.fd" "$(_dst_dir)/${flash_ro}"
+
+    case $BOARD in
+        amd64-usr)
+            cp "/usr/share/edk2/OvmfX64/OVMF_CODE_4M.secboot.qcow2" "$(_dst_dir)/${flash_ro}"
+            flash_in="/usr/share/edk2/OvmfX64/OVMF_VARS_4M.secboot.qcow2"
+            ;;
+        arm64-usr)
+            cp "/usr/share/edk2/ArmVirtQemu-AARCH64/QEMU_EFI.secboot_INSECURE.qcow2" "$(_dst_dir)/${flash_ro}"
+            flash_in="/usr/share/edk2/ArmVirtQemu-AARCH64/QEMU_VARS.secboot_INSECURE.qcow2"
```
krnowak, line 870: A short comment about the INSECURE secureboot stuff thing here would be nice. Possibly along the lines of "insecure because yadda yadda, but nothing to worry about, used in qemu only" or somesuch.

krnowak (approved):

build_library/vm_image_util.sh
chewi: Done! I also added another note to the changelog and the mokutil commit since it goes better with this PR than the next one.

krnowak (approved):

PR #2441, opened 2024-11-12
Add (temporary) signed shim and sign official builds with Azure Key Vault

chewi:
# Add (temp) signed shim and sign official builds with AKV

This adds the SBAT data to the shim binary, adds a pre-signed shim package for official builds, adds the tools needed to sign with Azure Key Vault, and reworks the signing code to do that.

Flatcar has not yet completed the shim review, so the signed shim is a temporary one that will be replaced shortly.

The Azure Key Vault signing only works when performed on Azure. This happens during the **images** job, which is currently still configured to run on Equinix. This configuration lives in another repository. It should be adjusted before we need to perform the next official build. We may split out the signing part of the job so that the actual building can still be performed on Equinix, but that has yet to be agreed.

azure-keyvault-pkcs11 is still undergoing a cleanup following its transformation from AWS, so the commit currently used by the package points to an early iteration, but the plugin is sufficiently functional to be used here.

## How to use

The changes mostly concern CI, but you can check that the flatcar_production_qemu_uefi_secure_efi image still works with Secure Boot enabled.

## Testing done

I will now perform an unofficial build with Jenkins. We have already seen that an official build works when the **images** job is run on Azure.

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) -- **The change is probably only worth mentioning to users when we add the officially signed shim.**


sayanchowdhury (approved):
Looks good to me - but let @krnowak / @dongsupark / @tormath1 do a review.

build_library/sbsign_util.sh
```diff
@@ -0,0 +1,55 @@
...
+PKCS11_MODULE_PATH="/usr/$(get_sdk_libdir)/pkcs11/azure_kms_pkcs11.so"
+
+PKCS11_ENV=(
+    AZURE_KEYVAULT_URL="https://chewi-test.vault.azure.net/"
+    PKCS11_MODULE_PATH="${PKCS11_MODULE_PATH}"
+    AWS_KMS_PKCS11_DEBUG=1
+)
+
+get_sbsign_cert() {
+    if [[ ${SBSIGN_KEY} != pkcs11:* || -s ${SBSIGN_CERT} ]]; then
```
krnowak, line 22: I think this will fail with `set -euo pipefail` for unofficial builds, because SBSIGN_CERT is undefined:

```
$ set -euo pipefail
$ SBSIGN_KEY="pkcs11:token=flatcar-dev-cert"
$ unset SBSIGN_CERT
$ if [[ ${SBSIGN_KEY} != pkcs11:* || -s ${SBSIGN_CERT} ]]; then echo 'early return'; else echo 'get the cert'; fi
bash: SBSIGN_CERT: unbound variable
```

So maybe:
```suggestion
    if [[ ${SBSIGN_KEY} != pkcs11:* || -s ${SBSIGN_CERT-} ]]; then
```

sdk_container/src/third_party/coreos-overlay/app-crypt/p11-kit/README.md
krnowak, line 1: Cool, another package moved from overlay to portage-stable. Thanks!

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```diff
@@ -120,3 +120,11 @@
...
 # Needed by updated app-containers/containers-common
 =sys-fs/fuse-overlayfs-1.14 ~amd64 ~arm64
+
+# azure-keyvault-pkcs11
+=app-crypt/azure-keyvault-pkcs11-0_p20240923 ~amd64
+=app-crypt/p11-kit-0.25.5 ~amd64
+=dev-cpp/azure-core-1.14.0 ~amd64
+=dev-cpp/azure-security-keyvault-certificates-4.2.1 ~amd64
+=dev-cpp/azure-security-keyvault-keys-4.4.1 ~amd64
+=dev-cpp/azure-identity-1.10.0 ~amd64
```
krnowak, line 130: Could you keep those entries sorted with the rest of the file? (Or is it something frowned upon in Gentoo? I suppose that Gentoo has conventions we are not really following here.)

Also, I'd add the arm64 keywords too, even if still don't have the arm64 SDK yet.

sdk_container/src/third_party/coreos-overlay/sys-boot/shim/shim-15.8-r1.ebuild
krnowak, line 1: Should we bump the revision to r2 here?

Also I see that the version of sys-boot/shim-signed matches a version of sys-boot/shim package (ignoring the revision number). Are there any steps that we need to take when we update shim to a newer version? If so, probably would be good to leave a comment in the ebuild with those steps or some link or at least a note to remember to regenerate the signed binaries and to keep the versions of both packages in sync.

chewi: This is best answered by @sayanchowdhury. It's not something we would do often because we need to go through a new shim review every time we want to update the signed shim.

sayanchowdhury: > Should we bump the revision to r2 here?

I actually have that in the todo for the parent PR, which is getting split inot smaller PRs - so let's bump it.

> 
> Also I see that the version of sys-boot/shim-signed matches a version of sys-boot/shim package (ignoring the revision number). Are there any steps that we need to take when we update shim to a newer version? If so, probably would be good to leave a comment in the ebuild with those steps or some link or at least a note to remember to regenerate the signed binaries and to keep the versions of both packages in sync.

We could write a small note in the shim README, on how we keep the version in sync, and if there is a update that means we would always be prepping for a shim-review.

sayanchowdhury: To add, we would be creating a entry in flatcar docs on Secure Boot maintenance, so I would add a small snippet in a README for now. Later I can elaborate the README or add a link to the main flatcar docs as required.

build_library/sbsign_util.sh
```diff
@@ -0,0 +1,55 @@
...
+    SBSIGN_CERT="/usr/share/sb_keys/shim.pem"
+else
+    SBSIGN_KEY="pkcs11:token=flatcar-dev-cert"
+    unset SBSIGN_CERT
+fi
+
+PKCS11_MODULE_PATH="/usr/$(get_sdk_libdir)/pkcs11/azure_kms_pkcs11.so"
+
+PKCS11_ENV=(
+    AZURE_KEYVAULT_URL="https://chewi-test.vault.azure.net/"
```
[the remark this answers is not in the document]
chewi: I've just spoken with @sayanchowdhury. We're going to leave this as-is for now, but I'll temporarily nobble the code so that signing doesn't actually happen for official builds yet. We don't want to be blocked from doing releases before the shim review is complete.

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
chewi: I decided to stabilise the SDK packages since it's been long enough. I'll move the other entries.

As previously mentioned, that's not what `~amd64 ~arm64` actually does. :wink: In any case, we haven't tested any of this on arm64 yet, and it wouldn't make much sense because this tooling is only used in a CI job that will only be run on Azure. Running that as arm64 would just cost more. :laughing:

sdk_container/src/third_party/coreos-overlay/sys-boot/shim/shim-15.8-r1.ebuild
chewi: I've bumped the revision.

I'll leave Sayan to add the README to this PR.

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
krnowak: > I decided to stabilise the SDK packages since it's been long enough. I'll move the other entries.

Cool.

> As previously mentioned, that's not what `~amd64 ~arm64` actually does. 😉

Yeah, I need to wrap my head around how this really works. But for atoms like `=foo-bar/pkg-1.2.3 ~amd64 ~arm64` works exactly the way we expect. ;) For other atom forms, it probably different and less obvious.

> In any case, we haven't tested any of this on arm64 yet, and it wouldn't make much sense because this tooling is only used in a CI job that will only be run on Azure. Running that as arm64 would just cost more. 😆

You'd need to add those keywords anyway if you wanted to build the arm64 SDK regardless of whether you want to use those tools or not - I think that building would fail as there wouldn't be any package version available to emerge. Anyway, it's water under the bridge now, because you stabilized them. ;)

sdk_container/src/third_party/coreos-overlay/sys-boot/shim/shim-15.8-r1.ebuild
chewi: Sayan has now added this.

krnowak (approved):

PR #2456, opened 2024-11-18
Free up /boot space with some GRUB module changes

chewi:
# Free up /boot space with some GRUB module changes

Secure Boot prevents you from loading additional GRUB modules so remove them from supported targets to save space. These modules could be useful for debugging with Secure Boot disabled, but manually copying the modules with debug symbols is even more useful and not that difficult.

Compress the GRUB modules with xz instead of gzip to save space. Giving the `--best` or `-9` option results in a heavier decompression cost with no gain on such small files. Note that this required [a Kola test to be dropped](https://github.com/flatcar/mantle/pull/565).

Prevent the developer test GRUB modules from being built. These are normally always installed, even by `grub-install`, but they have no use outside of testing and take up valuable space in /boot. Apologies for the use of awk here, but sed didn't cut it. Kudos to Copilot for the assistance. :smile: 

/boot usage has changed from 62MB to 60MB on amd64. arm64 remains at 62MB, despite the modules previously taking up 940KB. I suspect all these usage figures are actually quite close to 61MB. The unrounded figures would give a better picture.

We still need to be cautious about the kernel size for the time being because older systems being upgraded will not benefit from these savings. It would probably be easy to backport the changes into the updater though, so I'll look into that.

## How to use

Just test that the images boot as usual.

## Testing done

A [Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1834/) passed except for the cl.update.grubnop test, which has since been dropped as mentioned above.

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
+++ b/sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
@@ -12,6 +12,32 @@ cros_pre_src_prepare_adjust_version() {
 	sed -i "/AC_INIT/s/\b${PV//./\\.}\b/\0-${FLATCAR_VERSION}/g" configure.ac || die
 }
 
+# Prevent developer test modules from being built. These are normally always
+# installed, even by grub-install, but they have no use outside of testing and
+# take up valuable space in /boot. The best way to identify these is to look for
+# references to the tests/ directory.
+cros_post_src_prepare_drop_tests() {
+	gawk -i inplace '
+		/^module = \{/ {
+			in_mod = 1
+		}
+		in_mod {
+			block = block $0 "\n"
+		}
+		/^\};/ && in_mod {
+			if (block !~ /\<tests\//) {
+				printf "%s", block
+			}
+			in_mod = 0
+			block = ""
+			next
+		}
+		!in_mod {
+			print
+		}
+	' grub-core/Makefile.core.def || die
+}
+
 # Replace Gentoo's SBAT with Flatcar's.
 cros_post_src_install_sbat() {
 	insinto /usr/share/grub
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -41,7 +41,7 @@ switch_to_strict_mode
 GRUB_DIR="flatcar/grub/${FLAGS_target}"
 
 # Modules required to boot a standard CoreOS configuration
-CORE_MODULES=( normal search test fat part_gpt search_fs_uuid gzio search_part_label terminal gptprio configfile memdisk tar echo read btrfs )
+CORE_MODULES=( normal search test fat part_gpt search_fs_uuid xzio search_part_label terminal gptprio configfile memdisk tar echo read btrfs )
 
 SBAT_ARG=()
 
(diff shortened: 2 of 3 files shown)


sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
```diff
@@ -12,6 +12,32 @@ cros_pre_src_prepare_adjust_version() {
 	sed -i "/AC_INIT/s/\b${PV//./\\.}\b/\0-${FLATCAR_VERSION}/g" configure.ac || die
 }
 
+# Prevent developer test modules from being built. These are normally always
+# installed, even by grub-install, but they have no use outside of testing and
+# take up valuable space in /boot. The best way to identify these is to look for
+# references to the tests/ directory.
+cros_post_src_prepare_drop_tests() {
+	gawk -i inplace '
```
krnowak, line 20: Does `-i` have the same meaning in `gawk` as it has in `sed`? Checking `gawk` manual tells me that `-i` is a shortcut for `--include`, so here you are doing an include of a file called `inplace`? Is there a chance that you actually are not modifying the Makefile? The image diff reports would tell you if some module files were gone from initrd, but I can't find those reports anywhere.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
```diff
@@ -12,6 +12,32 @@ cros_pre_src_prepare_adjust_version() {
...
+		/^module = \{/ {
+			in_mod = 1
+		}
+		in_mod {
+			block = block $0 "\n"
+		}
+		/^\};/ && in_mod {
+			if (block !~ /\<tests\//) {
```
krnowak, line 28: What's the meaning of `\<` escape sequence?

chewi: It's awk's equivalent of `\b`, a word boundary.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
chewi: The inplace feature is implemented through an additional gawk module that you need to load with `-i`. It's not built-in like it is with sed. I know it's working because the image diff shows this and more:

```
23:57:31  @@ -232,22 +221,15 @@
23:57:31   /boot/flatcar/grub/i386-pc/search_part_uuid.mod
23:57:31   /boot/flatcar/grub/i386-pc/sendkey.mod
23:57:31   /boot/flatcar/grub/i386-pc/setjmp.mod
23:57:31  -/boot/flatcar/grub/i386-pc/setjmp_test.mod
23:57:31   /boot/flatcar/grub/i386-pc/setpci.mod
23:57:31   /boot/flatcar/grub/i386-pc/sfs.mod
23:57:31  -/boot/flatcar/grub/i386-pc/shift_test.mod
23:57:31  -/boot/flatcar/grub/i386-pc/signature_test.mod
23:57:31   /boot/flatcar/grub/i386-pc/sleep.mod
23:57:31  -/boot/flatcar/grub/i386-pc/sleep_test.mod
```

krnowak (approved):

PR #2477, opened 2024-11-25
Drop lots of obsolete CI, release signing, and upload code

chewi:
# Drop lots of obsolete CI, release signing, and upload code

The jenkins directory has been superseded by code in the ci-automation directory.

Most of the rest hinges on the `--upload` option being passed, and it never is any more. Much of it also concerns Google Buckets, which we no longer use, save for some GCE-specific bits.

## How to use

Nothing to use. The real acid test well be when we do an official release.

## Testing done

A [two-phase Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1845/) entirely passed, including the additional amd64 tests. I compared the file listings in bincache with the latest nightly and everything matched up, save for a few very minor size differences.


build_library/vm_image_util.sh
```diff
@@ -1224,7 +1224,7 @@ EOF
     "version": "${FLATCAR_VERSION_ID}",
     "providers": [{
       "name": "${provider}",
-      "url": "$(download_image_url "$(_dst_name ".box")")",
+      "url": "$(_dst_name ".box")",
```
krnowak, line 1227: I can see that this change is basically a noop due to `download_image_url` being short-circuited when upload is disabled, thus printing just the passed parameter. This means that the `url` field was an invalid field for a long time and no one complained about it. We probably could think about using bincache or release server url, but I guess it is not important anyway, so whatever.

core_date
krnowak, line 1: I'd imagine the release team using this script. @sayanchowdhury?

core_dev_sign_update
krnowak, line 1: This seems to be indeed unused - it still refers to `coreos_production` stuff.

signing/devel.key.pem
krnowak, line 1: I have no idea about the stuff in `signing` directory at all. Maybe release team knows about it. @sayanchowdhury?

krnowak (commented):
Looks alright for me, but it would be good if the release team has a look at it too.

build_library/vm_image_util.sh
chewi: You mean it's not really a URL? Yeah, I could probably put the bincache address here. I think it would have to be hardcoded though, none of the surrounding code is aware of that address. It's usually set by CI and only used for uploading at the end of each job.

core_date
chewi: Maybe, but I couldn't find a single reference to it anywhere.

core_dev_sign_update
chewi: Indeed, there are no Flatcar references at all.

signing/devel.key.pem
chewi: I couldn't find any references, and every function in transfer.sh except `devsign` uses public.update.core-os.net. `devsign` calls core_dev_sign_update, which you agree seems unused.

build_library/vm_image_util.sh
krnowak: Yeah, maybe it could be something like `https://${BUILDCACHE_SERVER:-bincache.flatcar-linux.net}/images/${BOARD%-usr}/${FLATCAR_VERSION}`?

core_date
krnowak: Right, I'd assume that this could be invoked manually when, for example, doing a release planning. So planning to do a release on third week of January, which likely means kicking off builds on January 13th, which means that the version would be:

```
$ ./core_date --date 2025-01-13
Unable to determine location for common.sh.  If you are sourcing
common.sh from a script run via dash or sh, you must do it in the
following way:
  COMMON_SH="$(dirname "$0")/../../scripts/common.sh"
  . "${COMMON_SH}"
where the first line is the relative path from your script to
common.sh.
```

Ok, you know what? Just drop it.

krnowak (approved):

PR #2484, opened 2024-11-26
sys-kernel/coreos-modules: Enable VFIO for arm64

lukasstockner:
# Enable VFIO for arm64

VFIO was configured as x86-only, but it also works and is useful on ARM.
CONFIG_VFIO_PCI_VGA is x86-specific, so it remains in the amd64 config.

## How to use

- Build and run Flatcar image
- modprobe vfio_pci
- Bind a PCI device to vfio-pci

## Testing done


jepio (approved):
Thank you for this contribution! We'll let CI pass and then this can be merged.

lukasstockner:
Thanks for the quick review!
Looks like the CI failed, but I don't see how the error is related to the change. Is this something you've seen before?

ader1990:
> Thanks for the quick review! Looks like the CI failed, but I don't see how the error is related to the change. Is this something you've seen before?

Hello, the CI failed because it could not use the code from your branch, it is expected when making a PR from a fork.

PR #2491, opened 2024-12-02
Do Secure Boot signing for official builds in a separate additional job

chewi:
# Do SB signing for official builds in a separate additional job

We only want to do the signing in Azure, not the whole image job. This new job downloads the unsigned image, signs it, and replaces it. This new job will only be invoked once flatcar/jenkins-os#354 has been merged.

We had temporarily nobbled the Azure signing codepath, but now the only thing we need to nobble here is copying the signed shim. In the meantime, we copy the unsigned shim instead. We will revert this new temporary commit once the shim review is complete.

This PR also includes a couple of clean ups I made along the way. I have dropped the "modify_image" code, which is completely unused. I could have possibly used it for this new job, but it did more than I needed it to. I have also deduplicated the build_image `--extract_update` and `--generate_update` options because they were just totally confusing as they were.

## How to use

Do a Jenkins run using the jenkins-os chewi/sbsign_image branch. You need to "replay" the image job and tweak the `is_official` check to force it to run the new job. You also need to tweak `COREOS_OFFICIAL=0` to `1` in ci-automation/sbsign_image.sh to force the job to sign using AKV rather than the dev key.

## Testing done

A [Jenkins SDK run](http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1859/) was performed and everything passed. I manually tweaked the run as described above to test the new job against AKV. I also grabbed one of the images to check that list of certificates found on vmlinuz-a and grubx64.efi using sbverify.


sbsign_image
```diff
@@ -0,0 +1,84 @@
...
+if [[ -e "${BUILD_DIR}" ]]; then
+  if [[ ${FLAGS_replace} -eq ${FLAGS_TRUE} ]]; then
+    sudo rm -rf "${BUILD_DIR}"
+  else
+    error "Directory ${BUILD_DIR} already exists."
+    error "Use --build_attempt option to specify an unused attempt."
+    error "Or use --replace if you want to overwrite this directory."
+    die "Unwilling to overwrite ${BUILD_DIR}."
+  fi
+fi
```
jepio, line 75: `sbsign_image` shouldn't be enforcing a clean build dir, much like `image_to_vm.sh`

sbsign_image
```diff
@@ -0,0 +1,84 @@
...
+
+# Create the output directory and temporary mount points.
+mkdir -p "${BUILD_DIR}"
+
+fix_mtab
+sbsign_prod_image "$1" "${FLATCAR_PRODUCTION_IMAGE_NAME}" "${FLAGS_disk_layout:-base}"
```
jepio, line 81: can you search for the input and produce the output in `BUILD_DIR/...` by default?

build_library/build_image_util.sh
```diff
@@ -894,3 +882,70 @@ EOF
...
+  local disk_img="${BUILD_DIR}/${image_out_name}"
+  local EFI_ARCH
+
+  case "${FLAGS_board}" in
+    amd64-usr) EFI_ARCH="x64" ;;
+    arm64-usr) EFI_ARCH="aa64" ;;
+  esac
+
+  # Modify a copy of the image, but reflink for efficiency.
+  cp --reflink=auto "${image_in_path}" "${disk_img}"
```
jepio, line 904: Could you operate on the image in-place instead.

build_library/prod_image_util.sh
```diff
@@ -225,3 +234,27 @@ create_prod_sysexts() {
...
+sbsign_prod_image() {
+  local image_in_path="$1"
+  local image_out_name="$2"
+  local disk_layout="$3"
+
+  info "Signing production image ${image_in_path} for Secure Boot as ${image_out_name}"
+  local root_fs_dir="${BUILD_DIR}/rootfs"
+  local image_kernel="${image_out_name%.bin}.vmlinuz"
+  local image_pcr_policy="${image_out_name%.bin}_pcr_policy.zip"
+  local image_grub="${image_out_name%.bin}.grub"
```
jepio, line 247: Use a local variable for the shared prefix

ci-automation/sbsign_image.sh
```diff
@@ -0,0 +1,97 @@
...
+    local docker_sdk_vernum="$(vernum_to_docker_image_version "${FLATCAR_SDK_VERSION}")"
+
+    docker_image_from_registry_or_buildcache "${sdk_name}" "${docker_sdk_vernum}"
+    local sdk_image="$(docker_image_fullname "${sdk_name}" "${docker_sdk_vernum}")"
+    echo "docker image rm -f '${sdk_image}'" >> ./ci-cleanup.sh
+
+    ./run_sdk_container -x ./ci-cleanup.sh -v "${FLATCAR_VERSION}" -U -C "${sdk_image}" \
+        ./sbsign_image --board="${arch}-usr" \
+                       --group="${channel}" --version="${FLATCAR_VERSION}" \
+                       --output_root="/mnt/host/source/src/scripts/${images_local%%/*}" \
```
jepio, line 88: Other jobs use `$CONTAINER_IMAGE_ROOT` for this

ci-automation/sbsign_image.sh
```diff
@@ -0,0 +1,97 @@
...
+                       --replace --only_store_compressed
+
+    # Delete uncompressed generic image before signing and upload
+    rm "${images_local}/flatcar_production_image.bin"
```
jepio, line 92: Shouldn't `--only_store_compressed` handle this?

jepio (commented):
It would be great if this followed the convention also worked when executed sequentially on a single machine like this:
```
./build_image
./sbsign_image
./image_to_vm.sh --format=qemu_uefi
```

ci-automation/sbsign_image.sh
chewi: Nope.

```bash
            # If requested, delete the input file after compression (only if 'none' is not part of the formats)
            # Exclude the generic image and update payload because they are needed for generating other formats
            if [ "${FLAGS_only_store_compressed}" -eq "${FLAGS_TRUE}" ] &&
               [ "${filename##*/}" != "flatcar_production_image.bin" ] &&
```

chewi: Which is why the existing image.sh also has a similar line.

ci-automation/sbsign_image.sh
chewi: ...which seems pointless? In vms.sh, if fires up a container to move the input into this containerised directory, does the work, and then fires up another container to move the output back to the host. Given that these containers are short-lived, why not just read from and write to the host in the first place?

sbsign_image
chewi: Modifying a given image in-place but writing other files to the build directory would be odd, so I will do as you suggest while dropping the argument entirely.

ci-automation/sbsign_image.sh
jepio: But that's still the host:
```
CONTAINER_IMAGE_ROOT="/home/sdk/trunk/src/build/images"
```
```
~/trunk/src/scripts $ readlink -f /home/sdk/trunk/src/build/
/mnt/host/source/src/build
```
```
run_sdk_container:142:        -v "${PWD}/__build__/images:/mnt/host/source/src/build"
```

sbsign_image
chewi: I've now dropped this so that it modifies in-place.

jepio (commented):
Lgtm but my vote should count for 0.5 😅 
So another maintainer should also review.

build_library/build_image_util.sh
```diff
@@ -61,57 +61,43 @@ delete_prompt() {
...
-  declare -a files_to_evaluate
-  files_to_evaluate+=( "${update}.bin" )
-  compress_disk_images files_to_evaluate
+zip_update_tools() {
+  # There isn't a 'dev' variant of this zip, so always call it production.
+  local update_zip="flatcar_production_update.zip"
+
+  info "Generating update tools zip"
+  # Make sure some vars this script needs are exported
+  export REPO_MANIFESTS_DIR SCRIPTS_DIR
```
krnowak, line 98: As a cleanup, this could be:
```suggestion
  local -x REPO_MANIFESTS_DIR SCRIPTS_DIR
```

build_library/build_image_util.sh
```diff
@@ -894,3 +882,66 @@ EOF
...
+  local pcr_policy="$5"
+  local image_grub="$6"
+
+  local disk_img="${BUILD_DIR}/${image_name}"
+  local EFI_ARCH
+
+  case "${FLAGS_board}" in
+    amd64-usr) EFI_ARCH="x64" ;;
+    arm64-usr) EFI_ARCH="aa64" ;;
+  esac
```
krnowak, line 900: Maybe add `*) die "Unknown board ${BOARD@Q}" ;;` here, so @ader1990 will know about another place to modify for the riscv support. ;)

krnowak (approved):

PR #2501, opened 2024-12-03
Update SSSD, move to portage-stable

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5083/cldsv/

Closes https://github.com/flatcar/Flatcar/issues/1489

--

- sys-auth/sssd: [PROD] [DEV]
  - from 2.3.1-r6 to 2.9.5
  - EAPI 8
  - fixes CVE-2023-3758
    - https://github.com/flatcar/Flatcar/issues/1489
  - dropped most of our modifications, the ones that are still needed are moved to config overrides or user patches
    - systemd sssd.service is replaced by upstream, but it has similar modifications to what we have done
    - tmpfiles config file is gone - it only had entries for /etc and /var
      - /etc entries are not needed - we now have /etc as an overlay
        - the only thing needed was to have a default config made from an example config, and it's done in config overrides
      - /var entries are not needed - they are generated during image build
    - stop modifying dependencies
      - pulls in net-dns/bind-tools instead of net-dns/bind…
      - enabled gssapi on net-dns/bind-tools (used to have issues on arm64)
      - enabled winbind in net-fs/samba (no idea why it was dropped before)
    - retained modifications:
      - pam config moved from /etc to /usr (config override)
      - create a default sssd config in /etc (config override)
      - cross-compilation fix for a broken openldap check (user patch)
      - enable nss lookups with sss plugin even if sssd is not running (user patch)
  - dropped USE=+locator, enabled unconditionally
  - dropped USE=pac, folded into USE=samba
  - dropped USE=valgrind
  - added USE=subid, something to do with subuid and subgid allocation, disabled
  - added USE=systemtap, for enabling instrumentation, disabled
  - replaced a dependency on dev-libs/libpcre with dev-libs/libpcre2
    - this makes dev-libs/libpcre unused
  - dropped a dependency on net-libs/http-parser
    - this makes net-libs/http-parser unused
  - added required USE=experimental on net-nds/openldap

- net-dns/bind-tools:
  - added back, dependency of sys-auth/sssd
    - it's temporary, the package is deprecated, so it will be eventually replaced with net-dns/bind

- net-nds/openldap:
  - enabled USE=experimental, required by sys-auth/sssd

- net-fs/samba:
  - enabled USE=winbind, required by sys-auth/sssd

- profiles:
  - dropped unused oem-aci profiles

--


krnowak:
First CI run succeeded, but I kicked off another one, since I added another user-patch and dropped two now-unused packages.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-auth/sssd
```diff
@@ -1 +1,11 @@
-export ac_cv_member_struct_ldap_conncb_lc_arg=no
```
chewi, line 1: Wouldn't it be better to just keep this instead of adding the patch? Or if you're confident about this check being redundant, I could just add this line to Gentoo.

sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-auth/sssd/0001-enable-nss-lookups.patch
chewi, line 1: How about `EXTRA_ECONF="--enable-sss-default-nss-plugin"` instead?

.github/workflows/portage-stable-packages-list
```diff
@@ -502,6 +501,7 @@ net-dialup/lrzsz
 net-dialup/minicom
 
 net-dns/bind
+net-dns/bind-tools
```
chewi, line 504: I would have just changed it to `bind` in the sssd ebuild. The automation scripts will undo this, but you can easily skip that change for a short time.

chewi (changes requested):

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-auth/sssd
krnowak: I wanted to switch it to "yes", because our version of OpenLDAP has the `lc_arg` member in `struct ldap_conncb` and using it seems to allow SSSD to track some referrals or something (in general this seems to be preferred). But the "yes" case has a runtime check for a bug that was fixed 16 years ago. And we know that runtime checks and cross-compilation do go along.

I think that long-term my preferences would be in this order:

1. Upstream to drop the runtime check, just like my user-patch is doing.
2. Upstream adding a `--enable-ldap-conncb-i-know-what-i-am-doing` flag that skips the runtime check.
3. Gentoo taking my patch.

.github/workflows/portage-stable-packages-list
krnowak: Welp, I'm wondering how short this "short time" will be. I can see that `net-dns/bind-tools` ebuilds are quite usable for versions <9.18.0, i.e. they are not wrappers around net-dns/bind, so my impression is that as long as there are ebuilds for 9.16.x, net-dns/bind-tools will stay. And people using sys-auth/sssd together with net-dns/bind-tools-9.16.x (because they masked bind 9.18.x) would be upset if you just changed the dependency from bind-tools to bind. Maybe we could ask for updating the dependency in sssd to be `<net-dns/bind-tools-9.17 || >net-dns/bind-9.17`?

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-auth/sssd
chewi: Ah, I had it backwards. In that case, a better option may be to populate the 4th `AC_RUN_IFELSE` argument, which is currently `[]`, to simply assume `HAVE_LDAP_CONNCB=1` when cross-compiling. This seems like a reasonable compromise given how old that release is now. Upstream may be more willing to do that than drop the check entirely.

.github/workflows/portage-stable-packages-list
chewi: My guess is they're just waiting for this [stabilisation bug](https://bugs.gentoo.org/941696) so that they can just point things directly at net-dns/bind and eventually drop all the old versions. I don't feel strongly about it so you choose.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-auth/sssd
krnowak: I have filed a PR to SSSD, let's see how it goes - https://github.com/SSSD/sssd/pull/7743

chewi (approved):
You can wait on the outcome of our sssd patch discussion, but this is fine as-is in any case.

PR #2502, opened 2024-12-03
sysext: turn vim to a sysext image

tormath1:
In this PR, we propose to turn `vim` to a sysext image built without `minimal` compiling option. The sysext will be enabled by default to not break existing Vim usage.

* Folks interested to have vim on Flatcar will enjoy the full experience
* Folks not interested to have vim on Flatcar (and its regular amount of [CVEs](https://github.com/flatcar/Flatcar/issues?q=is%3Aissue+vim+label%3Aadvisory+)) will be able to remove it

This sysext is called `tools` and can handle more software inside, see:
> We can rather bundle similar things into a feature set, and give it an appropriate name (E.g., we wanted a dev extension with gcc, gdb and other stuff). Here we could call it tools and later on we also add an optional extra-tools extension for anything else we want to add that wasn't part of the base image (E.g., htop, nano?). [^1]

[^1]: https://github.com/flatcar/scripts/pull/2502#discussion_r2314870321

One could drop Vim from Flatcar with the current configuration:
```yaml
variant: flatcar
version: 1.1.0
storage:
  files:
    - path: /etc/flatcar/enabled-sysext.conf
      overwrite: true
      contents:
        inline: |
          -tools
```

Note: In case of emergency, a `nano` editor is now shipped on the generic image and will be always available (even if `tools` sysext is unmerged) 
 
## How to use

Use `vim` as usual:
```
core@localhost ~ $ systemd-sysext status
HIERARCHY EXTENSIONS         SINCE
/opt      none               -
/usr      containerd-flatcar Tue 2024-12-03 14:33:23 UTC
          docker-flatcar
          flatcar-tools
          oem-qemu
core@localhost ~ $ vim --version | grep -i huge
Huge version without GUI.  Features included (+) or not (-):
core@localhost ~ $ nano --version
 GNU nano, version 8.5
 (C) 2025 the Free Software Foundation and various contributors
 Compiled options: --disable-justify --disable-libmagic --disable-nls --disable-utf8
core@localhost ~ $ sudo systemd-sysext unmerge
Unmerged '/usr'.
core@localhost ~ $ vim --version
-bash: /usr/bin/vim: No such file or directory
core@localhost ~ $ nano --version
 GNU nano, version 8.5
 (C) 2025 the Free Software Foundation and various contributors
 Compiled options: --disable-justify --disable-libmagic --disable-nls --disable-utf8
```

- [x] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc. (` Total size difference: decreased by 9833049 bytes (9602 kbytes, 9 mbytes)
`)
- [ ] CI: http://localhost:8080/job/container/job/packages_all_arches/6575/cldsv/

## Note 
* sysext size with vim huge:
```
$ sudo systemd-dissect ./flatcar-tools.raw
 File Name: flatcar-tools.raw
      Size: 9.9M
 Sec. Size: 512
     Arch.: n/a

Image Name: flatcar-tools
 sysext R.: ID=flatcar
            VERSION_ID=9999.0.0
            ARCHITECTURE=x86-64

    Use As: ✗ bootable system for UEFI
            ✗ bootable system for container
            ✗ portable service
            ✗ initrd
            ✓ sysext for system
            ✓ sysext for portable service
            ✗ sysext for initrd
            ✗ confext for system
            ✗ confext for portable service
            ✗ confext for initrd

RW DESIGNATOR PARTITION UUID PARTITION LABEL FSTYPE   ARCHITECTURE VERITY GROWFS PARTNO
ro root       -              -               squashfs -            -      no          -
```

Related to:
* https://github.com/flatcar/Flatcar/discussions/1463
* https://github.com/flatcar/sysext-bakery/pull/78

### DIFF
--- a/changelog/changes/2025-09-03-sysext-vim.md
+++ b/changelog/changes/2025-09-03-sysext-vim.md
@@ -1,0 +1,1 @@
+- `vim` is now provided as part of a `tools` sysext image which is enabled by default ([scripts#2502](https://github.com/flatcar/scripts/pull/2502))


sdk_container/src/third_party/coreos-overlay/coreos-base/misc-files/misc-files-0-r7.ebuild
```diff
@@ -207,4 +207,7 @@ src_install() {
     # (The below keepdir will result in a tmpfiles entry in base_image_var.conf)
     keepdir /var/kubernetes/kubelet-plugins/volume/exec
     dosym /var/kubernetes/kubelet-plugins/volume/exec /usr/libexec/kubernetes/kubelet-plugins/volume/exec
+
+    insinto "/etc/flatcar"
+    doins "${FILESDIR}/flatcar/enabled-sysext.conf"
```
pothos, line 212: This should go into `/usr/share/flatcar/enabled-sysext.conf`

build_library/extra_sysexts.sh
```diff
@@ -2,4 +2,5 @@ EXTRA_SYSEXTS=(
   zfs:sys-fs/zfs
   podman:app-containers/podman,net-misc/passt
   python:dev-lang/python,dev-python/pip
+  vim:app-editors/vim
```
pothos, line 5: I would like to avoid having too many extensions and treat it like packages. We can rather bundle similar things into a feature set, and give it an appropriate name (E.g., we wanted a dev extension with gcc, gdb and other stuff). Here we could call it `tools` and later on we also add an optional `extra-tools` extension for anything else we want to add that wasn't part of the base image (E.g., `htop`, `nano`?).

sdk_container/src/third_party/coreos-overlay/coreos-base/coreos/coreos-0.0.1.ebuild
```diff
@@ -104,7 +104,6 @@ RDEPEND="${RDEPEND}
 	app-crypt/gnupg
 	app-crypt/go-tspi
 	app-crypt/tpmpolicy
-	app-editors/vim
```
pothos, line 107: To state what was written elsewhere: We might want to still include an emergency editor in the base image. Something like `nano` is good because it's common, user friendly, and minimal (and many distros have it as default editor).

pothos: Unrelated comment: I think it would be nice to have mksquashfs and mkfs.erofs available to build sysext images on Flatcar itself, maybe they could also directly go to a tools extension instead of the base image. Not sure if `tools` or optional `extra-tools` would be better - in doubt `extra-tools` to keep the default small?

pothos:
Yes :) I think in a follow-up PR after testing that it works with the "tools" extension (Is nightly enough or would we need to test it with an Alpha release? Not sure if we need to wait that long.)

PR #2504, opened 2024-12-06
sys-kernel/coreos-modules: Enable CONFIG_GTP

chewi:
# sys-kernel/coreos-modules: Enable CONFIG_GTP

This is the GPRS Tunneling Protocol datapath for usage in telecoms scenarios. It has been requested by a user.

## How to use

We probably can't use this kernel module ourselves, but we should at least check that it loads.

## Testing done

I'll run the GHA pipeline and check the image.


tormath1 (commented):
Looks good to me, does it increase the size of the kernel?

katt-999:
Hello @tormath1 

I'm the initiator of this feature request, after checking on my side, the module seems to weigh something like 16K

image

PR #2527, opened 2024-12-13
accept_keywords: pull qemu-guest-agent for arm64 arch.

tormath1:
For some reasons, `~arm64` or `arm64` keyword is not set for this version of qemu-guest-agent.

---

When cherry-picking this: https://github.com/flatcar/scripts/pull/2490 to LTS, I did not tested and I was not expecting the `arm64` keyword to be missing from the ebuild. Sorry about that.

## Testing done

```
emerge-arm64-usr -pv qemu-guest-agent
```


chewi:
Ah, that's a cross issue. We've changed how we avoid such issues since the last LTS. :cry:

PR #2583, opened 2025-01-17
Upgrade Containerd in main from 1.7.23 to 2.0.2

flatcar-infra:
Subject says it all.


krnowak:
We have switched to containerd from Gentoo, so this is not applicable anymore.

PR #2607, opened 2025-01-23
Drop unused packages

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/1920/cldsv/

This is mostly about dropping unused packages and cleaning up related entries in profiles.

I also added two group packages to the production images - they should have been there, but probably something stopped pulling them at some point, so I'm bringing them back. I hope to use them to eventually get rid of the passwd and group files from baselayout.


krnowak:
CI passed.

.github/workflows/portage-stable-packages-list
```diff
@@ -127,8 +127,6 @@ app-crypt/mhash
 app-crypt/mit-krb5
 app-crypt/p11-kit
 app-crypt/pinentry
-app-crypt/rhash
```
chewi, line 130: rhash is a dependency of CMake, so something is amiss here?

krnowak: Bah, well spotted, thanks. Fortunately this does not warrant another CI run.

krnowak: Fixed.

chewi (approved):
I was a little surprised to see udisks being dropped, but it does appear to be more of a desktop thing. It is used by [fwupd](https://fwupd.org/) though, which got me wondering whether we should include that for bare metal users. Probably not something you should run in a container.

chewi:
Eh, scratch that, it's written in Python.

PR #2636, opened 2025-02-05
OOT modules sign

danzatt:

For out of tree modules (like ZFS or NVIDIA) to work with secureboot, they need to be signed by the ephemeral kernel modules key. This key is shredded after the upstream-included kernel modules are built, therefore it can't be reused during ZFS module build. This PR moves the key to /tmp, so that it stays in RAM and can be reused by out of tree modules. Moreover, by moving the key to `/tmp` we improve the security of the ephemeral module signing key (previously we wrote it to disk and then shredded it, but it might still stay in the disk or software cache, compromising the secure boot model).

Currently, this PR works when the packages are built manually in the order coreos-modules, zfs-kmod and coreos-kernel. We need to fix the dependecies, so that we enforce this order.

## How to use

## Testing done


chewi (changes requested):
I mentioned this on Teams, so perhaps you've initially taken a shortcut, but using a deterministic path in /tmp is dangerous. We should generate a random one with `mktemp` and set an environment variable.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.6.76.ebuild
```diff
@@ -61,9 +61,21 @@ src_prepare() {
...
 	local sig_key="$(getconfig MODULE_SIG_KEY)"
-	mkdir -p "build/${sig_key%/*}" || die
-	cp -v "${KV_OUT_DIR}/${sig_key}" "build/${sig_key}" || die
+
+	if [ "$sig_key" == "${sig_key#/tmp/}" ]
+	then
+		die "Refusing to use module key stored outside of /tmp."
+	fi
+
+	# keeping the old logic here for now, unreacheble due to the previous condition
```
chewi, line 72: Why keep it? This is git! :wink:

sdk_container/src/third_party/coreos-overlay/eclass/coreos-kernel.eclass
```diff
@@ -136,20 +136,44 @@ getconfig() {
 	echo "${value}"
 }
 
+get_sig_key() {
+	local sig_key="$(getconfig MODULE_SIG_KEY)"
+
+	if [ "$sig_key" == "${sig_key#/}" ]
+	then
+		echo "build/$sig_key"
```
chewi, line 144: Why do this when you're going to error out with the /tmp check anyway?

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.6.76.ebuild
```diff
@@ -61,9 +61,21 @@ src_prepare() {
...
 	cp -v "${KV_OUT_DIR}/.config" build/ || die
+
 	local sig_key="$(getconfig MODULE_SIG_KEY)"
-	mkdir -p "build/${sig_key%/*}" || die
-	cp -v "${KV_OUT_DIR}/${sig_key}" "build/${sig_key}" || die
+
+	if [ "$sig_key" == "${sig_key#/tmp/}" ]
+	then
+		die "Refusing to use module key stored outside of /tmp."
+	fi
```
chewi, line 70: Can't you move the /tmp check into `get_sig_key()` and call that here to avoid the duplication?

sdk_lib/sdk_entry.sh
```diff
@@ -49,6 +49,13 @@ sed -i -r '/^masters =/s/\bcoreos(\s|$)/coreos-overlay\1/g' /usr/local/portage/c
     fi
 )
 
+# SDK container is launched in another shell, so we need to smuggle the variables inside
```
chewi, line 52: Can you explain this a bit more? Why isn't just calling `export` here enough?

chewi (changes requested):

run_sdk_container
```diff
@@ -150,6 +151,7 @@ if [[ -z ${stat} ]] ; then
         --network host
         -e SDK_USER_ID="$(id -u)"
         -e SDK_GROUP_ID="$(id -g)"
+	-e MODULE_SIGNING_KEY_DIR="/tmp/$(uuidgen)"
```
chewi, line 154: The whitespace is off here, but I think it might be better to set this in sdk_entry.sh
anyway.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.6.76.ebuild
```diff
@@ -61,9 +61,8 @@ src_prepare() {
 	# Pull in the config and public module signing key
 	KV_OUT_DIR="${SYSROOT%/}/lib/modules/${COREOS_SOURCE_NAME#linux-}/build"
 	cp -v "${KV_OUT_DIR}/.config" build/ || die
+
 	local sig_key="$(getconfig MODULE_SIG_KEY)"
```
chewi, line 65: I said you could call `get_sig_key` here, but now it's not actually used below, so this line should go. I'm not sure whether you need something in its place. I guess it builds the modules first and will reuse the config from that, so you don't need to call `envsubst` again.

chewi (approved):

chewi (changes requested):
Arrgh, sorry, just noticed one more thing. It isn't enough for `/tmp/$(uuidgen)` to be random. It actually needs to be created with `mktemp` in order to be safe. Hopefully that isn't a problem.

danzatt:
> Arrgh, sorry, just noticed one more thing. It isn't enough for `/tmp/$(uuidgen)` to be random. It actually needs to be created with `mktemp` in order to be safe. Hopefully that isn't a problem.

I fixed this now

chewi (approved):
The shell quoting is still a bit wonky, so please use shellcheck in future. It's good enough though. Thanks!

sayanchowdhury:
Can you please create PR for changelog or add an entry with the nvidia sysext PR?

PR #2640, opened 2025-02-07
Pass image version to kola

krnowak:
The SELinux denial checking in kola is conditionally disabled - if the Flatcar image version is too old, then it's disabled for all the tests. For kola to decide whether to disable or enable the checking, it needs to know the version of the image. It can be done in two ways - either we pass the version to kola through a flag, or kola itself spins a temporary cluster with one machine and figures out the version from os-release file. The latter is obviously very slow, to a point that kola was skipping this step if the tests didn't need the version information. Now the version information is needed always, and our ci-automation already knows the version of the image, so we can pass it to kola to skip creating the temporary cluster.

Requires https://github.com/flatcar/mantle/pull/572. Will update this PR later, when mantle PR is merged.


krnowak:
Will recreate the PR - beta has gotten a new branch.

PR #2668, opened 2025-02-14
Fix non conforming partition table

gabriel-samfira:
# Fix non conforming partition table

This change removes the legacy_boot flag from the EFI system partition. We already have a BIOS boot partition which should offer compatibility with legacy bios systems.

## How to use

Upload the image in a fresh OpenStack devstack deploy and boot. 

## Testing done

* Downloaded the already existing image
* Removed the legacy_boot flag
* booted the image and it worked.

- [x ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)

### DIFF
--- a/build_library/disk_layout.json
+++ b/build_library/disk_layout.json
@@ -16,7 +16,7 @@
         "blocks":"262144",
         "fs_type":"vfat",
         "mount":"/boot",
-        "features": ["hybrid"]
+        "features": []
       },
       "2":{
         "label":"BIOS-BOOT",
--- a/changelog/bugfixes/2025-02-14-fix-efi-partition-flags.md
+++ b/changelog/bugfixes/2025-02-14-fix-efi-partition-flags.md
@@ -1,0 +1,1 @@
+- Fix non-conforming GPT partition table ([flatcar/Flatcar#1651](https://github.com/flatcar/Flatcar/issues/1651)


chewi (approved):
It looks straightforward, but I'll kick off a Jenkins run against it.

gabriel-samfira:
Let me also test it out manually. Will ping back once I confirm if it works.

gabriel-samfira:
Converted to draft. While using parted 3.6 (version 3.4 seemed to make the image unbootable - need to investigate) on an existing image did seem to make it work, generating a new image without the hybrid flag seems to:

* Not set the bootable flag in the MBR
* makes the CHS entry invalid ([0, 1, 0] instead or [0, 2, 0])

chewi:
I'm surprised by your issues, but I can at least report that CI passed on all platforms except OpenStack and Equinix Metal. I believe those are actually down to availability issues. Brightbox worked, and that uses OpenStack too.

gabriel-samfira:
You need to run the tests on a fresh devstack deployment. The change that triggers issues in OpenStack was added in December.

The cgpt binary in https://github.com/flatcar/seismograph/tree/flatcar-master/src/cgpt is quite outdated. Newer versions of cgpt (like the one in ubuntu 24.04) can properly set CHS entries (along with other PMBR fields) by running:

```bash
cgpt boot -p path/to/raw.img
```

We then set the bootable flag in the PMBR either via fdisk or by simply flipping the bit at offset 447 in the disk to `0x80`. We then have a proper bootable hybrid image that passes the OpenStack  disk image tests.

gabriel-samfira:
to set the bootable flag via fdisk we can run:

```bash
echo -e "M\na\nw\n" | fdisk flatcar_production_openstack_image.raw
```

or we can do it via python in the `disk_util` module via something like:

```python
def EnsurePMBRBootFlag(options, config=None, partitions=None):
  # Set bootable flag
  with open(options.disk_image, 'r+b') as fd:
    # Seek to partition table offset
    fd.seek(446)
    # Read the partition entry
    partition = fd.read(16)
    partition_data = struct.unpack('<B3BB3BII', partition)
    if partition_data[0] != 0x80:
        new_partition_data = struct.pack('<B3BB3BII', 0x80, *partition_data[1:])
        # Go back to offset of start of partition
        fd.seek(446)
        # write the modified partition data
        fd.write(new_partition_data)
```

gabriel-samfira:
> to set the bootable flag via fdisk we can run:
> 
> ```shell
> echo -e "M\na\nw\n" | fdisk flatcar_production_openstack_image.raw
> ```
> 
> or we can do it via python in the `disk_util` module via something like:
> 
> ```python
> def EnsurePMBRBootFlag(options, config=None, partitions=None):
>   # Set bootable flag
>   with open(options.disk_image, 'r+b') as fd:
>     # Seek to partition table offset
>     fd.seek(446)
>     # Read the partition entry
>     partition = fd.read(16)
>     partition_data = struct.unpack('<B3BB3BII', partition)
>     if partition_data[0] != 0x80:
>         new_partition_data = struct.pack('<B3BB3BII', 0x80, *partition_data[1:])
>         # Go back to offset of start of partition
>         fd.seek(446)
>         # write the modified partition data
>         fd.write(new_partition_data)
> ```

the boot flag is not needed. To pass the OpenStack Validation, the CHS values are enough.

gabriel-samfira:
seems to pass on OpenStack. @chewi could you kick off another test in the CI as well?

chewi:
[Everything passed except for EM](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5448/cldsv/), as expected.

PR #2696, opened 2025-02-26
overlay wa-linux-agent: apply patch to fix ssh public key override issue (stable, beta)

dongsupark:
Apply patch to fix an issue when overriding ssh public key from ignition configuration. Since the fix is not available in releases of wa-linux-agent, we should apply a separate patch.

See also https://github.com/Azure/WALinuxAgent/pull/3309.

See also https://github.com/flatcar/Flatcar/issues/1661.

## Testing done

CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5465/cldsv/ passed

### DIFF
--- a/changelog/bugfixes/2025-02-28-walinuxagent-ignition.md
+++ b/changelog/bugfixes/2025-02-28-walinuxagent-ignition.md
@@ -1,0 +1,1 @@
+- Fix issue of wa-linux-agent overriding ssh public key from ignition configuration during provisioning ([flatcar/Flatcar#1661](https://github.com/flatcar/Flatcar/issues/1661))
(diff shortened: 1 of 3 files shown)


t-lo:
Should this go to LTS 2025, too?

dongsupark:
I think it does not have to be in LTS.
As described in the issue, https://github.com/flatcar/Flatcar/issues/1661, it was not a critical bug, because that override happens only under limited circumstances. It was also not a serious security issue.

dongsupark:
Cherry-picked to flatcar-4230 (current Alpha, future Beta)

PR #2706, opened 2025-03-01
Monthly GLSA metadata 2025-03-01

flatcar-infra:
Updated GLSA metadata


krnowak (approved):
This is only about openssh stuff we already have in main. But I see that the gh actions are already running, so let's wait for them to finish before merging this.

PR #2723, opened 2025-03-07
coreos-modules: move mana kmod to commonconfig

t-lo:
This change moves `CONFIG_MICROSOFT_MANA=m` from `amd64_defconfig-6.6` to `commonconfig-6.6` to support the MANA network driver on ARM64 instances, too.

## How to use

Build the OS image, launch a qemu aarch64 instance, check if the module loads.

## Testing done

See "How to use".

## Before merge

Remove commit https://github.com/flatcar/scripts/pull/2723/commits/98ffef548d44d5b6bea21ce1594495cd4dcafa24 ; this was added because the last nightly SDK build did not complete, and today's SDK is not available - so we need to use yesterday's SDK.

## Backport / Cherry-Pick

This change should be backported to all active maintenance branches, i.e. flatcar-4230, flatcar-4186, flatcar-4152, and flatcar-4081.


chewi:
How much does this increase the initrd size by, if anything? arm64 is actually closer to the limit than amd64.

t-lo:
> How much does this increase the initrd size by, if anything? arm64 is actually closer to the limit than amd64.

The kmod is about ~40kB.

chewi (approved):

t-lo:
Build and test succeeded, reverting https://github.com/flatcar/scripts/pull/2723/commits/98ffef548d44d5b6bea21ce1594495cd4dcafa24 and force-merging.

t-lo:
Done.

PR #2751, opened 2025-03-17
app-crypt/clevis: Version bump to 21

chewi:
# app-crypt/clevis: Version bump to 21

This moves the systemd unit enabling to the image build scripts to make the ebuild less Flatcar-specific.

Unfortunately, Clevis is still very automagic, resulting in a poor quality ebuild. Improving this was actually the very first thing I tried to do for Flatcar back in 2022, 1½ years before I joined the team. I will try to revive this effort soon, and then we can maybe get the package upstreamed to Gentoo.

This also drops dev-libs/libpwquality and sys-libs/cracklib, which are no longer needed. They seemingly weren't explicitly included in Flatcar.

## How to use

Admittedly, I don't know much about Clevis, so I'm trusting CI.

## Testing done

A [QEMU-only Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5573/cldsv/) passed successfully. We appear to have some Clevis-specific tests.

Size-wise, unfortunately this adds awk and socat to the initrd, which inflates it by 1,322KB uncompressed. These are needed by the new pkcs11 Clevis module. We could prevent that from being installed, but the automagic nature of Clevis currently makes this awkward. Users may also find this feature useful. I looked into replacing its awk usage with sed, but I realised that the later releases of Dracut itself also install awk. The good news is that we're still within the limit, and for some unexplained reason, the arm64 kernel actually shrank considerably between 4186 and 4230. The size reported by this new build is still smaller than it previously was since 3815. I think we'll be okay between now and when I ultimately resolve these size limitations.


krnowak:
> On second thoughts, I want to tame Clevis a bit first. The pkcs11 module should actually depend on pcsc-lite. We have this in our repo for the SDK, but we don't actually install it in the image. If it is installed, it bloats the initrd too much.

Yeah, I think pcsc-lite was added to SDK purely for release engineering stuff.

PR #2752, opened 2025-03-17
add -nographic flag for x86_64 arch

thebigbone:
# Add -nographic in `build_library/qemu_template.sh`

Refer https://github.com/flatcar/Flatcar/issues/1687

## Testing done

- Use the flag to boot up flatcar linux image on a headless x86_64 system and it works fine.

### DIFF
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -228,7 +228,7 @@ else
                    -global driver=cfi.pflash01,property=secure,value=on \
                    "$@"
             # Emulate the host CPU closely in both features and cores.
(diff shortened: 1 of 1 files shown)


chewi:
I was actually looking to go the other way and remove `-nographic` from all cases, as having the display can help diagnose some early boot issues. There were some driver-related issues with arm64 though, so I didn't push it yet. You can always add `-- -nographic` yourself if that's what you want.

chewi:
Maybe we could make this automatic by checking the `DISPLAY` and `WAYLAND_DISPLAY` variables.

PR #2753, opened 2025-03-17
overlay sys-apps/baselayout: Don't create empty /etc configs

danzatt:
There is a bug in coreos baselayout which wipes
/etc/{group,gshadow,passwd,shadow} when
reinstalling/removing/upgrading the baselayout package.

The deleted files are touched in staging area, so the package ships empty configuration files, overwriting the original configs on the system. Instead we move the touch to postinst, which only touches the existing files (or creates them when they're nonexistent).

This is blocking the NVIDIA sysext work, as the nvidia-drivers ebuild assumes the video group exists in `/etc/group`.

Note: as upgrading baselayout package breaks the SDK container due to this bug, we likely also need to rebuild the SDK

## How to use

## Testing done


chewi (approved):
I'm not sure how this broke in the SDK, but I think I can guess. If building a new SDK works, that should be fine.

PR #2765, opened 2025-03-24
coreos-base/update-ssh-keys: Bump for improved docs with better Clap use

chewi:
# Bump update-ssh-keys for improved docs with better Clap use

See flatcar/update-ssh-keys#18.

## Testing done

I've built and run the new binary using the new ebuild in the SDK.


sdk_container/src/third_party/coreos-overlay/coreos-base/update-ssh-keys/update-ssh-keys-9999.ebuild
```diff
@@ -61,15 +71,13 @@ else
 		windows_x86_64_gnullvm@0.48.5
 		windows_x86_64_msvc@0.48.5
 	"
-
-	SRC_URI="https://mirror.release.flatcar-linux.net/coreos/openssh-keys-0.5.1-alpha.0.crate"
```
krnowak, line 65: I guess it's not needed anymore, but what was that?

chewi: I'm actually not sure. A newer openssh-keys is listed in the crates above. `cargo tree` showed this wasn't being used.

krnowak: Ah, seems to be a remnant from https://github.com/flatcar/update-ssh-keys/pull/7/files and https://github.com/flatcar-archive/coreos-overlay/pull/2289. Anyway, good riddance.

krnowak (approved):

PR #2300, opened 2024-09-10
Upgrade Linux kernel from 6.6 to 6.12

ader1990:
# Upgrade Linux kernel from the 6.6.y stable branch to 6.12.y stable branch.

See: https://github.com/flatcar/Flatcar/issues/1527

This PR is mostly to reveal any possible big blockers before getting to the new 6.12 LTS release.

Tested 6.10.y and it works as expected.
Tested 6.11.y and it works as expected.

Now testing 6.12.y.

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

Boot partition size:

```
arm64: /dev/nvme0n1p1     129039    63368     65672  50% /boot
amd64: /dev/vda1          129039    62852     66187  49% /boot
```

PR #2766, opened 2025-03-24
cpio: do not fail on ghost rootfs

ader1990:
On Linux >= 6.10, the first rootfs is an extra ghost rootfs of 336K, that has a corrupted CPIO.

To overcome this issue, do not fail on `cpio --extract`.

### DIFF
--- a/build_library/extract-initramfs-from-vmlinuz.sh
+++ b/build_library/extract-initramfs-from-vmlinuz.sh
@@ -81,11 +81,9 @@ perform_round() {
     for rnd in "${round_prefix}"*; do
         if [[ $(file --brief "${rnd}") =~ 'cpio archive' ]]; then
             mkdir -p "${out}/rootfs-${ROOTFS_IDX}"
-            while cpio --quiet --extract --make-directories --directory="${out}/rootfs-${ROOTFS_IDX}" --nonmatching 'dev/*'; do
-                ROOTFS_IDX=$(( ROOTFS_IDX + 1 ))
-                mkdir -p "${out}/rootfs-${ROOTFS_IDX}"
-            done <${rnd}
-            rmdir "${out}/rootfs-${ROOTFS_IDX}"
+            # On Linux 6.10, the first rootfs is an extra ghost rootfs of 336K, that has a corrupted CPIO
+            cpio --quiet --extract --make-directories --directory="${out}/rootfs-${ROOTFS_IDX}" --nonmatching 'dev/*' < $rnd || true
+            ROOTFS_IDX=$(( ROOTFS_IDX + 1 ))
         fi
     done
 }


ader1990:
Required by: https://github.com/flatcar/scripts/pull/2300

build_library/extract-initramfs-from-vmlinuz.sh
```diff
@@ -81,11 +81,9 @@ perform_round() {
     for rnd in "${round_prefix}"*; do
         if [[ $(file --brief "${rnd}") =~ 'cpio archive' ]]; then
             mkdir -p "${out}/rootfs-${ROOTFS_IDX}"
-            while cpio --quiet --extract --make-directories --directory="${out}/rootfs-${ROOTFS_IDX}" --nonmatching 'dev/*'; do
-                ROOTFS_IDX=$(( ROOTFS_IDX + 1 ))
-                mkdir -p "${out}/rootfs-${ROOTFS_IDX}"
-            done <${rnd}
-            rmdir "${out}/rootfs-${ROOTFS_IDX}"
+	    # On Linux 6.10, the first rootfs is an extra ghost rootfs of 336K, that has a corrupted CPIO
```
chewi, line 84: Your whitespace is off.

chewi (changes requested):
I'm struggling to make sense of the original code. Why was it in a loop? Wouldn't it have extracted the same thing over and over?

chewi:
Ah, now I get it. I was thinking it would reread `$rnd` each time, but it's only read once. I have heard of archives being concatenated like this, so this change does concern me. I'm trying to think how it could loop as before but only break when reaching the end of `$rnd`. Nothing is springing to mind.

ader1990:
The failure happens only on ARM64, awith linux kernel >= 6.10. it would be rather hard to bisect the issue to a kernel commit, maybe we can ask on lkml what actually happened?

krnowak:
So I did the following steps (`../../scripts/main` is where I have a checkout of the main branch of the scripts repo):

1. `wget https://bincache.flatcar-linux.net/images/arm64/9999.0.0+kernel-6.12-rev4/flatcar_production_qemu_uefi_image.img`
2. `qemu-img convert -f qcow2 -O raw flatcar_production_qemu_uefi_image.img flatcar_production_qemu_uefi_image.raw`
3. `../../scripts/main/build_library/disk_util mount flatcar_production_qemu_uefi_image.raw rootfs`
4. `cp -a rootfs/boot/flatcar/vmlinuz-a .`
5. `../../scripts/main/build_library/disk_util umount rootfs`
6. `../../scripts/main/build_library/extract-initramfs-from-vmlinuz.sh vmlinuz-a initrd/`

(These steps are what more-or-less happens in the `finish_image` function in `build_library/build_image_util.sh`.)

The output from the last point was:

```
cpio: premature end of archive
done, found 1 rootfs(es)
```

The `cpio: premature end of archive` warning was always printed as far as I can remember. The `initrd` directory contents were:

```
$ ls -la initrd/
total 12
drwxrwxr-x  3 krnowak krnowak 4096 Mär 26 10:37 .
drwxrwxr-x  4 krnowak krnowak 4096 Mär 26 10:41 ..
drwxrwxr-x 12 krnowak krnowak 4096 Mär 26 10:37 rootfs-0

$ ls -la initrd/rootfs-0/
total 56
drwxrwxr-x 12 krnowak krnowak 4096 Mär 26 10:37 .
drwxrwxr-x  3 krnowak krnowak 4096 Mär 26 10:37 ..
lrwxrwxrwx  1 krnowak krnowak    7 Mär 26 10:37 bin -> usr/bin
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 dev
drwxr-xr-x 10 krnowak krnowak 4096 Mär 26 10:37 etc
lrwxrwxrwx  1 krnowak krnowak   23 Mär 26 10:37 init -> usr/lib/systemd/systemd
lrwxrwxrwx  1 krnowak krnowak    7 Mär 26 10:37 lib -> usr/lib
lrwxrwxrwx  1 krnowak krnowak    9 Mär 26 10:37 lib64 -> usr/lib64
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 proc
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 root
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 run
lrwxrwxrwx  1 krnowak krnowak    7 Mär 26 10:37 sbin -> usr/bin
-rwxr-xr-x  1 krnowak krnowak 4396 Mär 26 10:37 shutdown
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 sys
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 sysroot
drwxr-xr-x  2 krnowak krnowak 4096 Mär 26 10:37 tmp
drwxr-xr-x  7 krnowak krnowak 4096 Mär 26 10:37 usr
drwxr-xr-x  4 krnowak krnowak 4096 Mär 26 10:37 var
```

So, honestly I don't know what's in here to fix. Maybe the issue went away at some point?

chewi:
Now you mention it, I remember seeing that error too.

krnowak:
But what is certainly interesting that I haven't noticed before is the disappearance of the early cpio between 3975.2.2 and 4054.0.0. Normally extracting stuff from vmlinuz-a for the amd64 qemu image resulted in 2 rootfses found, now it's just one. The missing one, the early-cpio, used to contain CPU firmware. Maybe something has changed in the kernel, so that the early-cpio is no more or our extract script has no clue how to get it.

Compare [initrd contents for 3975.2.2](https://bincache.flatcar-linux.net/images/amd64/3975.2.2/flatcar_production_image_initrd_contents.txt) and [initrd contents for 4054.0.0](https://bincache.flatcar-linux.net/images/amd64/4054.0.0/flatcar_production_image_initrd_contents.txt), right in the beginning.

This possibly looks like a change between kernel 6.6.54 and 6.6.60.

But anyway, it's a side observation.

chewi:
I think this might be because of [this](https://github.com/flatcar/scripts/commit/809ef73438b5fb8515a8587ac60558ea38f9e6eb#diff-c6918e3926cabe521a551503d0a52daed4a7849d1f0fbbeb7f699ace5cd71076R15-R20).

chewi:
I think we can merge this (but fix the whitespace :wink:). CPU microcode used to be installed with the kernel because everyone should take it, but not everyone uses an initrd. Under this new model, the scripts that install the kernel know whether you're using Dracut or not, so they can be smarter about it. I don't expect anything else to create an additional cpio while using Dracut.

krnowak:
> I think it might be because of [this](https://github.com/flatcar/scripts/commit/809ef73438b5fb8515a8587ac60558ea38f9e6eb#diff-c6918e3926cabe521a551503d0a52daed4a7849d1f0fbbeb7f699ace5cd71076R15-R20).

I can't wrap my head around this stuff at the moment - so are firmware updates applied or is there some regression now?

chewi:
Hmmm, actually it seems to be a regression. I was assuming that Dracut was including this firmware and that it was previously duplicated, but I can't see any sign of it in the current initrds. This is probably due to the way we call Dracut. I'm working in this area right now, so I'll take a look. It sounds bad, but I guess it's been like this for a while, and no one has shouted yet.

krnowak:
> Hmmm, actually it seems to be a regression. I was assuming that Dracut was including this firmware and that it was previously duplicated, but I can't see any sign of it in the current initrds. This is probably due to the way we call Dracut. I'm working in this area right now, so I'll take a look. It sounds bad, but I guess it's been like this for a while, and no one has shouted yet.

Thanks! I must have missed that during the weekly updates… It would be something that would show image in image changes.

chewi:
That's a gap in the image changes script, I guess. It only checks the "main" initrd.

krnowak:
> > I think it might be because of [this](https://github.com/flatcar/scripts/commit/809ef73438b5fb8515a8587ac60558ea38f9e6eb#diff-c6918e3926cabe521a551503d0a52daed4a7849d1f0fbbeb7f699ace5cd71076R15-R20).
> 
> As @krnowak said, maybe the issue has been resolved in the meantime, although it might still linger in some versions of the Linux or the firmware code, so I would like to merge this PR or maybe keep it in our collective `memory` if this issue happens again.
> 
> I will try running the 6.12.20 build without this change and see what happens.
> 
> What do you think of merging it nonetheless? - this PR basically loosens the conditions.

With my bash hat on - unless I missed something, I can't see any loosening of conditions. Previously the loop was executing until cpio failed (so in practice - just once, the second run of cpio failed) - now it will run just once. So the difference will be that the warning about premature end of archive won't be printed.

I mean, I don't mind in the end. This introduces no regression, makes the code simpler (`while cpio …; do …; done <"${path}"` looks confusing, as it is not actually rereading the file on every iteration).

chewi:
Yeah, even if I fix the missing firmware, we won't go back to installing it the way we did before. It'll be Dracut that does it.

krnowak:
> That's a gap in the image changes script, I guess. It only checks the "main" initrd.

Image changes script just consumes the generated initrd contents files that are stored in bincache. Those files are generated in build_image_util.sh and they should contain listings of all the rootfses that the extract script manages to extract from vmlinuz.

krnowak:
> Yeah, even if I fix the missing firmware, we won't go back to installing it the way we did before. It'll be Dracut that does it.

Did the old way become deprecated?

chewi:
It's not deprecated, but it's intended for people not using Dracut. It would have been duplicating the files if we'd been using Dracut the normal way.

ader1990:
@chewi the code was rebased and the small indent was fixed. I cannot run some github actions on it, as it seems to be down still.

chewi:
I'll give it a run in Jenkins then.

chewi:
Running [here](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5659/cldsv/). The build will be versioned as **main-9999.0.0-initrd-cpio-1**.

chewi:
The previous nightly failed, so I've rebased on the one before.

chewi:
The [Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5663/cldsv/) passed. There were practically no changes. I'll merge this now.

ader1990:
Thanks @chewi, will remove it from https://github.com/flatcar/scripts/pull/2300 and rebase.

PR #2798, opened 2025-03-31
Add nvidia-drivers sysext

danzatt:
# Add prebuilt NVIDIA drivers in a sysext

- Add capability to specify per-sysext USE flags and compile different versions of upstream portage nvidia-drivers (including open and non-open variants).
- Allow architecture-specific OS-dependent sysexts
- Pull `nvidia-drivers` from portage and build sysexts from the package

Related PRs:
NVIDIA tests using sysext: [mantle #598](https://github.com/flatcar/mantle/pull/598)
NVIDIA runtime modifications to remove `nvidia-smi` symlink: [sysext-bakery #153](https://github.com/flatcar/sysext-bakery/pull/153)
## Testing done

Created a build in Jenkins and tested using kola on Azure instances with NVIDIA GPUs

- [*] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)


build_library/extra_sysexts.sh
```diff
@@ -2,4 +2,24 @@ EXTRA_SYSEXTS=(
...
   python:dev-lang/python,dev-python/pip
+  nvidia-drivers-535:=x11-drivers/nvidia-drivers-535.230.02:-kernel-open
+  nvidia-drivers-535-open:=x11-drivers/nvidia-drivers-535.230.02:kernel-open
+  nvidia-drivers-550:=x11-drivers/nvidia-drivers-550.144.03:-kernel-open
+  nvidia-drivers-550-open:=x11-drivers/nvidia-drivers-550.144.03:kernel-open
 )
+
+_get_unversioned_sysext_packages_unsorted() {
+  for sysext in "${EXTRA_SYSEXTS[@]}"; do
+    IFS=":" read SYSEXT_NAME PACKAGE_ATOMS USEFLAGS < <(echo "$sysext");
```
chewi, line 13: Always include `-r` when using `read`.
```suggestion
    IFS=":" read -r SYSEXT_NAME PACKAGE_ATOMS USEFLAGS < <(echo "$sysext");
```

build_library/extra_sysexts.sh
```diff
@@ -2,4 +2,24 @@ EXTRA_SYSEXTS=(
...
+
+    IFS=, read -ra atoms <<< "$PACKAGE_ATOMS"
+    for atom in "${atoms[@]}"
+    do
+       qatom $atom -F "%{CATEGORY}/%{PN}"
+    done
```
chewi, line 19: You don't need to convert to an array first.
```suggestion
    IFS=,
    for atom in $PACKAGE_ATOMS
    do
       qatom $atom -F "%{CATEGORY}/%{PN}"
    done
    unset IFS
```

build_library/extra_sysexts.sh
```diff
@@ -2,4 +2,24 @@ EXTRA_SYSEXTS=(
   zfs:sys-fs/zfs
   podman:app-containers/podman,net-misc/passt
   python:dev-lang/python,dev-python/pip
+  nvidia-drivers-535:=x11-drivers/nvidia-drivers-535.230.02:-kernel-open
+  nvidia-drivers-535-open:=x11-drivers/nvidia-drivers-535.230.02:kernel-open
+  nvidia-drivers-550:=x11-drivers/nvidia-drivers-550.144.03:-kernel-open
+  nvidia-drivers-550-open:=x11-drivers/nvidia-drivers-550.144.03:kernel-open
```
chewi, line 8: We could possibly make maintenance easier if we refer to these by SLOT instead, e.g. `nvidia-drivers-535|x11-drivers/nvidia-drivers:0/535|-kernel-open`. The `emerge` command will understand that, although I've had to change the separator to `|`.

build_packages
```diff
@@ -292,6 +293,25 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
...
+info "Merging sysext packages now"
+for sysext in "${EXTRA_SYSEXTS[@]}"; do
+  IFS=":" read SYSEXT_NAME PACKAGE_ATOMS USEFLAGS < <(echo "$sysext");
+
+  info "Building packages for $SYSEXT_NAME sysext with USE=$USEFLAGS"
+  IFS=, read -ra atoms <<< "$PACKAGE_ATOMS"
+  for package in "${atoms[@]}"
+  do
+     USE="$USEFLAGS" FEATURES="-ebuild-locks binpkg-multi-instance" \
+       sudo -E "${EMERGE_CMD[@]}" \
```
chewi, line 305: If you don't need `MODULES_SIGN_*` here...
```suggestion
     sudo env USE="$USEFLAGS" FEATURES="-ebuild-locks binpkg-multi-instance" \
       "${EMERGE_CMD[@]}" \
```

build_packages
```diff
@@ -326,7 +346,16 @@ if [[ "${FLAGS_usepkgonly}" -eq "${FLAGS_FALSE}" ]]; then
...
 fi
 
-eclean-$BOARD -d packages
+exclusions_file=$(mktemp)
+if [ ! -f "$exclusions_file" ]; then
+    die_notrace "Couldn't create temporary exclusions file $exclusions_file for eclean"
+fi
+get_unversioned_sysext_packages > $exclusions_file
+eclean-$BOARD -d --exclude-file=$exclusions_file packages
+rm -f "$exclusions_file"
```
chewi, line 355: This trick doesn't always work, but it seems to here:
```suggestion
eclean-$BOARD -d --exclude-file=<(get_unversioned_sysext_packages) packages
```

chewi (changes requested):

build_library/extra_sysexts.sh
```diff
@@ -1,5 +1,26 @@
...
+  "nvidia-drivers-550-open|x11-drivers/nvidia-drivers:0/550|kernel-open"
 )
+
+_get_unversioned_sysext_packages_unsorted() {
+  for sysext in "${EXTRA_SYSEXTS[@]}"; do
+    IFS="|" read -r SYSEXT_NAME PACKAGE_ATOMS USEFLAGS < <(echo "$sysext");
+
+    IFS=,
+    for atom in $PACKAGE_ATOMS
+    do
```
jepio, line 17: For consistency with other usages:
```suggestion
    for atom in $PACKAGE_ATOMS; do
```

sdk_container/src/third_party/coreos-overlay/eclass/coreos-kernel.eclass
```diff
@@ -136,20 +136,43 @@ getconfig() {
...
 
+get_sig_key() {
+	local sig_key="$(getconfig MODULE_SIG_KEY)"
+
+	if [[ "${sig_key}" == "build/certs/signing_key.pem" ]]; then
+		die "MODULE_SIG_KEY is using the default value"
+	fi
+
+	if [[ ${sig_key} != /tmp/* ]]
+	then
```
jepio, line 147:
```suggestion
	if [[ ${sig_key} != /tmp/* ]]; then
```

sdk_container/src/third_party/coreos-overlay/eclass/coreos-kernel.eclass
```diff
@@ -136,20 +136,43 @@ getconfig() {
...
+		die "Refusing to to continue with modules key outside of /tmp, so that it stays in RAM only."
+	fi
+	if [ "$sig_key" != "${MODULES_SIGN_KEY}" ]
+	then
```
jepio, line 151:
```suggestion
	if [ "$sig_key" != "${MODULES_SIGN_KEY}" ]; then
```

sdk_container/src/third_party/coreos-overlay/eclass/coreos-kernel.eclass
```diff
@@ -169,19 +192,20 @@ setup_keys() {
...
 
-# Discard the module signing key but keep public certificate.
-shred_keys() {
-	local sig_key
-	sig_key="build/$(getconfig MODULE_SIG_KEY)"
-	shred -u certs/modules.key.pem "${sig_key}" || die
-	cp certs/modules.pub.pem "${sig_key}" || die
+	# copy the cert/key to desired location
+	mkdir -p "${MODULES_SIGN_CERT%/*}" "${MODULES_SIGN_KEY%/*}" || die
+	cat gen_certs/modules.pub.pem gen_certs/modules.key.pem > "$MODULES_SIGN_KEY" || die
```
jepio, line 202: why is the public key prepended to the private key?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```diff
@@ -83,3 +83,7 @@
 =sys-libs/libsemanage-3.7 ~arm64
 =sys-libs/ncurses-6.5_p20250125 ~arm64
 =sys-process/audit-4.0.2-r1 ~arm64
+
+# enable NVIDIA drivers for ARM
+=x11-drivers/nvidia-drivers-550.144.03 ~arm64
+=x11-drivers/nvidia-drivers-535.230.02 ~arm64
```
jepio, line 89: So slight complication here... I believe only 570 builds for arm64 with the kernel+gcc combination we're using - at least that's what I found when adding support for it to nvidia drivers recently.

How about disabling nvidia sysexts for arm64 for the time being, to focus on getting amd64 in shape.

ci-automation/image.sh
```diff
@@ -88,7 +88,7 @@ function _image_build_impl() {
 
     source ci-automation/base_sysexts.sh 'local'
     local base_sysexts_param
-    base_sysexts_param=$(export IFS=,; echo "${ciabs_base_sysexts[*]}")
+    base_sysexts_param=$(export IFS=|; echo "${ciabs_base_sysexts[*]}")
```
jepio, line 91:
```
12:42:03  jenkins/systemd-run-wrap.sh: ci-automation/image.sh: line 91: syntax error near unexpected token `;'
12:42:03  jenkins/systemd-run-wrap.sh: ci-automation/image.sh: line 91: `    base_sysexts_param=$(export IFS=|; echo "${ciabs_base_sysexts[*]}")'
```

'|' needs quoting here.

build_packages
danzatt: We actually need the `MODULES_SIGN_*` here, as we are building it for the first time. We don't need it when installing for a second time from the binpkg.

chewi: Passing through the whole environment is not ideal, but I just found out you can do `sudo --preserve-env=MODULES_SIGN_KEY,MODULES_SIGN_CERT`, which is slightly more convenient than `sudo env MODULES_SIGN_KEY="${MODULES_SIGN_KEY}" MODULES_SIGN_CERT="${MODULES_SIGN_CERT}"`.

sdk_container/src/third_party/coreos-overlay/eclass/coreos-kernel.eclass
danzatt: I've just taken the original logic from the existing eclass: https://github.com/flatcar/scripts/blob/32416aabbcb756c6bdd2231dd2e39dac7097978b/sdk_container/src/third_party/coreos-overlay/eclass/coreos-kernel.eclass#L176
Gentoo also seems to do the same thing:
https://github.com/gentoo/gentoo/blob/b64e85fb80f47f5a12ca90947069eb433d212540/eclass/kernel-build.eclass#L166

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
danzatt: Ah, right. I was wondering why the ARM builds always fail with compilation errors. I'll disable it for now, which reminds me there currently isn't a way to make a sysext for AMD only. I might need to add another field to the sysext specifier..

danzatt:
Should be ready now, running a build in Jenkins

build_packages
```diff
@@ -285,6 +286,48 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
 info "Merging board packages now"
 sudo -E "${EMERGE_CMD[@]}" "${EMERGE_FLAGS[@]}" "$@"
 
+info "Merging sysext packages now"
+for sysext in "${EXTRA_SYSEXTS[@]}"; do
+  IFS="|" read SYSEXT_NAME PACKAGE_ATOMS USEFLAGS ARCHES < <(echo "$sysext");
```
chewi, line 291: Always use `-r` with `read` and make use of here strings.

```suggestion
  IFS="|" read -r SYSEXT_NAME PACKAGE_ATOMS USEFLAGS ARCHES <<< "$sysext"
```

build_packages
```diff
@@ -285,6 +286,48 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
...
+
+  arch_array=(${ARCHES//,/ })
+  if [[ ! -z "$ARCHES" ]]; then
```
chewi, line 294: No one likes double negatives. :wink: Also, you don't need to quote the LHS in `[[`.

```suggestion
  if [[ -n $ARCHES ]]; then
```

It can even be just `[[ $ARCHES ]]` although my personal preference is to be a little more explicit.

build_packages
```diff
@@ -285,6 +286,48 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
...
+    should_skip=1
+    for arch in "${arch_array[@]}"; do
+      if [ "$arch" == "$ARCH" ]; then
```
chewi, line 297: Please be more consistent with shell standards. If you know this is Bash, rather than POSIX shell (/bin/sh) then always use `[[` over `[` because it's safer and doesn't require quoting the LHS. I'm flagging this particular example because `==` is a Bash extension, so you would need to use `=` to support POSIX shells. shellcheck doesn't flag it but mixing `[` with `==` is confusing.

```suggestion
      if [[ $arch == "$ARCH" ]]; then
```

build_packages
```diff
@@ -285,6 +286,48 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
...
+        should_skip=0
+      fi
+    done
+    if [ "$should_skip" -eq 1 ]; then
```
chewi, line 301:
```suggestion
    if [[ $should_skip -eq 1 ]]; then
```

build_packages
```diff
@@ -285,6 +286,48 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
...
+
+
+  info "Building packages for $SYSEXT_NAME sysext with USE=$USEFLAGS"
+  IFS=,
+  for package in $PACKAGE_ATOMS; do
+     # --buildpkgonly does not install dependencies, so we install them
+     # separately before building the binary package
+     sudo --preserve-env=MODULES_SIGN_KEY,MODULES_SIGN_CERT \
+       env USE="$USEFLAGS" FEATURES="-ebuild-locks binpkg-multi-instance" "${EMERGE_CMD[@]}" \
+       ${EMERGE_FLAGS[@]} \
```
chewi, line 314: You need to quote this to use it as an array and avoid unintended word splitting.

```suggestion
       "${EMERGE_FLAGS[@]}" \
```

build_packages
```diff
@@ -285,6 +286,48 @@ export KBUILD_BUILD_HOST="${BUILD_HOST:-pony-truck.infra.kinvolk.io}"
...
+       --quiet \
+       --onlydeps \
+       --binpkg-respect-use=y \
+       "${package}"
+
+     sudo --preserve-env=MODULES_SIGN_KEY,MODULES_SIGN_CERT \
+       env USE="$USEFLAGS" FEATURES="-ebuild-locks binpkg-multi-instance" "${EMERGE_CMD[@]}" \
+       ${EMERGE_FLAGS[@]} \
```
chewi, line 322:
```suggestion
       "${EMERGE_FLAGS[@]}" \
```

chewi (changes requested):
Thanks, we're nearly there now. I haven't highlighted every instance of the things I've flagged, so please check for others.

danzatt:
@chewi thanks for the review! I've just run shellcheck on this PR and hopefully fixed all instances of the bug

t-lo:
Happens to me too occasionally. Time for the good old "make a diff to main and add a single patch on top of a new pristine branch forked from main"?

danzatt:
I just rebased on main and force-pushed, seems to be OK now.

chewi (approved):

ader1990:
Hello, I have just a comment on the naming -> [Rename nvidia-drivers to nvidia-drivers-service](https://github.com/flatcar/scripts/pull/2798/commits/ab90a853041d944c47855c6a8b6d6dc4ce7bb8ad) - when I started reading, I was expecting to see a systemd service implementation, where there was a systemd nvidia-drivers unit which was renamed to nvidia-drivers-service. Maybe a rename in something that does not contain the suffix `service`? nvidia-drivers was rather more suitable, any reason to rename it nvidia-drivers-service?

ader1990:
> Hello, I have just a comment on the naming -> [Rename nvidia-drivers to nvidia-drivers-service](https://github.com/flatcar/scripts/pull/2798/commits/ab90a853041d944c47855c6a8b6d6dc4ce7bb8ad) - when I started reading, I was expecting to see a systemd service implementation, where there was a systemd nvidia-drivers unit which was renamed to nvidia-drivers-service. Maybe a rename in something that does not contain the suffix `service`? nvidia-drivers was rather more suitable, any reason to rename it nvidia-drivers-service?

Okay, so I understand - the Flatcar package name was changed and not the underlying service name, which was left as is for backwards compatibility. It was a little bit confusing as I expected a service name change too. I suppose the naming can be left as is `nvidia-drivers-service` or `nvidia-drivers-systemd-unit` (to be more clear in intent).

danzatt:
For a reference, there are also related open PRs for [kola](https://github.com/flatcar/mantle/pull/598) and [sysext-bakery](https://github.com/flatcar/sysext-bakery/pull/153)

sayanchowdhury (approved):

PR #2799, opened 2025-04-01
Monthly GLSA metadata 2025-04-01

flatcar-infra:
Updated GLSA metadata


dongsupark (approved):
No actual change, should be ok

PR #2808, opened 2025-04-07
build_library: Fix path handling in QEMU launcher script

chewi:
# build_library: Fix path handling in QEMU launcher script

I couldn't take it anymore! The launcher script could not handle paths outside the script's own directory, and it was driving me crazy. Now only the default values are relative to the script's directory. Given paths are relative to the current directory and absolute paths work as you would expect.

## How to use

You don't need to grab any new images, just the .sh files from [bincache](https://bincache.flatcar-linux.net/images/amd64/9999.0.0+qemu-script-paths-2/).

## Testing done

I ran [CI](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages/9100/cldsv/) but without the tests because they don't use this script. I grabbed the new .sh files, compared them against the old ones, and tried them out with different paths.

### DIFF
--- a/build_library/qemu_template.sh
+++ b/build_library/qemu_template.sh
@@ -260,9 +260,9 @@ fi
 if [ -n "${VM_IMAGE}" ]; then
     case "${VM_BOARD}" in
         amd64-usr)
-            set -- -drive if=virtio,file="${SCRIPT_DIR}/${VM_IMAGE}" "$@" ;;
+            set -- -drive if=virtio,file="${VM_IMAGE}" "$@" ;;
         arm64-usr)
-            set -- -drive if=none,id=blk,file="${SCRIPT_DIR}/${VM_IMAGE}" \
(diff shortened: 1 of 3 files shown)


sayanchowdhury (approved):

krnowak (commented):
Looks like the `SCRIPT_DIR` variable is now unused, so maybe you could drop it.

chewi:
> Looks like the `SCRIPT_DIR` variable is now unused, so maybe you could drop it.

No, it is still used, just in a different way now.

krnowak:
> > Looks like the `SCRIPT_DIR` variable is now unused, so maybe you could drop it.
> 
> No, it is still used, just in a different way now.

Oh, ok. I just did a search for SCRIPT_DIR in the script and the only occurence was in the third line. But maybe the script is being too clever somewhere. Anyway, looks good.

krnowak (approved):

PR #2835, opened 2025-04-17
Various changes in preparation for upgrading Dracut

chewi:
# Various changes in preparation for upgrading Dracut

My changes to upgrade Dracut and use its improved `--sysroot` functionality are quite heavy, so I decided to split the changes into two parts. This first part is the miscellaneous changes that are needed but can be applied on their own.

* sys-auth/google-oslogin: Install soname symlinks and general tidy up
  The initrd was previously created in `pkg_setup` to avoid the sandbox. This will no longer be the case. `ldconfig` was creating missing symlinks, which now breaks the sandbox.

* setup_board: Add ldconfig wrapper for non-native arches
  We will no longer run the whole of Dracut with QEMU, but we do still need to run `ldconfig` with it. Portage will also use this wrapper.

* sys-kernel/coreos-kernel: Add missing dependencies
  This wasn't an issue in CI, but I hit issues when building coreos-kernel locally without the entire board root present.

* build_dev_binpkgs: Refactor the script with better Bash techniques
  This script was hurting my eyes when dealing with the issue below.

* build_dev_binpkgs: Only build packages available to the board profile
  Updating the SDK to systemd-257 (see below) caused this script to break.

* sys-apps/systemd: Bump to v257.5 for the SDK only
  We need this for dracut-install to have the new JSON support.

* sys-kernel/coreos-modules: Compress kernel with xz rather than zstd
  The new initrd is heavier for various reasons. I trimmed some aspects, but I chose to keep all the keymaps as only supporting US English is unfair to bare metal users.

* sys-kernel/coreos-modules: Compress initrd with xz rather than zstd
  See above.

* build_library: Change extraction script to xz and rewrite to be simpler
  As above and the existing script was hard to understand.

## How to use

Nothing directly user-facing has changed here. Just check that nothing has broken.

## Testing done

A [full CI run with all tests](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5806/cldsv/) did pass eventually after some flakiness. I have already successfully run these changes together with the rest of the changes in part 2.

With the new initrd extraction script, we don't get the phantom extra "rootfs" directory on amd64 anymore. It says it has reduced by 413KB, but that usage was never actually real. Part 2 will add the CPU microcode, which will result in an extra rootfs directory again, but this one will be real.


sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.6
krnowak, line 1: Why arm64 kernel can't be compressed?

build_library/extract-initramfs-from-vmlinuz.sh
```diff
@@ -65,37 +49,22 @@ if [[ ! -s "${image}" ]]; then
...
 
-tmp_dec="${tmp}/decompress"
-mkdir "${tmp_dec}"
-fr_prefix="${tmp}/first-round"
+# arm64 kernels are not compressed, so try decompressing once.
+# Other kernels are compressed, so also try decompressing twice.
+for OFF1 in $(find_xz_headers "${image}")
+do
+    decompress_at "${image}" "${OFF1}" > "${tmp}/initrd.cpio.xz"
+    try_extract "${tmp}/initrd.cpio.xz"
```
krnowak, line 61: Maybe call this `${tmp}/initrd.cpio.maybe_xz`?

chewi: Okay, I did consider that. It's not like extensions mean much in Linux anyway!

setup_board
```diff
@@ -111,6 +115,18 @@ exec ${BOARD_CHOST}-gdb -iex 'set sysroot ${BOARD_ROOT}' "\$@"
 EOF
   wrappers+=( "${wrapper}" )
 
+  # ldconfig cannot generate caches for non-native arches. Use QEMU and the
+  # native ldconfig to work around that.
+  wrapper="/usr/local/sbin/ldconfig-${BOARD_VARIANT}"
+  sudo_clobber "${wrapper}" <<EOF
+#!/bin/sh
+exec ${qemu} "${BOARD_ROOT}"/sbin/ldconfig -r "${BOARD_ROOT}" "\$@"
```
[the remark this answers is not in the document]
chewi: I hadn't noticed before, but our strict mode only sets `-e`. I'll fix this anyway.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.6
chewi: If you grep the kernel sources, you'll see that most arches have `HAVE_KERNEL_XZ` and friends but not arm64. [Documentation/admin-guide/efi-stub.rst](https://github.com/torvalds/linux/blob/v6.15-rc3/Documentation/admin-guide/efi-stub.rst) mentions it, although it doesn't say why.

krnowak (approved):

build_library/extract-initramfs-from-vmlinuz.sh
krnowak: Yeah, these work for me. Thanks.

PR #2837, opened 2025-04-22
Upgrade Dracut to 106, use its sysroot feature, fix CPU microcode

chewi:
# Upgrade Dracut, use its sysroot feature, fix CPU microcode

This must be reviewed in tandem with flatcar/bootengine#105.

Flatcar has lagged behind on Dracut 053 for quite a while. The latest version is currently 106, although it did jump from 060 to 100 when it changed hands.

As well as upgrading, this moves the sys-kernel/dracut package from coreos-overlay to portage-stable. Patches are now applied as user patches.

The patches included here are large because we are relying on [my new Dracut improvements](https://github.com/dracut-ng/dracut-ng/pull/1260), which allow it to parse the ELF `.note.dlopen` dependency metadata used by JSON and reliably determine dependencies across foreign architectures. These are not yet merged but upstream seem willing. They will not be merged for version 107, but the hope is to get them in for 108.

We now run Dracut with its `--sysroot` argument, removing the need to chroot and jump through hoops to avoid the Portage sandbox in the coreos-kernel ebuild. The sandbox still needs a little taming due to ldconfig calling chroot internally, but this is a lot saner than it was before.

For now, Dracut still needs to be installed in both the SDK and the board root. dracut-install is a native binary, but the modules are expected to be found within the sysroot. This may change later. bootengine has no native binaries, so it only needs to be installed in the board root.

The CPU microcode was accidentally dropped from Flatcar a while back (don't know when) because Gentoo started telling Dracut to not include it when `USE=-initramfs`. Flatcar disabled that flag because the microcode is installed to /boot. This is only done under the board root though, not in the final image, so there's no harm in enabling it. That Dracut setting also affected the AMD microcode, which is part of coreos-firmware, so this fixes that too.

## How to use

You will need the latest SDK due to changes in my earlier PR. You should be able to update dracut in the SDK, plus dracut and bootengine in the board root, before building coreos-kernel.

## Testing done

I have done a lot of manual testing and [CI testing](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5825/cldsv/), as well as a close inspection of the initrd changes. Everything is looking good.

The size of the initrd was a concern. Upgrading Dracut initially increased the size massively, but I took measures here and in bootengine to trim it back down. Changing the compression from zstd to xz in my other PR was needed to bring the size back to around what it was previously though.

Dropping a Flatcar-specific patch to drop most of the keymaps was one reason for the size increase, but I deliberately chose to keep them as only supporting US English is unfair to bare metal users.

Adding the CPU microcode should have increased the size yet further. The raw intel-mc.img file is 9MB. Compressing it manually with xz only brings it down to 8MB. However, it made a minimal difference to the final vmlinuz, and I haven't been able to figure out why. The microcode definitely wasn't there before, and it definitely is there now. I have extracted the vmlinuz to make sure. The amd64 nightly report says the raw initrd is 19MB larger, and yet the final vmlinuz is 132KB smaller. Perhaps it compresses very well alongside the rest? :man_shrugging: 


sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-kernel/dracut/001-dracut-post-106.patch
krnowak, line 1: I usually am adding a README.md in the same directory explaning what the patches are for and when they can be dropped (if they can be dropped).

chewi: Ah yep, forgot. Done.

krnowak (approved):

chewi:
The step to delete bootengine.cpio simply isn't needed any more. See flatcar/flatcar-website#420.

PR #2838, opened 2025-04-22
Drop confusing and obsolete binpkg-related code

chewi:
# Drop confusing and obsolete binpkg-related code

I got very confused while looking into how the dev container is built. This code didn't help. I think it can be dropped.

## How to use

N/A

## Testing done

I have run [amd64 and arm64 builds](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages/9267/cldsv/) with the qemu_* tests. They passed successfully.


t-lo:
Looks OK to me - these are likely remains from the days where we built toolchains separately (they are now built with the SDK container and also published after that step). We have a dev container test that will verify we didn't break anything.

PR #2848, opened 2025-04-24
build_library/qemu_template.sh: Make it use bash

krnowak:
Insisting on just sh is pointless. Use bash, so the checkbashisms script running somewhere will just shut up.


chewi (changes requested):
CI will still fail because vm_image_util.sh calls `checkbashisms`.

sayanchowdhury (commented):
there is a explicit check for checkbashisms in `build_library/vm_image_util.sh` -- this would still fail.

PR #2849, opened 2025-04-24
Increase selinux coverage of the host system

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/sdk/2052/cldsv/

- switch to selinux profiles
- add more sec-policy packages
- do some cleanups in profiles wrt selinux, audit, python, perl and caps USE flags

TODO:

- mask python files from sys-libs/libselinux for generic images
- drop systemd patch that removes selinux checks


pothos:
When ready we should also make kola enforce early and not at runtime: https://github.com/flatcar/mantle/pull/487

PR #2857, opened 2025-04-29
pkg-auto: Add parsing of md5-metadata to aid report generation

krnowak:
I think it is best to review it commit by commit.

The largest items in this PR are:
- reworked report generation, so it runs twice instead of four times (so once for old state and once for new state now), added explicit call to egencache and the results are also copied to reports
- added code for computing diffs between two cache files (to figure out changes in dependencies, EAPIs, keywords, IUSE and eclasses
  - the eclasses part is not used anywhere yet, but I'd like to use it as a base for detecting which eclasses are used and which are not

I have used this code for couple weeks now for weekly package updates and it seems to be in a good enough shape to get merged.


pkg_auto/impl/inside_sdk_container_lib.sh
```diff
@@ -366,21 +305,29 @@ function package_sources_sdk() {
...
+        name=${arch^^}_BOARD_EO_W
+        files+=( "${!name}" )
+    done
+
+    local file
+    for file in "${files[@]}"; do
+        if cat "${file}" | grep --quiet --fixed-strings 'ERROR'; then
             fail "there are errors in emerge output warnings files"
         fi
     done
```
chewi, line 333:
```suggestion
    if grep --quiet --fixed-strings 'ERROR' "${files[@]}"; then
        fail "there are errors in emerge output warnings files"
    fi
```

pkg_auto/impl/inside_sdk_container_lib.sh
```diff
@@ -308,9 +242,12 @@ function packages_for_sdk() {
...
 # @ - parameters passed to sed
 function packages_for_board() {
-    # args are passed to sed_eo_and_sort
+    local arch=${1}; shift
+    # rest goes to sed
 
-    sed_eo_and_sort board "${@}"
+    local name=${arch^^}_BOARD_EO_F
+
+    cat "${!name}" | sed "${@}" | sort
```
chewi, line 250:
```suggestion
    sed "${@}" "${!name}" | sort
```

pkg_auto/impl/md5_cache_lib.sh
```diff
@@ -0,0 +1,1177 @@
...
+# PCF_KEYWORDS_IDX - a name of an array containing keyword objects
+# PCF_IUSE_IDX     - a name of an array containing iuse objects
+# PCF_BDEPEND_IDX  - a name of a group object with build dependencies
+# PCF_DEPEND_IDX   - a name of a group object with dependencies
+# PCF_IDEPEND_IDX  - a name of a group object with install dependencies
+# PCF_PDEPEND_IDX  - a name of a group object with post dependencies
+# PCF_RDEPEND_IDX  - a name of a group object with runtime dependencies
+# PCF_LICENSE_IDX  - a name of a group object with licenses
+# PCF_ECLASSES_IDX - a name of an array with used eclasses
+declare -gri PCF_EAPI_IDX=0 PCF_KEYWORDS_IDX=1 PCF_IUSE_IDX=2 PCF_BDEPEND_IDX=3 PCF_DEPEND_IDX=4 PCF_IDEPEND_IDX=5 PCF_PDEPEND_IDX=6 PCF_RDEPEND_IDX=7 PCF_LICENSE_IDX=8 PCF_ECLASSES_IDX=9
```
chewi, line 33: How sure are you that this order is fixed?

pkg_auto/impl/pkg_auto_lib.sh
```diff
@@ -2255,26 +2160,37 @@ function handle_pkg_update() {
     local pkg_name
     pkg_name=${new_pkg#*/}
     local -a lines
-    lines=( "from ${old} to ${new}")
+    lines=( "0:from ${old} to ${new}")
```
chewi, line 2163: It's not clear to me why `0:` was added to all these lines. Would some example output make it clearer?

chewi (changes requested):
I'll confess my eyes glazed over for the massive additions. I trust it works, you always write great Bash, and you've put it through shellcheck.

I have heard that egencache might get phased out. I think pkgcore's `pmaint regen` is the new cool thing, but I haven't tried it myself. It's at least supposed to be even faster.

krnowak:
> I'll confess my eyes glazed over for the massive additions. I trust it works, you always write great Bash, and you've put it through shellcheck.

Had I picked a proper language, there'd be a lot less boilerplate. :/

> 
> I have heard that egencache might get phased out. I think pkgcore's `pmaint regen` is the new cool thing, but I haven't tried it myself. It's at least supposed to be even faster.

Haven't heard of it earlier. But if that happens, it should be a simple change in the script, because we already have pkgcore in SDK.

pkg_auto/impl/md5_cache_lib.sh
krnowak: I'm not sure if this is what you are asking about, but this order is not an order of entries showing up in the cache file on disk. These indices are for accessing `cache_file` members. If I wrote this in C, then the `cache_file` could be:

```C
struct cache_file {
    int EAPI;
    struct keyword* keywords;
    int keywords_len;
    struct iuse* iuses;
    int iuses_len;
    …
};
struct keywords kw[2] = {…};
struct iuse iu[5] = { … }
struct cache_file cf = { 8, kw, 2, iu, 5, … }:
…
printf("EAPI: %d\n", cf.EAPI);
```

Since it's bash, I implemented the struct as an array:

```bash
declare -a kw=( … )
declare -a iu=( … )
declare -a cf=( 8 kw ia … )
…
printf 'EAPI: %d\n' "${cf[PCF_EAPI_IDX]}"
```

pkg_auto/impl/pkg_auto_lib.sh
krnowak: It's a number:text pair - the number is the indentation level. It's done this way, because they will also contain lines from `diff_report` (top of the `md5_cache_diff_lib.sh`) which stores its lines in that way. The `generate_summary_stub` function handles those lines. So the example output of `lines=( "0:first" "1:second" )` could be:

```
- first
  - second
```

chewi (approved):

PR #2859, opened 2025-04-29
Krnowak/nvidia drivers bump

krnowak:
CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5877/cldsv/

Bumps nvidia drivers to 535.247.01 for amd64 and 570.133.20 for arm64.

Closes https://github.com/flatcar/Flatcar/issues/1731


krnowak:
CI kinda sorta passed - some tests were failing because of `cp: error writing 'flatcar_developer_container.bin': No space left on device`. Not sure if the test runner was running out of space or if the 5G disk starts to be not enough. I'd say the former, because the difference between the size of dev container from last nightly and from this run is around mere 1MB.

chewi:
Perhaps we shouldn't do this given that #2798 is close to merging?

jepio:
> Perhaps we shouldn't do this given that #2798 is close to merging?

We need to bump the driver version in both cases (sysext and driver service). If #2798 is going to be released at the same time as this then yeah, lets incorpote this change into that one, but this PR likely needs to be backported to other channels?

> CI kinda sorta passed - some tests were failing because of cp: error writing 'flatcar_developer_container.bin': No space left on device. Not sure if the test runner was running out of space or if the 5G disk starts to be not enough. I'd say the former, because the difference between the size of dev container from last nightly and from this run is around mere 1MB.

Which test?

dongsupark:
Friendly reminder: this PR is blocking https://github.com/flatcar/scripts/pull/2979, https://security.gentoo.org/glsa/202505-04.

chewi (approved):

PR #2872, opened 2025-05-05
app-emulation/amazon-ssm-agent: sync with upstream

tormath1:
this pulls release 3.3.2299

---

## Testing done
```
ebuild amazon-ssm-agent-3.3.2299.0.ebuild clean compile
```

- [x] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.: `Total size difference: increased by 17301797 bytes (16896 kbytes, 16 mbytes)` for the OEM ami.
- [x] CI: http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/5913/cldsv/

Note for reviewers: the changelog entry is currently 404, but the actual URL targets "latest": https://github.com/aws/amazon-ssm-agent/blob/mainline/RELEASENOTES.md#latest - which won't be the same of course in the next days / weeks. 

Closes: https://github.com/flatcar/Flatcar/issues/1738

### DIFF
--- a/changelog/updates/2025-05-05-amazon-ssm-agent.md
+++ b/changelog/updates/2025-05-05-amazon-ssm-agent.md
@@ -1,0 +1,1 @@
+- AWS: Amazon SSM Agent ([3.3.2299.0](https://github.com/aws/amazon-ssm-agent/blob/mainline/RELEASENOTES.md#3322990))


chewi (approved):
Looks trivial, but were you expecting such a big size jump? I guess it makes sense since the tarball has also increased.

PR #2976, opened 2025-05-29
build_library: Fix depmod issues with sysext kmods

danzatt:
# build_library: Fix depmod issues with sysext kmods

OS-dependent sysexts that ship kernel modules, usually also ship the files in `/usr/lib/modules/*-flatcar/modules.XXX` When multiple such sysexts get activated, depmod files from just one sysext win and other kernel modules cannot be loaded using modprobe. We get around this by removing the depmod files from every sysext with kernel modules. Instead, we set up modprobe hook, which dynamically runs depmod in a temporary directory on every sysext kernel module activation.

Fixes: [Flatcar #1576](https://github.com/flatcar/Flatcar/issues/1576)

## Testing done

### DIFF
--- a/build_library/sysext_mangle_kmod
+++ b/build_library/sysext_mangle_kmod
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+configure_modprobe() {
+  local sysext_name="${1}"
+  shift
+
+  local module_directories=(./usr/lib/modules/*-flatcar/)
+
+  mkdir -p ./usr/lib/modprobe.d/
+  for module_name in $(find "${module_directories[@]}" -type f \( -name "*.ko" -o -name "*.ko.*" \) -printf "%f\n" | sed -E 's/\.ko(\.\w+)?$//'); do
+    cat <<EOF >> "./usr/lib/modprobe.d/10-${sysext_name}-kmod-sysext.conf"
+install $module_name /usr/libexec/_${sysext_name}_modprobe_helper $module_name
+remove $module_name /usr/libexec/_${sysext_name}_modprobe_helper -r $module_name
+EOF
+  done
+
+  mkdir -p ./usr/libexec/
+  install -m0755 -D /dev/stdin "./usr/libexec/_${sysext_name}_modprobe_helper" <<'EOF'
+#!/bin/bash
+
+set -euo pipefail
+
+action="Loading"
+for arg in "$@"; do
+    if [[ $arg == "-r" ]]; then
+        action="Unloading"
+    fi
+done
+echo "$action kernel module from a sysext..."
+
+KMOD_PATH=/usr/lib/modules/$(uname -r)
+TMP_DIR=$(mktemp -d)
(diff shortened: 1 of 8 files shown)


chewi (changes requested):
Much better, thank you. We're still missing the symlinks, but other than that, it's looking good. Please check whether the `remove` hook is actually needed.

danzatt:
Forgot the symlinks, thanks for checking.

For the remove hook, I've just checked without it, and it works (removing ZFS module removes the SPL module even without the hook). The dependencies of an already loaded kmods seems to be tracked in kernel, so I've removed the remove hook.

build_library/sysext_mangle_kmod
```diff
@@ -0,0 +1,47 @@
...
+
+set -euo pipefail
+
+action="Loading"
+for arg in "$@"; do
+    if [[ $arg == "-r" ]]; then
+        action="Unloading"
+    fi
+done
+echo "$action kernel module from a sysext..."
```
chewi, line 28: I think you can simplify this part now. `-r` will never be there.

chewi (changes requested):

chewi:
> @danzatt Why do we do this at runtime, instead of generating the modules.dep file for all available kmod sysexts during image build?

I guess that would only work if we didn't have multiple different versions of the same modules. Maybe they produce the same depmod results, but we cannot always assume that.

danzatt:
@jepio Yes, Chewi is right. That was my reasoning. We currently ship sysexts with different versions (and different useflags) of nvidia kmods. I have checked all nvidia sysexts from the latest alpha (4344) and the `modules.dep` is indeed the same for all of them, but this might change accross versions in the future.

jepio:
alright - can you test this with the GPU operator? It runs modprobe from the host, i want to make sure that still works.

build_library/sysext_mangle_kmod
```diff
@@ -0,0 +1,48 @@
...
+
+  mkdir -p ./usr/lib/modprobe.d/
+  for module_name in $(find "${module_directories[@]}" -type f \( -name "*.ko" -o -name "*.ko.*" \) -printf "%f\n" | sed -E 's/\.ko(\.\w+)?$//'); do
+    cat <<EOF >> "./usr/lib/modprobe.d/10-${sysext_name}-kmod-sysext.conf"
+install $module_name     /usr/local/bin/_${sysext_name}_modprobe_helper $module_name
+remove  $module_name     /usr/local/bin/_${sysext_name}_modprobe_helper -r $module_name
+EOF
+  done
+
+  mkdir -p ./usr/local/bin/
```
jepio, line 17: I just realized /use/libexec would be a better location for these internal helpers.

There are scenarios where /ust/local needs to be writeable and that is accomplished by using a sysext that redirects /usr/local to a different path using a symlink. Using the same path would conflict with those scenarios.

jepio (changes requested):

jepio (approved):

PR #2997, opened 2025-06-04
ci: use ubuntu self hosted runners

tormath1:
GARM runners are now Ubuntu based. Let's update the workflows in consequence. To be backported on all maintenance branches.

Tested :point_down:


ader1990 (approved):
LGTM. Great seeing the runners back! Thanks!

PR #3007, opened 2025-06-06
sdk_lib: Fix starting the SDK from a fork clone without the tags

chewi:
Closes: https://github.com/flatcar/Flatcar/issues/1705

## How to use

1. Create a fork of this repo (with just main?)
2. Clone it.
3. Try to start the SDK.

## Testing done

The above. @tormath1 has tested it too. I'll run GHA now.


chewi:
Apparently, this fix isn't entirely sufficient.

PR #2556, opened 2024-12-31
[WIP] Enable Rockchip arch in kernel

sambonbonne:
# Set CONFIG_ARCH_ROCKCHIP

After some discussion on Matrix about Odroid M1S (based on RK3566), I made this PR to enable Rockchip arch in the kernel.

The goal is to test the generated aarch64 image on my hardware and to see if the initrd size is not increased to much before discussing about the possible inclusion of this configuration in Flatcar.

:warning: This PR is based on https://github.com/flatcar/scripts/pull/2300 to use 6.12 kernel, because Odroid M1S support has been added on 6.12, so it can't be merged prior to #2300.

## How to use

Installing the image in Odroid M1S requires U-Boot binaries and multiple steps. I will add testing commands if it works on my hardware and someone want to try it on real hardware.

## Testing done

No testing for now. I plan to download the generated image (when it's generated) and install it on my hardware to test if it boots.

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

PR #3027, opened 2025-06-18
Increase partition sizes

ader1990:
The /usr partition was too small some time ago and we gained space again
    by switching to btrfs with compression and also removing/splitting out
    content. The /boot partition is too small all the time and we added
    many hacks to fit the kernel+initrd under 60 MB. To handle the case
    where the /oem partition is too small for the A/B-updated OEM extensions
    we added the workaround to write the inactive one (or both) to the
    rootfs. All this would not be needed if we had increased the partition
    sizes a few years ago so that we could now assume that most nodes have
    the increased sizes and we can make use of them. Still, we can do it now
    to prepare for the next time when in five or ten years we have serious
    size problems and run out of workarounds. We have to do the change now
    and wait a few years so that most nodes have been provisioned with the
    new layout. Then we can drop the workarounds and have a full featured
    kernel and initrd, and we can also increase the /usr filesystem to make
    use of the larger partition. Ideally we use large enough sizes that we
    never have to worry again but since we also want to support small ARM
    boards which might only have 8 GB internal storage, let's target this
    when increasing the partition sizes. With 1 GB /boot, two 2 GB /usr, and
    1 GB /oem partitions we are already at 6 GB, leaving 2 GB for the
    rootfs. For now, reduce the extracted /usr update payload size to the
    current combined filesystem and verity data usage (same size as before).
    The rootfs size was also reduced for the initial .bin image so that we
    don't overshoot 8 GB - it will be resized to fit the disk anyway on
    first boot.

Testing done: GitHub Actions passed, [Jenkins passed for all platforms](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/6783/cldsv/)
Requires https://github.com/flatcar/mantle/pull/675

I created a long-term issue for tracking when we can make use of the new space: https://github.com/flatcar/Flatcar/issues/1917

Original PR:

> 
> **increase the boot partition to 256 MB**
> 
> The boot partition is currently 128 MB, and reaching a usage size that is approaching soon 50+1% when Flatcar is freshly installed. During the upgrade process, the Flatcar upgrade workflow will put the new initrd in the /boot partition, thus using more space (almost doubling the used space).
> 
> Let's make this used space calculation with a more mathematical description.
> Example of `du` output:
> 
> ```
> du -sh /boot/**/**
> 2.8M    /boot/EFI/boot
> 1.0K    /boot/boot/grub
> 1.8M    /boot/flatcar/grub
> 59M     /boot/flatcar/vmlinuz-a
> 683K    /boot/xen/pvboot-x86_64.elf
> ```
> 
> /boot partition contains the following data:
> 
>  * grub/efi/xen files - these will not change during the upgrade process - size GRUB_FILES_SIZE ~ 5 MB
>  * initial initrd size - size INITRD_PART_A_SIZE ~ 59MB
> 
> When an upgrade is made, the boot partition size increases with the new intird size INITRD_PART_B_SIZE, currently at 59MB.
> 
> So, after an upgrade, we have: GRUB_FILES_SIZE  + INITRD_PART_A_SIZE + INITRD_PART_B_SIZE ~= 5 + 59 + 59 ~= 123 MB size. Very close to 128 MB.
> 
> Doubling the boot partition will solve the future issue.
> 
> I consider that making the boot partition bigger is a more elegant and simpler solution than going with the quite complicated way presented here: 
> https://hackmd.io/@flatcar/H1MDJuu7lg
> There are security implications with the moving approach, as for example, the verity hash workflow is changed, thus the security profile changes.
> 
> The downside of this approach would be that, after some time, the upgrade of the older versions won't work anymore, but by the time it becomes an issue, the whole global fleet should be replaced with new Flatcar. Same as with cgroups issue, we can stop the upgrade with an error message if the /boot won't have enough space for existing setups.
> 

### DIFF
--- a/build_library/vm_image_util.sh
+++ b/build_library/vm_image_util.sh
@@ -224,9 +224,11 @@ IMG_ami_vmdk_DISK_FORMAT=vmdk_stream
 IMG_ami_vmdk_OEM_USE=ami
 IMG_ami_vmdk_OEM_PACKAGE=common-oem-files
 IMG_ami_vmdk_SYSEXT=oem-ami
+IMG_ami_vmdk_DISK_LAYOUT=vm
 IMG_ami_OEM_USE=ami
 IMG_ami_OEM_PACKAGE=common-oem-files
 IMG_ami_OEM_SYSEXT=oem-ami
+IMG_ami_DISK_LAYOUT=vm
 
 ## openstack
 IMG_openstack_DISK_FORMAT=qcow2
@@ -342,6 +344,7 @@ IMG_kubevirt_OEM_SYSEXT=oem-kubevirt
 IMG_kubevirt_DISK_EXTENSION=qcow2
 
 ## akamai (Linode)
+IMG_akamai_DISK_LAYOUT=vm
 IMG_akamai_OEM_PACKAGE=common-oem-files
 IMG_akamai_OEM_USE=akamai
 IMG_akamai_OEM_SYSEXT=oem-akamai


t-lo:
I think both approaches have some merit, and it's great to have a good pro/con discussion.

Re: active instances; the vast majority of Flatcar instances out there appears to use in-place updates for upgrading. All these, even the ones newly deployed today, would be affected if we only implement boot partition resizing.

Looking at the benefits of _both_ approaches on the other hand, I believe they are complementary: It would be great to have more flexibility re: partition layout and sizes in future versions of Flatcar, _and_ it would be good to have a more robust bootloader process like the one the kernel move implicitly includes.

t-lo:
Both good points, I welcome the discussion.
@ader1990 We cannot make changes that sacrifices compatibility with active deployments, that's one of the foundational guarantees we give to our users. "We don't break your deployments; ever" is one of the basic promises of the project, we've been through great lengths to uphold it - years-long cgroupsv1 backwards compatibility is a good example - and violating this would have significant impact to user trust.

Introducing _only_ the new partition format would have limited impact as we would still need to test with the smaller partition sizes, and releases failing this test won't be greenlit. There is an immediate benefit for custom-built images though, they can easily activate kernel options and increase  initrd sizes w/o messing with the partitioning themselves.

The scheme discussed in the "kernel move" document will allow us to transition without any user impact, and at the same time improve bootloader robustness. The security concepts introduced in that process also work towards our signing / trusted boot efforts (to which Chewi has made significant contributions) - if we doubt any of these, we should absolutely, and openly, discuss.

I do second your impression that we're proposing a significant change that needs to be thoroughly discussed. We brought it up in the DevSync and announced the write-up in the latest Office Hours - https://www.youtube.com/live/IXx2ZI963_Q?si=6fCyDtv3xDoBHQqP&t=899 - and we will continue to provide opportunities to further discuss. Your proposal of getting user feedback is awesome, we should definitely take a stab at this.

chewi:
#2556 is actually one of my big motivators for doing this. If we were to merge that today, it would immediately push us over the limit. Postponing the migration to when it's absolutely required is just delaying the inevitable, and not even for very long. It would also make upgrades fairly unpredictable for users. Will it do the migration this time or not?

I don't believe these changes will make arm64 or RISC-V support any harder. If anything, allowing GRUB upgrades and more flexibility with command line arguments could make it easier.

The ability to switch to a pre-migration slot is an essential part of the work. The new GRUB configuration will know how to boot the older kernel. With the new configuration being more flexible, that's not difficult to do. For UEFI systems, the old GRUB image will be installed as a fallback to be invoked by the shim or manually. I also aim to have a menu entry in the new GRUB pointing to the old GRUB. I'm not planning to allow for an actual downgrade (as opposed to just switching the slot) but I gather we don't support that anyway.

My document is fairly long because I went into some detail explaining how we'll get there and why I've taken this approach. The _actual_ amount of change may be less than you expect. Most of it concerns the load-a script. Much of it is literally changing the kernel path or just reordering existing parts of the build process.

I know there has been interest in UKIs, so I did say early on that having the kernel in /usr would rule them out because EFI firmware cannot load from btrfs. Maybe you can load them from GRUB, but then what's the point? I'm not convinced that Flatcar needs to go down the UKI route. They can simplify the security model, but at a cost of flexibility that I don't think Flatcar can afford. Not using UKIs doesn't make things inherently insecure. Given all that, I don't see my proposal as a temporary workaround.

Sorry, I'm not sure what having USR-A/USR-B as entire disk images has to do with UKIs.

t-lo:
@ader1990 would you be available for discussing this (open ended! No bias from our side) in the dev sync on Wednesday? Would be awesome to chat about this more broadly, with the occasional deep dive.

ader1990:
> @ader1990 would you be available for discussing this (open ended! No bias from our side) in the dev sync on Wednesday? Would be awesome to chat about this more broadly, with the occasional deep dive.

Hello, a discussion is always good.

pothos:
Thanks for the PR, yes, this increase has to happen and should have happened already a few years ago so that we could have relied now on most nodes having a large enough boot partition so that we don't need any size tweaks… but here we are and so I did one more, hopefully final, size tweak with the minimal initrd work: https://github.com/flatcar/scripts/pull/3241 This reduces the vmlinuz size to a half, so that we have much room left again to enable all these kernel features we want.

With the partition size increase we plan ahead for the future. A few years later we can then have a vmlinuz size greater than 60 MB and remove the minimal initrd again. This means long-running nodes with a 128 MB partition won't be able to update but we can let them update with an opt-in removal of the currently running vmlinuz file shortly before the update. This means rollbacks won't work, though, but it still might be an interesting tradeoff for those who want to avoid/delay a redeployment.

On our side we would need to create an issue and a calendar entry for a future date that captures the plans so that they are not forgotten. When we touch this area and plan ahead like this I would suggest to do a bold move: Increase /boot to at least 1 GB, increase /usr to at least 2 GB, and /oem to at least 1 GB. These numbers are small and with 1+2*2+1 are only 6 GB in total, leaving a 2 GB room for a rootfs when deployed on a 8 GB storage device - but seriously, this is still a very conservative approach and normally all storage is plenty! We could also do 2 GB /boot, 3 GB /usr and 2 GB /oem and be at 2+2*3+2=10 GB, which is tiny for most servers and also most SBCs have appropriate storage options.
In the past /usr was already full and we took carefully planned measures to reduce the used size. The OEM partition is also too small and we spill A/B-updated OEM sysexts over to the root partition as workaround. Having a very large boot partition is not only future proof but also allows us to include more features without having to worry about size again, be it kernel or initrd features or additional extensions getting stored on the boot partition.

Increasing the partition size for /usr would need to still leave the filesystem size at 1 GB for now until the future date where we can switch over to generate a larger /usr filesystem which results in a larger update payload (uncompressed) and thus requires most nodes to have the increased partiton layout. In contrast to the /boot partition there is no workaround possible for outdated nodes.

This has to come anyway, doing it now leaves us more time to wait until we can be sure the new layout is present on most nodes and do the switch to finally be able to benefit from larger sizes.

ader1990:
@pothos had to re-push so that I respect the new DCO rule, now the PR is ready.

Thanks.

chewi:
I dunno, I think 1GB or even 512MB seems like overkill for /boot now that you've taken action. We're not a desktop distro, so we don't need all the drivers that other distros do. We only keep 2 at a time while some other distros keep more. You got the image down to 27MB. You want to store 37 of them? :grin:

pothos:
With the original/regular setup for the initrd we would have no space left since years ago. We add hacks like the first stage initrd because we have to, not we want to ;) Best is if we can drop all these workarounds in 3-4 years and enable a lot of kernel features that people expect to be there from other distros. That might require > 100 MB for kernel+initrd.

ader1990:
I also agree that this change should have happened two years ago, when we started the discussion. Diverging from simplicity and embracing complexity should be, well, not be. But on the other hand, as we are now forced to do something and do it soon, this also brings inovation.

@pothos feel free to take over the PR if needed, to adjust the numbers as you see fit. 1GB/2Gb/1GB sounds reasonable for the future (as long as we do not need to put future AGI models inthere).

The verity offset is just pertaining to the initrd infile location, so it should not be affected afaik?

ader1990:
The only question here related to space is if we want to still support small edge devices, that have less than 4GB of disk space (emmc, sd cards). The current Flatcar image partitioning scheme uses > 4GB, it is somewhere around 4.4 GB. What we can do is to at least make sure we don t go over the 8GB if not needed 100%.

ader1990:
Also, PXE booting requires allocating all the image into RAM, and same as with the disk size, we should still support booting on edge devices with < 8GB RAM.

chewi:
Media size is a good point, but also remember that firmware will be duplicated in USR and the initrd, so adding GPU firmware could bloat the image download size quite a lot. Recent Dracut versions can do clever things to avoid that duplication, but it won't help Flatcar because of how it is built and also because it requires the initrd to be on the same partition. We will now have the second stage initrd on the same partition, but we're talking about the minimal initrd here. Presumably putting the firmware in the second stage initrd isn't good enough if you're concerned about the emergency shell.

pothos:
Yes, I don't think we would fill 1 GB soon but the Fedora story shows that past assumptions can be proven wrong quite quickly. E.g., even if we don't care about GPU firmware it could be new NIC firmware instead :upside_down_face:

The PXE boot doesn't load the partition layout to RAM, only the compressed `/usr` contents as a squashfs image wrapped in an additional initrd cpio.

build_library/disk_util
```diff
@@ -136,6 +136,8 @@ def LoadPartitionConfig(options):
       part.setdefault('fs_block_size', metadata['fs_block_size'])
       part.setdefault('fs_blocks', part['bytes'] // part['fs_block_size'])
       part['fs_bytes'] = part['fs_blocks'] * part['fs_block_size']
+      if part.get('extract_blocks', None):
```
tormath1: That'd be nice to have a comment here defining what is the `extract_blocks`. Sometimes it's used (`rootfs`), sometimes not (the rest)

PR #3063, opened 2025-06-30
profiles: Prefix comment with # in package.mask

jepio:
# profiles: Prefix comment with # in package.mask

This multi-line comment accidentally only has a # prefix on the first line, which leads to this warning during the build:

    --- Invalid atom in /mnt/host/source/src/third_party/coreos-overlay/profiles/coreos/base/package.mask: mask a specific version in hope that the future update may be smaller,
    --- Invalid atom in /mnt/host/source/src/third_party/coreos-overlay/profiles/coreos/base/package.mask: who knows.

Add missing prefixes to lines 2 and 3.

## How to use

Check build log to see that those lines no longer show up.

## Testing done


krnowak:
Thanks!

PR #3080, opened 2025-07-03
A couple of small fixes for performing Flatcar builds locally

chewi:
# A couple of small fixes for performing Flatcar builds locally

The shim packages must be installed when running grub_install.sh. These are not present in a fresh SDK instance.

The coreos-kernel dependencies must be tightened to ensure working initrd. In particular, sys-fs/lvm2 includes dmsetup, and systemd will fail to recognise /dev/mapper/usr if that is built without udev support.
    
This is not a problem in CI, but a fresh SDK will have board packages installed without their final USE flags in place due to cyclic dependencies.

## How to use

Perform a local build.

## Testing done

Both changes have been testing locally. The grub_install.sh change has had several runs through CI.

- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc. -- N/A

### DIFF
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -75,7 +75,9 @@ esac
 info "Updating GRUB in ${BOARD_ROOT}"
 emerge-${BOARD} \
         --nodeps --select --verbose --update --getbinpkg --usepkgonly --newuse \
-        sys-boot/grub
+        sys-boot/grub \
+        sys-boot/shim \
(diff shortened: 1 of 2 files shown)


t-lo (approved):
LGTM. Kicked off a build for thoroughness, let's have it conclude and merge after that.

PR #3141, opened 2025-07-25
Upgrade Linux Kernel for flatcar-4081 from 6.6.98 to 6.6.100

flatcar-infra:
Subject says it all.


sayanchowdhury (approved):
CI green(only Openstack having transient issues)

PR #3156, opened 2025-07-29
Publish AZ SIG images

sayanchowdhury:
To be merged with https://github.com/flatcar/jenkins-os/pull/371

The script publishes vhd images to Azure Shared Image Galleries

Pending Item:
- Need to move the download into the `az_sig_publish`. lbzip2 is not available in package repos. Also, need to check if curl works.
- replication to target regions is still not working. need to look into
- remove cleanup items before merge.
- update default values before merge 

CI http://jenkins.infra.kinvolk.io:8080/job/container/job/az_sig_publish/98/consoleFull


az_sig_publish
```diff
@@ -0,0 +1,194 @@
...
+
+  if [[ -z "${FLATCAR_LOCAL_FILE_URL:-}" ]]; then
+    echo "FLATCAR_LOCAL_FILE_URL variable is empty. Please set."
+    exit 1
+  fi
+
+  echo "Blob doesn't exist. Downloading flatcar-linux-${FLATCAR_GALLERY_VERSION}-${FLATCAR_CHANNEL}-${FLATCAR_ARCH}"
+  # Download the VHD file (only if blob doesn't exist)
+  echo "Downloading ${FLATCAR_LOCAL_FILE_URL}..."
+  if ! curl -L -o "/data/${FLATCAR_LOCAL_FILE_NAME}.bz2" "${FLATCAR_LOCAL_FILE_URL}"; then
```
chewi, line 90:
```suggestion
  if ! curl -f -L -o "/data/${FLATCAR_LOCAL_FILE_NAME}.bz2" "${FLATCAR_LOCAL_FILE_URL}"; then
```

chewi (changes requested):
Nice work, thanks.

Just a general tip, although Gentoo is admittedly quote-heavy in its ebuilds, you don't need to quote:

* When assigning `FOO=${}` or `FOO=$()`
* Inside `[[ ]]` unless it's a variable on the right-hand-side.
* `case ${FOO} in`

PR #3157, opened 2025-07-29
Add overlaybd sysext

t-lo:
This adds overlaybd, a user space container-optimised block device implementing a novel layering block-level image format, which is designed for containers, with security and performance in mind.

Documentation is available at https://containerd.github.io/overlaybd/.

**sys-fs/overlaybd** ships the low-level user space block device (https://github.com/containerd/overlaybd) as well as low-level tooling for operating it.

**app-containers/accelerated-container-image** (https://github.com/containerd/accelerated-container-image) ships integration with containerd.

Both are part of the Containerd CNCF project.

# BUILDING

1. Check out this branch
   ```sh
   git checkout t-lo/containerd-overlaybd
   ```
2. Start the SDK
   ```sh
   ./run_sdk_container -t
   ```
3. Build packages and image
   ```sh
    ./build_packages
    ./build_image prod sysext
    ./image_to_vm.sh --from=../build/images/amd64-usr/latest --image_compression_formats=none
   ```

## Rebuilding

After an initial build only the sysext needs to be rebuilt.
To limit the scope of system-dependent sysext builds it is useful to set
```sh
EXTRA_SYSEXTS=(                                                          
  "overlaybd|sys-fs/overlaybd,app-containers/accelerated-container-image"
)                                                                        
```
in `build_library/prod_image_util.sh`.

To pick up modifications in `overlaybd`, run
```sh
emerge-amd64-usr --buildpkg overlaybd
```

Similarly, to include `accelerated-container-image` changes, use
```sh
emerge-amd64-usr --buildpkg accelerated-container-image
```

Then rebuild the sysext(s):
```sh
./build_image sysext
```
The new `flatcar-overlaybd.raw` sysext is now available in `__build__/images/images/amd64-usr/latest`.

# TESTING

Testing is manual at this point because `kola` cannot (yet?) provision arbitrary system extensions to qemu test instances. Also, testing is limited to AMD64 as no overlaybd enabled ARM64 images are [available on dockerhub](https://hub.docker.com/search?q=overlaybd&architecture=arm64).

1. Provision `flatcar-overlaybd` extension on test instance. The [sysext bakery's ](https://github.com/flatcar/sysext-bakery)`bakery.sh boot` is particularly useful. In `__build__/images/images/amd64-usr/latest`, run
   ```sh
   ~/code/sysext-bakery/bakery.sh boot flatcar-overlaybd.raw --nutane custom.yaml
   ```
   <details><summary>Here's the "custom.yaml" butane config:</summary>
   
   ```yaml
   version: 1.0.0
   variant: flatcar
  
   storage:
     files:
       - path: /etc/config.toml
         contents:
           inline: |
             version = 2
             
             # persistent data location
             root = "/var/lib/containerd"
             # runtime state information
             state = "/run/containerd"
             # set containerd as a subreaper on linux when it is not running as PID 1
             subreaper = true
             # set containerd's OOM score
             oom_score = -999
             disabled_plugins = []
             
             # grpc configuration
             [grpc]
             address = "/run/containerd/containerd.sock"
             # socket uid
             uid = 0
             # socket gid
             gid = 0
             
             [plugins."io.containerd.runtime.v1.linux"]
             # shim binary name/path
             shim = "containerd-shim"
             # runtime binary name/path
             runtime = "runc"
             # do not use a shim when starting containers, saves on memory but
             # live restore is not supported
             no_shim = false
             
             [plugins."io.containerd.grpc.v1.cri"]
             # enable SELinux labeling
             enable_selinux = true
             
             [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
             # setting runc.options unsets parent settings
             runtime_type = "io.containerd.runc.v2"
             [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
             SystemdCgroup = true
             
             [proxy_plugins.overlaybd]
                type = "snapshot"
                address = "/run/overlaybd-snapshotter/overlaybd.sock"
       - path: /etc/extensions/flatcar-overlaybd.raw
         mode: 0644
         contents:
           # QEmu's default traffic-to-host IP
           source: http://10.0.2.2:12345/flatcar-overlaybd.raw
   
   systemd:
     units:
       - name: update-engine.service
         mask: true
       - name: locksmithd.service
         mask: true
       - name: containerd.service
         dropins:
           - name: 10-configtoml.conf
             contents: |
               [Service]
               ExecStart=
               ExecStart=/usr/bin/containerd --config /etc/config.toml
    ```

   </details>
2.  Start redis test image as per
   https://github.com/containerd/accelerated-container-image/blob/main/docs/QUICKSTART.md#run-overlaybd-images
   ```sh
    sudo /opt/overlaybd/snapshotter/ctr rpull registry.hub.docker.com/overlaybd/redis:6.2.1_obd
    sudo ctr run --net-host --snapshotter=overlaybd --rm -t registry.hub.docker.com/overlaybd/redis:6.2.1_obd demo
   ```
   and on a separate console run
   ```sh
    journalctl --no-pager -f
   ```
   and/or
   ```
   tail -f /var/log/overlaybd.log
   ```

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
@@ -88,14 +88,12 @@ src_install() {
 	mv "${ED}"/opt/${PN}/* "${ED}"/usr/local/${PN}/ || die
 	mv "${ED}"/etc/${PN}/* "${ED}"/usr/local/${PN}/etc/ || die
 
-  # Handle /etc (overlaybd.json), create /opt/opverlaybd and symlink
+	# Handle /etc (overlaybd.json), create /opt/overlaybd and symlink
 	# all contents of /usr/local/overlaybd to /opt/overlaybd.
-  elog "Scanning '${ED}/usr/local/${PN}/' and generating tmpfiles symlink entries..."
-	cp "${FILESDIR}"/10-${PN}.conf "${T}"
-  local entry
-  for entry in $(ls -1 "${ED}/usr/local/${PN}/") ; do
-    echo "L /opt/overlaybd/${entry} - - - - /usr/local/${PN}/${entry}" \
-      | tee -a "${T}/10-${PN}.conf"
-  done
-	dotmpfiles "${T}"/10-${PN}.conf
+	elog "Scanning '${ED}/usr/local/${PN}/' and generating tmpfiles symlink entries..."
+	cat "${FILESDIR}"/10-${PN}.conf <(
+		for entry in "${ED}"/usr/local/${PN}/*; do
+			echo "L /opt/overlaybd/${entry##*/} - - - - /usr/local/${PN}/${entry##*/}"
+		done
+	) | tee /dev/stderr | newtmpfiles - 10-${PN}.conf
 }
--- a/sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
@@ -1,0 +1,101 @@
+# Copyright 2025 The Flatcar Container Linux Maintainers
+# Distributed under the terms of the Apache License 2.0
+
+EAPI=8
+
+inherit cmake git-r3 systemd tmpfiles
+
+DESCRIPTION="Novel layering block-level image format for containers"
+HOMEPAGE="https://containerd.github.io/overlaybd"
+EGIT_REPO_URI="https://github.com/containerd/overlaybd.git"
+
+if [[ ${PV} == 9999* ]]; then
+	KEYWORDS="~amd64 ~arm64"
+else
+	EGIT_COMMIT="v${PV}"
+	KEYWORDS="amd64 arm64"
+fi
+
+LICENSE="Apache-2.0"
+SLOT="0"
+IUSE="cpu_flags_x86_avx2 dsa qat isal"
+REQUIRED_USE="dsa? ( cpu_flags_x86_avx2 )"
+RESTRICT="test"
+
+# FIXME HACK ALERT: overlaybd build pulls sources during src_configure.
+# (https://github.com/alibaba/PhotonLibOS.git/
+# This fails if network sandbox is enabled.
+RESTRICT="${RESTRICT} network-sandbox"
+
+DEPEND="
+	app-arch/zstd:=
+	dev-libs/libaio
+	dev-libs/libnl:3
+	dev-libs/openssl:=
+	net-misc/curl
+	sys-fs/e2fsprogs
+	sys-libs/zlib
+	dsa? ( sys-apps/pciutils )
+	qat? ( sys-apps/pciutils )
+"
+
+RDEPEND="
+	${DEPEND}
+"
+
+PATCHES=(
+	"${FILESDIR}"/${PN}-cross.patch
+)
+
+src_prepare() {
+	cmake_src_prepare
+	sed -i "s:@FILESDIR@:${FILESDIR}:g" CMake/Findphoton.cmake || die
+}
+
+src_configure() {
+	# crc32c.cpp explicitly uses special instructions but checks for them at
+	# runtime. Only DSA hard requires at least AVX2. However, the code doesn't
+	# try especially hard to avoid these instructions from being implicitly used
+	# outside these runtime checks. :(
+  # ISAL similarly leads to "illegal instruction" termination on QEMU.
+	local mycmakeargs=(
+		-DBUILD_SHARED_LIBS=no
+		-DBUILD_TESTING=no
+		-DENABLE_DSA=$(usex dsa)
+		-DENABLE_ISAL=$(usex isal)
+		-DENABLE_QAT=$(usex qat)
+		-DORIGIN_EXT2FS=yes
+	)
+
+	# Make erofs-utils configure work when cross-compiling.
+	host_alias="${CHOST}" build_alias="${CBUILD:-${CHOST}}" \
+	cmake_src_configure
+}
+
+src_install() {
+	cmake_src_install
+
+	# We want to ship our binaries in /usr/local (so we're sysext compatible)
+	# but upstream has hard-wired everything to /opt/overlaybd.
+
+	sed "s,/opt/${PN},/usr/local/${PN},g" \
+		"${ED}"/opt/${PN}/${PN}-tcmu.service |
+		systemd_newunit - ${PN}-tcmu.service
+	rm "${ED}"/opt/${PN}/${PN}-tcmu.service || die
+	systemd_enable_service multi-user.target ${PN}-tcmu.service
+
+	dodir /usr/local/${PN}/etc
+	mv "${ED}"/opt/${PN}/* "${ED}"/usr/local/${PN}/ || die
+	mv "${ED}"/etc/${PN}/* "${ED}"/usr/local/${PN}/etc/ || die
+
+  # Handle /etc (overlaybd.json), create /opt/opverlaybd and symlink
+	# all contents of /usr/local/overlaybd to /opt/overlaybd.
+  elog "Scanning '${ED}/usr/local/${PN}/' and generating tmpfiles symlink entries..."
+	cp "${FILESDIR}"/10-${PN}.conf "${T}"
+  local entry
+  for entry in $(ls -1 "${ED}/usr/local/${PN}/") ; do
+    echo "L /opt/overlaybd/${entry} - - - - /usr/local/${PN}/${entry}" \
+      | tee -a "${T}/10-${PN}.conf"
+  done
+	dotmpfiles "${T}"/10-${PN}.conf
+}
--- a/sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
@@ -49,7 +49,7 @@ PATCHES=(
 
 src_prepare() {
 	cmake_src_prepare
-	sed -i "s:@FILESDIR@:${FILESDIR}:g" CMake/Findphoton.cmake || die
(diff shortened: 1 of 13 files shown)


t-lo:
Testing on AMD64 ran into `illegal instruction` errors when using ISAL; will disable it and leave a comment in the ebuild.

t-lo:
Also, when testing I'm seeing multiple
```
Jul 30 10:15:13 localhost overlaybd-tcmu[1652]: *** buffer overflow detected ***: terminated
Jul 30 10:15:17 localhost systemd[1]: overlaybd-tcmu.service: Main process exited, code=dumped, status=6/ABRT
Jul 30 10:15:17 localhost systemd[1]: overlaybd-tcmu.service: Failed with result 'core-dump'.
```
on both ARM64 and AMD64. This eventually stops and the test redis starts on AMD64 (not on ARM64 since test container images are only available for x86). But start-up takes forever because systemd restarts `overlaybd-tcmu` at least 20 times.

jepio:
Related: https://github.com/containerd/overlaybd/issues/370

krnowak:
So the failing code seems to be this one in https://github.com/data-accelerator/photon-libtcmu/blob/main/scsi.cpp#L255-L256:

```c
		len = snprintf(&ptr[4], sizeof(data) - used - 4, "%s",
			       tcmu_dev_get_cfgstring(dev));
```

`data` is `char[512]`, `ptr` points somewhere inside it. At the time of execution, `ptr` is `&data[41]`. `tcmu_dev_get_cfgstring` returns a string of length 101. So it should be working fine. But at glibc side we invoke `snprintf`, and `snprintf` invokes `__snprintf_chk` like so:

```c
  return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
                                   __glibc_objsize (__s), __fmt,
                                   __va_arg_pack ());
```

The failing check is this code inside `__snprintf_chk`:

```
  if (__glibc_unlikely (slen < maxlen))
    __chk_fail ();
```

`maxlen` is 475` in our case, `slen` is a result of `__glibc_objsize (__s)`, where `__s` is the string returned by `tcmu_dev_get_cfgstring(dev)`. The catch here is that this function returns `dev->cfgstring` and `cfgstring` in `struct tcmu_device` is defined as `char cfgstring[PATH_MAX]` (so an array of 4096 chars).

My guess is that `__glibc_objsize` returns 4096, so `__snprintf_chk` bails out.

`__glibc_objsize` is defined in `/usr/include/sys/cdefs.h` and its definition depends on fortification level.

Possibly a workaround could be to lower the fortification configuration we use, or do a strdup of the device's cfgstring and pass the result to `snprintf`.

t-lo:
> Another possible workaround could be increasing size of `data` from 512 to 512 + `PATH_MAX` and then take care of places where `sizeof(data)` is used. But this is rather silly.

This seems cleaner to me since it would ensure that `dev->cfgstring` always fits? Even using plain `PATH_MAX` would suffice, no?

jepio:
> `maxlen` is 475 in our case, `slen` is a result of `__glibc_objsize (__s)`, where `__s` is the string returned by `tcmu_dev_get_cfgstring(dev)`. The catch here is that this function returns `dev->cfgstring` and `cfgstring` in `struct tcmu_device` is defined as `char cfgstring[PATH_MAX]` (so an array of 4096 chars).

Is it possible that the string is not null terminated?

t-lo:
@krnowak I patched `scsi.cpp` with
```diff
diff --git a/scsi.cpp b/scsi.cpp
index d8c27a9..0113eb2 100644
--- a/scsi.cpp
+++ b/scsi.cpp
@@ -181,7 +181,7 @@ int tcmu_emulate_evpd_inquiry(
 	break;
 	case 0x83: /* Device identification */
 	{
-		char data[512];
+		char data[PATH_MAX+512]; //tcmu_dev_get_cfgstring(dev) may be up to PATH_MAX length
 		char *ptr, *p, *wwn;
 		size_t len, used = 0;
 		uint16_t *tot_len = (uint16_t*) &data[2];
```

and I can still see the issue - same buffer overflow, same stack trace.

krnowak:
Bah, I'll add some debug symbols for glibc and investigate more tomorrow.

t-lo:
So I did a bit of `printf` debugging (I know...) and it seems that line 255 in `scsi.cpp` indeed causes the issue.
Just the proposed workaround of using `char data[PATH_MAX+512];`  doesn't work for some reason.

krnowak:
Continuation of the analysis from above:

I'll bring three code snippets first.

The callsite of snprintf:

```c
		len = snprintf(&ptr[4], sizeof(data) - used - 4, "%s",
			       tcmu_dev_get_cfgstring(dev));
```

The invocation of `__snprintf_chk` inside `snprintf`:

```c
  return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
                                   __glibc_objsize (__s), __fmt,
                                   __va_arg_pack ());
```

The check inside `__snprintf_chk`:

```
  if (__glibc_unlikely (slen < maxlen))
    __chk_fail ();
```


I made a bunch of mistakes earlier:

I wrote that `__glibc_objsize` is called on the result of `tcmu_dev_get_cfgstring()` and it possibly returns something like 4096 (or PATH_MAX) and this being larger trips up the condition in `__snprintf_chk`. This is wrong.

1. `__glibc_objsize` is called on `&ptr[4]` that is passed as a first parameter to `snprintf`.
2. The check is `if (__glibc_objsize(buffer) < buffer_size) fail;`, so `4096 < 475` (`475` is the result of `sizeof(data) - used - 4` in the call of `snprintf`) evaluates to false and failure wouldn't happen.

So, actually:

The result of `__glibc_objsize` on that pointer is `471` and the condition check is `471 < 475`. This looks like plain mistake in accounting of used memory in `data`.

I will be testing the following patch:
```diff
From 0ee2c928de51ef56376c4ab6528ca1e23c13c4a7 Mon Sep 17 00:00:00 2001
From: Krzesimir Nowak <knowak@microsoft.com>
Date: Fri, 8 Aug 2025 16:29:52 +0200
Subject: [PATCH] fix

---
 scsi.cpp | 8 ++++----
 1 file changed, 4 insertions(+), 4 deletions(-)

diff --git a/scsi.cpp b/scsi.cpp
index d8c27a9..4c7f480 100644
--- a/scsi.cpp
+++ b/scsi.cpp
@@ -183,7 +183,7 @@ int tcmu_emulate_evpd_inquiry(
 	{
 		char data[512];
 		char *ptr, *p, *wwn;
-		size_t len, used = 0;
+		size_t len, used = 4;
 		uint16_t *tot_len = (uint16_t*) &data[2];
 		uint32_t padding;
 		bool next;
@@ -207,7 +207,7 @@ int tcmu_emulate_evpd_inquiry(
 
 		ptr[3] = 8 + len + 1;
 		used += (uint8_t)ptr[3] + 4;
-		ptr += used;
+		ptr += (uint8_t)ptr[3] + 4;
 
 		/* 2/5: NAA binary */
 		ptr[0] = 1; /* code set: binary */
@@ -340,9 +340,9 @@ int tcmu_emulate_evpd_inquiry(
 finish_page83:
 		/* Done with descriptor list */
 
-		*tot_len = htobe16(used);
+		*tot_len = htobe16(used - 4);
 
-		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, used + 4);
+		tcmu_memcpy_into_iovec(iovec, iov_cnt, data, used);
 
 		free(wwn);
 		wwn = NULL;
-- 
2.49.1

```

chewi:
See how overlaybd-cross.patch references photon-cross.patch. You'll need to add another two patches.

krnowak:
> See how overlaybd-cross.patch references photon-cross.patch. You'll need to add another two patches.

Yeah, thought about that shortly after writing this message. Thanks.

chewi:
FYI, upstream merged my PRs very quickly, although overlaybd is using quite an old version of photon, so they'd need to catch that up before they could take that fix.

sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
```diff
@@ -0,0 +1,101 @@
...
+  # Handle /etc (overlaybd.json), create /opt/opverlaybd and symlink
+	# all contents of /usr/local/overlaybd to /opt/overlaybd.
+  elog "Scanning '${ED}/usr/local/${PN}/' and generating tmpfiles symlink entries..."
+	cp "${FILESDIR}"/10-${PN}.conf "${T}"
+  local entry
+  for entry in $(ls -1 "${ED}/usr/local/${PN}/") ; do
+    echo "L /opt/overlaybd/${entry} - - - - /usr/local/${PN}/${entry}" \
+      | tee -a "${T}/10-${PN}.conf"
+  done
+	dotmpfiles "${T}"/10-${PN}.conf
```
chewi, line 100:
```suggestion
	# Handle /etc (overlaybd.json), create /opt/overlaybd and symlink
	# all contents of /usr/local/overlaybd to /opt/overlaybd.
	elog "Scanning '${ED}/usr/local/${PN}/' and generating tmpfiles symlink entries..."
	cat "${FILESDIR}"/10-${PN}.conf <(
		for entry in "${ED}"/usr/local/${PN}/*; do
			echo "L /opt/overlaybd/${entry##*/} - - - - /usr/local/${PN}/${entry##*/}"
		done
	) | newtmpfiles - 10-${PN}.conf
```

chewi (changes requested):

t-lo:
Not sure about that, for one they use cpp instead of plain C files, and then there are patches e.g. on top of scsi.cpp that are not in tcmu-runner's scsi.c (https://github.com/open-iscsi/tcmu-runner/commits/main/scsi.c vs. https://github.com/data-accelerator/photon-libtcmu/commits/main/scsi.cpp). So photon-libtcmu is more recent.

It looks to me as if photon-libtcmu started itself from a bunch of files from tcmu-runner, but has since diverged.

sdk_container/src/third_party/coreos-overlay/sys-fs/overlaybd/overlaybd-9999.ebuild
t-lo: Wouldn't that drop the `tee`? (i.e. printing the scan results to the build log)?

t-lo: @chewi ☝️

PR #3159, opened 2025-07-31
systemd: backport network-nic-unmanaged bugfix to v256

jepio:
# systemd: backport network-nic-unmanaged bugfix to v256

This is a backport for issue https://github.com/systemd/systemd/issues/36997 which I have been hitting in testing. It seems networkd is reevaluating if a nic should be unmanaged when it reconfigures a nic (on link up/down), which can result in performing DHCP on Azure's enslaved NIC, breaking networking to the instance. This has been seen on Arm64 instances in Azure.

I will be running this branch through kola testing on various arm64 instances (v5/v6) hoping to catch the issue.

## How to use

Check build log to see patches applied, check for no regressions on other platforms.

## Testing done
- not easy to repro, will paste link to kola jobs


chewi (changes requested):
You need to update the version in sdk_container/src/third_party/coreos-overlay/profiles/coreos/targets/sdk/package.accept_keywords.

chewi (approved):

PR #3162, opened 2025-08-01
Signed OS-dependent sysexts

danzatt:
# Sign OS-dependent sysexts using an ephemeral key which is baked into the image.

The signed sysexts are generated using `systemd-repart` and they are built as a DDI with a dm-verity partition and signature partition.

TODO: determine if we enable the `CONFIG_DM_VERITY_VERIFY_ROOTHASH_SIG` in this PR or separate (@chewi was planning to use this for the kernel move, but the plans have shifted now).

## How to use

Just enable any OS dependent sysexts, they should be verified using dm-verity has signature when loading.

## Testing done

When you boot the built image, all OS-dependent sysexts (including OEM) should be signed. When booting, the systemd-sysext will merge them and verify the signature automatically. To merge again with explicitly stricter policy (which forces signed sysexts only), you can use the following command:
`systemd-sysext refresh --image-policy="root=verity+signed+absent:usr=verity+signed+absent"`

See related sysext-bakery [PR#175](https://github.com/flatcar/sysext-bakery/pull/175)


chewi (changes requested):
Aside from the security model, I have one concern. systemd-repart only gained the ability to apply compression to erofs in v257, but as far as I can tell, this is only configurable through the repart.d files, which aren't used when creating DDIs. The built-in sysexts will therefore no longer be compressed. Maybe this doesn't matter because USR is already compressed via btrfs. It might even be a good thing as we avoid the double compression. Please compare the compressed size on a live Flatcar instance against an older instance.

We'll presumably need to do work on systemd to be able to publish signed and compressed sysexts in the bakery. CC @t-lo.

changelog/changes/2025-11-05-signed-built-in-sysexts.md
```diff
@@ -0,0 +1 @@
+- Built-in system extensions (e.g., docker-flatcar, containerd-flatcar) are now cryptographically signed using dm-verity roothash signatures. This enables stricter sysext policies via systemd-sysext and provides a foundation for verifying user-provided extensions in future releases. The format changed from squashfs to erofs-based Discoverable Disk Images (DDI). ([scripts#3162](https://github.com/flatcar/scripts/pull/3162))
```
pothos: Are the Flatcar extensions such as podman, python, zfs also covered?

sdk_lib/sdk_entry.sh
```diff
@@ -88,6 +88,43 @@ if ! grep -q 'export MODULE_SIGNING_KEY_DIR=' /home/sdk/.bashrc; then
...
+fi
+grep -q 'export SYSEXT_SIGNING_KEY_DIR' /home/sdk/.bashrc || {
+    if [[ ${COREOS_OFFICIAL:-0} -eq 1 ]]; then
+        SYSEXT_SIGNING_KEY_DIR=$(su sdk -c "mktemp -d")
+    else
+        SYSEXT_SIGNING_KEY_DIR="/home/sdk/.sysext-signing-keys"
+        su sdk -c "mkdir -p ${SYSEXT_SIGNING_KEY_DIR@Q}"
+    fi
+    if [[ ! "$SYSEXT_SIGNING_KEY_DIR" || ! -d "$SYSEXT_SIGNING_KEY_DIR" ]]; then
+        echo "Failed to create directory for sysext signing keys."
```
pothos: Same here, if we really want to continue, better make the error more visible

chewi (approved):

PR #3180, opened 2025-08-11
sys-kernel/coreos-modules: Reenable console support for DRM

pothos:
When the virtio graphic mode instead of qemu's bochs was used, the
interactive console did not show up because fbdev emulation recently
got disabled.
Reenable fbdev emulation and also add the "simple" and QXL FB DRM
driver to make TTM_HELPER available for the out-of-tree nvidia driver
(which is needed when fbdev emulation is enabled).

Fixes https://github.com/flatcar/Flatcar/issues/1834

## Testing done

`./flatcar_production_qemu.sh -- -vga virtio` now gives an interactive login prompt

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
@@ -7,7 +7,7 @@ EGIT_REPO_URI="https://github.com/flatcar/bootengine.git"
 if [[ "${PV}" == 9999 ]]; then
 	KEYWORDS="~amd64 ~arm ~arm64 ~x86"
 else
-	EGIT_COMMIT="1eb892c1970a158247b3709f81d03b44de7d806e" # flatcar-master
+	EGIT_COMMIT="31ba2964ce5e77ae5553eb0a3624afcc7078bb09" # flatcar-master
 	KEYWORDS="amd64 arm arm64 x86"
 fi
 


jepio:
If it works then good but it doesn't look like `SIMPLEDRM` selects `DRM_TTM_HELPER` so this might not work.

pothos:
I first tried it with DRM_BOCHS and that worked. But it has the side effect of giving a large screen in the qemu script (same as with the uefi script). I tried with simple DRM but that isn't enough to enable the helpers required for nvidia. I guess the bochs support is an ok workaround? It would mean that the qemu window size is always the same in the end, regardless of UEFI or BIOS. One can still get the old behavior with setting, e.g., `-vga qxl`. We could also try it the other way round, otherwise: Enable DRM support for something like qxl and then the bochs mode would stay as is.

pothos:
With bochs the vmlinuz size increase was 200 KB

pothos:
Ah, with QXL only TTM but not VRAM gets enabled. Yes, I don't know if there's a better way (`CONFIG_DRM_VIRTIO_GPU=m` is already in use).

pothos:
Jeremi only mentioned TTM, maybe that's enough then?

jepio:
> Ah, with QXL only TTM but not VRAM gets enabled. Yes, I don't know if there's a better way (`CONFIG_DRM_VIRTIO_GPU=m` is already in use).

I think any one of these work:
```
$ git grep DRM_TTM_HELPER
drivers/gpu/drm/Kconfig:232:config DRM_TTM_HELPER
drivers/gpu/drm/Makefile:128:obj-$(CONFIG_DRM_TTM_HELPER) += drm_ttm_helper.o
drivers/gpu/drm/amd/amdgpu/Kconfig:18:  select DRM_TTM_HELPER
drivers/gpu/drm/hisilicon/hibmc/Kconfig:11:     select DRM_TTM_HELPER
drivers/gpu/drm/loongson/Kconfig:10:    select DRM_TTM_HELPER
drivers/gpu/drm/nouveau/Kconfig:14:     select DRM_TTM_HELPER
drivers/gpu/drm/qxl/Kconfig:8:  select DRM_TTM_HELPER
drivers/gpu/drm/radeon/Kconfig:14:      select DRM_TTM_HELPER
drivers/gpu/drm/vboxvideo/Kconfig:9:    select DRM_TTM_HELPER
drivers/gpu/drm/vmwgfx/Kconfig:8:       select DRM_TTM_HELPER
drivers/gpu/drm/xe/Kconfig:40:  select DRM_TTM_HELPER
```

chewi:
I haven't tried it, but I think one way to force it is to patch the kernel like this:

```
--- a/drivers/gpu/drm/Kconfig
+++ b/drivers/gpu/drm/Kconfig
@@ -300,7 +300,7 @@ config DRM_VRAM_HELPER
          Helpers for VRAM memory management
 
 config DRM_TTM_HELPER
-       tristate
+       tristate "I REALLY WANT THIS ON"
        depends on DRM
        select DRM_TTM
        select DRM_KMS_HELPER if DRM_FBDEV_EMULATION
```

I'd be up for that. This is a relatively low traffic part of the code.

chewi:
I haven't tried this either, but Copilot tells me that it should work if we do this _after_ our call to `make olddefconfig`.

```
./scripts/config --enable CONFIG_DRM_TTM_HELPER
```

pothos:
The kernel image is 267 KiB larger now compared to the last Alpha, which I find acceptable.

Patching is a fun idea - I still think that a plain config based way is a bit nicer even if the patch could be easily rebased. Less differences compared to other distro's kernel configs also helps against hitting weird corner cases (like the one we are fixing here).

chewi:
> I haven't tried this either, but Copilot tells me that it should work if we do this _after_ our call to `make olddefconfig`... and possibly `make savedefconfig`.
> 
> ```
> ./scripts/config --enable CONFIG_DRM_TTM_HELPER
> ```

Unfortunately, this doesn't work. The build process calls `make syncconfig`, which resets it.

> The kernel image is 267 KiB larger now compared to the last Alpha, which I find acceptable.

I don't think we can afford that. We're down to our last 850KB per kernel image. It may not look like this in CI, but I've been tracking this over time across different releases. If you originally installed 3975.2 and upgrade 4368.0.0 then 4372.0.1, you would have about 1700KB left in /boot.

chewi (approved):

PR #3184, opened 2025-08-13
sys-fs/overlaybd: Actually pin TCMU this time

chewi:
I accidentally forgot to stage this change when committing the fix in #3183. 🤦‍♂️


ader1990 (approved):
LGTM - as long as the patch applies :).

PR #3187, opened 2025-08-14
Upgrade mantle container image to latest HEAD in main

flatcar-infra:
Automated changes by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action


chewi:
@ader1990 I can't see any sign that your new test ran, but I think it should have?

chewi:
Thanks. I was going to add that, but `cl.update.payload` is mentioned 3 times in that file, so I wonder if more is needed. I'm struggling to understand how this works though.

PR #3210, opened 2025-08-21
pkg-auto: Parallelize generating SDK reports and package updates handling.

krnowak:
The automation generates reports using emerge in two separate SDK containers - one with old packages and one with new packages. Both jobs create their reports in separate directories, so if we take care of printing messages to the terminal without producing garbled text and with being able to discern which job produced the terminal output, then there is nothing else that would prevent to run them in parallel.

Parallelizing handling of package updates was a bit more involved:

- We don't want to spawn 400+ processes, so each one processes one package. But rather spawn a small number of processes and tell them to process packages in batches.
- These jobs were writing information into the same file (like summary and changelog stubs), so we give each job its own directory to write to, and after all packages were processed, we merge the files into one.

In order to get the last point, I needed to refactor the some of the code to take an output directory path instead of hardcoding it to some subdirectory of `${REPORTS_DIR}` and to split off the code that handles the package update as this code would be running inside the job instead of the main process.

I think it's best to review the PR commit by commit.

### DIFF
--- a/pkg_auto/impl/inside_sdk_container_lib.sh
+++ b/pkg_auto/impl/inside_sdk_container_lib.sh
@@ -434,20 +434,6 @@ function clean_empty_warning_files() {
     done
 }
 
-function get_num_proc() {
-    local -n num_proc_ref=${1}; shift
-    local -i num_proc=1 rv=1
-
-    # stolen from portage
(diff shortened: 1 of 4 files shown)


krnowak:
Rebased, mostly for adding DCO to commits.

PR #3231, opened 2025-08-29
Change oem-hyperv to use OEM sysext

krnowak:
CI: http://localhost:8080/job/container/job/packages_all_arches/6552/cldsv/

Besides the conversion to sysext, this also fixes the hyperv images, so that they actually contain the hv daemons after the oem sysext is mounted. Previously, the `/oem` partition only had the standard `oem-release` and `grub.cfg` files. The old image didn't contain hv daemons because the app-emulation/hv-daemons package is not installing anything into `/oem`, but rather into `/usr`, which was discarded at a build time.

This PR also makes it easier to enable selinux in future. Earlier, trying to pull app-emulation/hv-daemons resulted in pulling 6 packages. With selinux, the number balooned to 130+, mostly because glibc pulls in some extra stuff. After converting to sysext, only one package is pulled.

I have been running no tests - the hyperv image isn't used in CI at all.

CC: @ader1990

### DIFF
--- a/changelog/changes/2025-08-29-hyperv-sysext.md
+++ b/changelog/changes/2025-08-29-hyperv-sysext.md
@@ -1,0 +1,1 @@
+- Hyper-V images now use a systemd-sysext image for layering additional platform-specific software on top of `/usr`


chewi (approved):
I don't really understand this (common-oem-files calls `equery which` in `src_compile`, whaaaaaat!?), but it looks far simpler than you made it sound. :sweat_smile:

PR #3241, opened 2025-09-03
test: initrd in /usr

pothos:
Still WIP

## Testing done

### DIFF
--- a/sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.44.ebuild
+++ b/sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.44.ebuild
@@ -89,6 +89,14 @@ src_compile() {
 
 	tc-export PKG_CONFIG
 	"${ESYSROOT}"/usr/bin/update-bootengine -k "${KV_FULL}" -o "${S}"/build/bootengine.cpio "${BE_ARGS[@]}" || die
+	mkdir "${S}"/build/bootengine
+	cat "${S}"/build/bootengine.cpio | while cpio --no-absolute-filenames -d -m -D "${S}"/build/bootengine -i ; do :; done
+	# Lacking read permissions even for user
+	chmod u+r "${S}"/build/bootengine/etc/gshadow
+	# CPU microcode should stay in the minimal initrd
+	rm -rf "${S}"/build/bootengine/kernel "${S}"/build/bootengine/early_cpio
+	# TODO: rework /usr verity mount to reuse the one from the minimal initrd
+	mksquashfs "${S}"/build/bootengine "${S}"/build/bootengine.img -all-root -noappend -xattrs-exclude ^btrfs.
 	kmake "$(kernel_target)"
 
 	# sanity check :)
@@ -111,4 +119,7 @@ src_install() {
 	# For easy access to vdso debug symbols in gdb:
 	#   set debug-file-directory /usr/lib/debug/usr/lib/modules/${KV_FULL}/vdso/
 	kmake INSTALL_MOD_PATH="${ED}/usr/lib/debug/usr" vdso_install
+
+	insinto "/usr/lib/flatcar"
+	doins build/bootengine.img
 }


pothos:
@chewi My idea is to load the "normal" initrd as loopback mount from `/usr` and switch to it for Ignition, network drivers and so on. If you want we could try to plug your busybox experiments in on this branch. To keep things simple and avoid risking breakage I would assume that the minimal initrd still ships the CPU microcode for the kernel to load because when we load it from userland there are more things to be aware of that aren't well supported, as far as I remember. Also, the `/usr` verity mount should be done by the minimal initrd and we would bind mount it into the "normal" initrd mount to reuse it (or maybe dm-verity also has no problems doing the work twice? For performance reasons we still might want to only do it once). In the end we could even drop the wrappers if we start to include afterburn and ignition (and other wrappers) again in the initrd and remove it from `/usr` where they aren't actually needed.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.44.ebuild
```diff
@@ -89,6 +89,14 @@ src_compile() {
...
 	tc-export PKG_CONFIG
 	"${ESYSROOT}"/usr/bin/update-bootengine -k "${KV_FULL}" -o "${S}"/build/bootengine.cpio "${BE_ARGS[@]}" || die
+	mkdir "${S}"/build/bootengine
+	cat "${S}"/build/bootengine.cpio | while cpio --no-absolute-filenames -d -m -D "${S}"/build/bootengine -i ; do :; done
+	# Lacking read permissions even for user
+	chmod u+r "${S}"/build/bootengine/etc/gshadow
+	# CPU microcode should stay in the minimal initrd
+	rm -rf "${S}"/build/bootengine/kernel "${S}"/build/bootengine/early_cpio
+	# TODO: rework /usr verity mount to reuse the one from the minimal initrd
+	mksquashfs "${S}"/build/bootengine "${S}"/build/bootengine.img -all-root -noappend -xattrs-exclude ^btrfs.
```
chewi, line 99:
```suggestion
	mkdir "${S}"/build/bootengine || die
	pushd "${S}"/build/bootengine || die
	sudo lsinitrd --unpack "${S}"/build/bootengine.cpio || die
	# TODO: rework /usr verity mount to reuse the one from the minimal initrd
	sudo mksquashfs . "${S}"/build/bootengine.img -noappend -xattrs-exclude ^btrfs. || die
	popd || die
```

This is running with `RESTRICT="userpriv"`, so you can use `sudo` here. That will ensure the device nodes are created correctly. `lsinitrd --unpack` will not include the microcode.

chewi (changes requested):
Although I shared the same concern about losing functionality that we would have to reimplement, I hadn't yet identified any such functionality, so I'm not quite ready to throw out my proposal to go straight from tiny initrd to real /usr. I'd really like to know what your specific concerns are.

This is an interesting approach in any case. My own alternative would have been to mount /usr as an overlay with the initrd, deleting all the duplicate files from the initrd, but I hadn't fully thought it through.

Regarding verity, I think it only needs to be set up once. I didn't enable verity in my own experiment, but /sysroot/usr was simply a bind mount of /usr. I think that would still work with verity applied.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.44.ebuild
chewi: Also, might as well use `mkfs.erofs` now!

chewi: With bsdtar (which is in the SDK), you can even avoid extracting if `dracut` is called with `--no-compress --no-early-microcode`.

```sh
bsdtar -c @"${S}"/build/bootengine.cpio | mkfs.erofs --tar=f -zlz4hc "${S}"/build/bootengine.raw

**Comment by chewi** [comment]
Okay, but I wasn't proposing rewriting all that. Dracut puts those scripts into an initrd. I was just going to put them in /usr instead. It's more or less the same thing. It's the scripts that Dracut itself provides through its own modules that I was concerned about.

**Comment by pothos** [comment]
The question is on how these things are started because they run in a context with dependencies. Having only one set of systemd units for both the initrd and the final system doesn't work if we want to make use of systemd in the initrd - it would run all enabled units under `/usr` (unless we inject the initrd check into most of them). Creating a separate environment manually without dracut we would have more work and risk compared to moving the current initrd as a whole. When we get things working and have time left we can still try this out as optimization (preparing an `/etc` for the initrd stage with `/etc/initrd-release`, pulling in the ignition systemd units and so on, and with masking or customizing all unnecessary units through drop-ins).

File: changelog/changes/2025-09-19-minimal-initrd.md [hunk 2]
```diff
@@ -0,0 +1 @@
+- Reduced the kernel+initrd size on `/boot` to the half. Flatcar now uses a minimal first initrd only to access the `/usr` partition and switch to the full initrd that does the full system preparation as before. Since this means that the set of kernel modules available in the first initrd is reduced, please report any impacts.
```
chewi, line 1:
```suggestion
- Reduced the kernel+initrd size on `/boot` by half. Flatcar now uses a minimal first stage initrd just to access the `/usr` partition and then switches to the full initrd that does the full system preparation as before. Since this means that the set of kernel modules available in the first initrd is reduced, please report any impact.
```

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -72,6 +73,16 @@ src_prepare() {
 	use amd64 && config_update "CONFIG_EXTRA_FIRMWARE_DIR=\"${fw_dir}\""
 }
 
+copy_in() {
+  # Simple setup, assume we only have /lib64 to care about
+  cp "${ESYSROOT}"/usr/"$1" ./"$1"
```
chewi, line 78:
```suggestion
  cp "${ESYSROOT}"/usr/"$1" ./"$1" || die
```

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -89,6 +100,77 @@ src_compile() {
...
 	"${ESYSROOT}"/usr/bin/update-bootengine -k "${KV_FULL}" -o "${S}"/build/bootengine.cpio "${BE_ARGS[@]}" || die
+	# Copy full initrd over to /usr as filesystem image
+	mkdir "${S}"/build/bootengine || die
+	pushd "${S}"/build/bootengine || die
+	sudo lsinitrd --unpack "${S}"/build/bootengine.cpio || die
+	sudo mksquashfs . "${S}"/build/bootengine.img -noappend -xattrs-exclude ^btrfs. || die
+	popd || die
+	# Create minimal initrd
+	if use amd64; then
+		microcode=$(cpio -t < "${S}"/build/bootengine.cpio 2>&1 > /dev/null | cut -d " " -f 1)
```
chewi, line 111: This looks scary. `lsinitrd` does have an `--unpackearly` option!

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -89,6 +100,77 @@ src_compile() {
...
+	fi
+	cp -a "${ESYSROOT}"/usr/bin/minimal-init ./init
+	# Make it easier to debug by not relying too much on the first commands
+	ln -s /bin/busybox ./bin/sh
+	mknod ./dev/console c 5 1
+	mknod ./dev/null c 1 3
+	mknod ./dev/tty c 5 0
+	mknod ./dev/urandom c 1 9
+	mknod ./dev/random c 1 8
+	mknod ./dev/zero c 1 5
```
chewi, line 170: You shouldn't need these since devtmpfs became a thing years ago.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -89,6 +100,77 @@ src_compile() {
...
+	done
+	echo '$MODALIAS=.*	0:0 660 @/sbin/modprobe "$MODALIAS"' > ./etc/mdev.conf
+	copy_in /bin/veritysetup
+	copy_in /bin/dmsetup
+	copy_in /bin/busybox
+	# We can't use busybox's modprobe because it doesn't support the globs in module.alias, breaking module loading
+	copy_in /sbin/kmod
+	ln -s /sbin/kmod ./sbin/modprobe
+	if use arm64; then
+		ln -s ../lib64/ld-linux-aarch64.so.1 ./lib/ld-linux-aarch64.so.1
```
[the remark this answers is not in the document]
pothos: Yes, the amd64 ld-linux entry already uses `/lib64`

changelog/changes/2025-09-19-minimal-initrd.md
pothos: Thanks!

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -89,6 +100,77 @@ src_compile() {
...
+		# Only keep early cpio for microcode
+		truncate -s $((microcode*512)) "${S}"/build/bootengine.cpio || die
+		# Debug: List contents after truncation
+		cpio -t < "${S}"/build/bootengine.cpio
+	else
+		# No early cpio, drop full initrd
+		> "${S}"/build/bootengine.cpio
+	fi
+	mkdir "${S}"/build/minimal || die
+	pushd "${S}"/build/minimal || die
```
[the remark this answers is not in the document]
pothos:
```suggestion
	mkdir "${T}"/build/minimal || die
	pushd "${T}"/build/minimal || die
```

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -89,6 +100,77 @@ src_compile() {
 
 	tc-export PKG_CONFIG
 	"${ESYSROOT}"/usr/bin/update-bootengine -k "${KV_FULL}" -o "${S}"/build/bootengine.cpio "${BE_ARGS[@]}" || die
+	# Copy full initrd over to /usr as filesystem image
+	mkdir "${S}"/build/bootengine || die
+	pushd "${S}"/build/bootengine || die
```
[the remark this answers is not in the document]
pothos:
```suggestion
	mkdir "${T}"/build/bootengine || die
	pushd "${T}"/build/bootengine || die
```

sdk_container/src/third_party/portage-stable/sys-apps/busybox/Manifest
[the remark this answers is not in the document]
pothos: Ah, good, thanks

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
```diff
@@ -89,6 +101,79 @@ src_compile() {
 
 	tc-export PKG_CONFIG
 	"${ESYSROOT}"/usr/bin/update-bootengine -k "${KV_FULL}" -o "${S}"/build/bootengine.cpio "${BE_ARGS[@]}" || die
+	# Copy full initrd over to /usr as filesystem image
+	mkdir "${T}"/build/bootengine || die
```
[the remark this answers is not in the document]
pothos: Just also noticed this and stopped the build I started :upside_down_face: 
`Path to a temporary directory which may be used by the ebuild. For example: "${PORTAGE_BUILDDIR}/temp".`

But since we already prepare the bootengine.cpio in `${S}/build` we could also keep it as is and have the directory also be prepared there, or? Is the temporary path cleaned up on failure or can one inspect the contents?

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.48.ebuild
pothos: We can try that out - I didn't find that much information, for example, I'm not sure if this `early_cpio` file is also required to be there or if this is not really required by the kernel.

chewi: It's not used by the kernel. I think it's only used by Dracut's `lsinitrd`. Yes, I guess that's a little circular!

One thing to note. The main cpio can be compressed, but the early cpio must not be.

pothos: I've done the change

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.49.ebuild
```diff
@@ -89,6 +101,82 @@ src_compile() {
...
+	for MODULE in $(find ./lib/modules/"${KV_FULL}"/ -type f); do
+		MODULE=${MODULE##*/}
+		MODULE=${MODULE%.ko.xz}
+		for DEP in $(modprobe -S "${KV_FULL}" -d "${S}"/build/bootengine -D "${MODULE}" | grep "^insmod " | cut -d " " -f 2); do
+			DEP=${DEP#"${S}/build/bootengine"}
+			DEP_DIR=${DEP%/*}
+			mkdir -p ./"${DEP_DIR}"
+			cp "${S}"/build/bootengine/"${DEP}" ./"${DEP}"
+		done
+	done
```
chewi, line 157: dracut-install can handle modules too, including their dependencies and their firmware! How about this?

```bash
pushd "${S}/build/bootengine/usr/lib/modules/${KV_FULL}" || die
find kernel/drivers/{ata,block,hid,hv,input/serio,mmc,nvme,pci,scsi,usb} kernel/fs/{btrfs,overlayfs,squashfs} kernel/security/keys -name "*.ko.*" -printf "%f\0" | xargs --null "${BROOT}"/usr/lib/dracut/dracut-install --destrootdir "${S}"/build/minimal --kerneldir . --module dm-verity dm-mod virtio_console
popd || die
```

chewi (changes requested):
Many lines are missing `|| die`. More death threats please! :grin:

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.49.ebuild
chewi: Probably want `--firmwaredirs "${ESYSROOT}"/lib/firmware` in there too so that it doesn't take that from the SDK.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.49.ebuild
```diff
@@ -89,6 +101,82 @@ src_compile() {
...
+	echo '$MODALIAS=.*	0:0 660 @/sbin/modprobe "$MODALIAS"' > ./etc/mdev.conf
+	copy_in /bin/veritysetup
+	copy_in /bin/dmsetup
+	copy_in /bin/busybox
+	# We can't use busybox's modprobe because it doesn't support the globs in module.alias, breaking module loading
+	copy_in /sbin/kmod
+	ln -s /sbin/kmod ./sbin/modprobe
+	if use arm64; then
+		ln -s ../lib64/ld-linux-aarch64.so.1 ./lib/ld-linux-aarch64.so.1
+	fi
```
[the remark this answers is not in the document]
pothos: Good idea, we can try this out and remove code here.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-kernel/coreos-kernel-6.12.49.ebuild
pothos: That helps, I need one more workaround for this broken `--firmwaredirs` flag and then have to test again…

chewi: If you find something is broken, let me know. I wrote half of dracut-install, but hopefully not the half you're having trouble with. :sweat_smile: Obviously, it doesn't normally get invoked like this.

chewi: Okay, it _does_ respect `--sysrootdir`, it was just deeper in the logic than I had looked. Passing this should do it:

```
--sysrootdir "${ESYSROOT}" --firmwaredirs "${ESYSROOT}"/lib/firmware
```

I don't know why you chose `"${S}"/build/bootengine/usr/lib/firmware` though?

chewi: I've tested this out locally to save you waiting to find out. It works. :smile:

pothos:
Ok, dropped the sudo calls.

Yes, good question - I assume that the kernel build system knows whether the first cpio can be compressed or not. We could check with real hardware if we get the microcode update applied or not (with any Flatcar release as we didn't change this).

chewi:
Yes, probably best to check that the microcode actually works, not merely whether we've changed anything. The microcode was actually missing entirely until I fixed that a few months back! See #2837.

PR #3278, opened 2025-09-17
Flatcar nano: Minor README improvements

rata:
# Flatcar NANO README improvements

This adds the missing step of starting the SDK container and using parallelism for one flatcar nano specific emerge

## How to use

Follow the instructions in the readme, they should be working now :)


t-lo (approved):
Good stuff, thank you @rata !

PR #3288, opened 2025-09-18
Flatcar nano: remove ~14 more packages

rata:
# Remove some more packages in flatcar nano hackathon

This makes nano now ~1/3 smaller than a regular flatcar release.
I think we can try to reduce this further, though.

## How to use

Compile following the readme in the nano folder.

## Testing done

Doom in wasm as a container works fine. For a hackathon, this is all the testing done :-D


t-lo (approved):
Tested on qemu and on Azure, still works. Thank you @rata !

PR #3313, opened 2025-09-29
sys-apps/systemd: Bump systemd v257.7 in base image

sayanchowdhury:
CI http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/6691/cldsv/


chewi (approved):
This is fine although 257.9 is stable in Gentoo now.

sayanchowdhury:
> This is fine although 257.9 is stable in Gentoo now.

Thanks for pointing out. I'll update and create a new PR. I had checked v257.9 and it was still unstable then - probably this was before they stablized.

PR #3321, opened 2025-10-01
Monthly GLSA metadata 2025-10-01

flatcar-infra:
Updated GLSA metadata


chewi (approved):

chewi:
Sorry, I thought it was harmless. :cry: I think PRs like this need a note about what we should check or watch out for when merging.

krnowak:
Let's go with 1. I'm working on the PAM update.

PR #3323, opened 2025-10-02
ci-automation/release.sh: Fix upload to R2

pothos:
The upload to R2 was added experimentally and we now want to make use of it. The CHANNEL variable wasn't defined and it failed because of that.

Do the upload for all channels and set the variable up first. Existing releases should get synced from the current Origin server via a FUSE mount that we anyway want to rely on for the directory listing that Caddy creates. Left to decide is how we manage the "current" version but that is done manually anyway as of now.

## Testing done

none

### DIFF
--- a/ci-automation/release.sh
+++ b/ci-automation/release.sh
@@ -164,9 +164,20 @@ function copy_from_bincache_to_bucket() {
     local arch="${2}"
     local version="${3}"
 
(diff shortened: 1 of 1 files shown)


pothos:
Backported to all channels. We don't want the release pipeline to differ unexpectedly or require manual actions when publishing releases for all channels.

PR #3606, opened 2025-12-29
app-admin/google-guest-configs: New package for udev rules and scripts

chewi:
We already have GCE disk rules in coreos-init, but a user has pointed out that the newer NVMe rules are missing. Let's take the rules directly from upstream instead. This is loosely based on the ChromiumOS package of the same name.

When bumping, we must ensure that the Dracut modules do not install files that would make runtime changes to systems to other than GCE VMs because the initrd is shared between image types. The udev disk rules are currently safe.

This also adds Google's [60-gce-network-security.conf](https://github.com/GoogleCloudPlatform/guest-configs/blob/0850c65b6d3340798bc7669a3397efe04fc7498f/src/etc/sysctl.d/60-gce-network-security.conf) sysctl file. These settings are actually generic and not even networking-specific, but even if we're not going to apply them to Flatcar in general, we should apply them to Flatcar on GCE so that it behaves like other GCE VMs. I had to renumber our baselayout.conf file to take precedence though because Google's file disabled IP forwarding, which breaks Kubernetes.

The wider GCE packages are very outdated. I started looking into this in early 2025. I then noticed this had already been attempted the year before in #1826. This change at least implements a small part of what was in that PR without touching the rest.

This is being merged in tandem with flatcar/bootengine#125, flatcar/init#140, and flatcar/baselayout#43.

## How to use

Spin up a VM with Kola using `--gce-machinetype c3-standard-4` and check whether the "google" symlink exists under /dev/disk.

You can also take this further by adding an extra disk and trying to provision it by-id with Ignition. This is awkward to pull off though because Kola doesn't let you add an extra disk, so you need to stop it tearing down the VM, add a disk manually with `gcloud`, and then use `flatcar-reset`.

It's easy to check whether Google's sysctl settings have taken effect on GCE (and not elsewhere). One such setting is `kernel.randomize_va_space = 2`.

## Testing done

This [Jenkins run](https://jenkins.flatcar.org/job/container/job/packages_all_arches/43/cldsv/) using GCE has passed. I've also done a lot of manual testing as above.

```
core@kola-1bbaa453890f254e976c ~ $ ls -l /dev/disk/by-id/google*            
lrwxrwxrwx. 1 root root 13 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd -> ../../nvme0n1
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part1 -> ../../nvme0n1p1
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part2 -> ../../nvme0n1p2
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part3 -> ../../nvme0n1p3
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part4 -> ../../nvme0n1p4
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part6 -> ../../nvme0n1p6
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part7 -> ../../nvme0n1p7
lrwxrwxrwx. 1 root root 15 Dec 29 22:26 /dev/disk/by-id/google-nvme_card-pd-part9 -> ../../nvme0n1p9
```


changelog/bugfixes/2025-12-29-gce-udev.md
krnowak, line 1: The date could be bumped, this one is soo last-year.

sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/oem-gce-20260504.ebuild
krnowak, line 1: I see you are changing the versioning of the package. This makes sense, since the package now is not tied only with `app-emulation/google-compute-engine`, but I'm not sure if using a current date is a way to go - would making it `1.0` make more sense?

krnowak (commented):
This seems to be missing your baselayout PR (https://github.com/flatcar/baselayout/pull/43) or is the PR going to be pulled as a separate scripts PR?

sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/oem-gce-20260504.ebuild
chewi: I can't remember why I picked 20260102. It was probably the version of google-guest-configs at the time, and I must have decided to use the later of the two. I still think that makes sense. Better than having a downgrade to some meaningless 1.0.

chewi: I've now bumped that and added a note to the ebuild.

chewi:
> This seems to be missing your baselayout PR...

As discussed, this was still in draft.

sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/oem-gce-20260504.ebuild
```diff
@@ -3,6 +3,8 @@
 # Copyright (c) 2020 Kinvolk GmbH. All rights reserved.
 # Distributed under the terms of the GNU General Public License v2
 
+# Version this package using the greater version of its date-based dependencies.
```
krnowak, line 6: I'm sure that we will eventually update either app-admin/google-guest-configs or app-emulation/google-compute-engine and will not bump a version of this ebuild, because it will pick up the updated deps automatically. But I also don't have any better idea here. Unless you specify dependencies as something like:
```
<=app-admin/google-guest-configs-${PV}
<=app-emulation/google-compute-engine-${PV}
```

Although not sure if this will work if for some reason we bump app-admin/google-guest-configs from 20260504 to 20260504-r1.

sdk_container/src/third_party/coreos-overlay/sys-kernel/bootengine/bootengine-9999.ebuild
```diff
@@ -10,7 +10,7 @@ if [[ ${PV} == 9999 ]]; then
 	EGIT_REPO_URI="https://github.com/flatcar/bootengine.git"
 	inherit git-r3
 else
-	EGIT_VERSION="a19dace6ad74359e6c3677fccadc36a7ce09767a" # flatcar-master
+	EGIT_VERSION="a3b96e3cfd7b8c3e0ff38540831fb0de98487404"
```
krnowak:
```suggestion
	EGIT_VERSION="a3b96e3cfd7b8c3e0ff38540831fb0de98487404" # flatcar-master
```

krnowak (approved):
I have some note about versioning of the coreos-base/oem-gce package, but I'm approving it anyway.

sdk_container/src/third_party/coreos-overlay/coreos-base/oem-gce/oem-gce-20260504.ebuild
chewi: I like the idea, but yeah, that doesn't work unfortunately.

PR #3678, opened 2026-01-30
Support building the SDK for arm64

chewi:
# Support building the SDK for arm64

Catalyst 4 adds support for building with QEMU, so I initially leveraged this, but it turned out to be very much slower than emulating the amd64 SDK on arm64, where an arm64 build could then be mostly run without emulation. I have kept the code for the slower approach anyway since it is small and may be useful to somebody.

There were several places where we assumed that amd64 was native and arm64 required emulation via QEMU. The scripts are now more architecture-agnostic, paving the way for riscv support later.

We no longer set `QEMU_LD_PREFIX` because it prevents the SDK itself from being emulated. It also assumes there is only one non-native target, which may not always be the case. bubblewrap does a better job of running binaries under QEMU. I've put a lot of work into catching all the instances where we were implicitly relying on QEMU when building packages. The new upstream sysroot.eclass has addressed some of the more stubborn cases.

Note that this doesn't address the issue of how we actually publish a supported arm64 SDK on an ongoing basis as that requires more work. It would make sense to have a multi-arch image, but that _might_ require building them in tandem. We wouldn't want a build of one to break the other. The naming of the SDK is also ambiguous in places.

## How to use

* Start with Docker on some arm64 distro (e.g. Ubuntu 24.04) running natively.
* Download the [SDK .tar.zst tarball](https://bincache.flatcar-linux.net/containers/9999.0.0-arm64-sdk-3/flatcar-sdk-all-9999.0.0-arm64-sdk-3.tar.zst) from my Jenkins run.
* Download the latest [Gentoo stage3 arm64 systemd tarball](https://www.gentoo.org/downloads/#arm64-advanced).
* Add `--exclude dev-lang/perl` to the list of excludes in build_library/catalyst.sh. This is a workaround for Gentoo's stage3 being unstable. Maybe we could add this permanently.

```
docker run --privileged --rm tonistiigi/binfmt --install amd64
zstdcat flatcar-sdk-all-*.tar.zst | docker load
./run_sdk_container -t -U -n amd64-sdk -C ghcr.io/flatcar/flatcar-sdk-all:*
sudo env ARCH=arm64 ./bootstrap_sdk --seed_tarball stage3-arm64-systemd-*.tar.xz --profile coreos-overlay:coreos/arm64/sdk
```

At this point, my amd64-sdk container had seemingly corrupted itself because it refused to execute amd64 binaries. I don't recall this happening on my earlier attempts, but recreating the container works. The next step is done outside the container anyway.

```
./build_sdk_container_image __build__/images/catalyst/builds/flatcar-sdk/flatcar-sdk-arm64-*.tar.bz2
./run_sdk_container -t -U -n arm64-sdk -C ghcr.io/flatcar/flatcar-sdk-all:*

./build_packages --board arm64-usr --nousepkg
./build_image --board=arm64-usr prod sysext oem_sysext

./build_packages --board amd64-usr --nousepkg
./build_image --board=amd64-usr prod sysext oem_sysext
```

By this point, you will have successfully built arm64 and amd64 images from the arm64 SDK. Note that building the arm64 SDK in a single phase like this leaves you with a slightly unclean result. Some of the board root packages will have been built with the amd64 SDK and this will be recorded in each package's metadata. This doesn't cause an issue (now that I have fixed sysroot.eclass) but for a truly clean result, do a two-phase SDK build.

## Testing done

This [two-phase SDK build in Jenkins](https://jenkins.flatcar.org/job/container/job/sdk/21/cldsv/) was entirely successful.

I have also built an arm64 SDK and used it to build arm64 and amd64 images as above.

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update) -- N/A


ader1990 (approved):
LGTM.

chewi:
Okay, I think this is finally ready to go now. I needed to bump to the unstable docker-cli package, but I think that's fine because it's overdue a bump anyway. I haven't tried an arm64 SDK build using my latest Jenkins run, but I have no reason to believe it would be broken now. The important thing is that it doesn't break amd64 SDK builds. We can fix any fresh arm64 SDK issues later.

ader1990:
Yay! Thanks for the awesome work!

PR #3752, opened 2026-02-23
build_sysext: Squash build Portage profile parsing failure warning

chewi:
# Squash build Portage profile parsing failure warning

The Portage config is visible through the sysext root directory via overlayfs while installing packages, but the overlay gets unmounted before `equery` is called. Use the board root's config instead.

## How to use

Check the `image` job log and the *_packages.txt files that get uploaded to bincache.

## Testing done

[This Jenkins run](http://jenkins.infra.kinvolk.io:8080/job/container/job/packages_all_arches/7457/cldsv/) passed. There were no signs of the warning, and the *_packages.txt files looked normal.

### DIFF
--- a/build_sysext
+++ b/build_sysext
@@ -248,7 +248,7 @@ if [[ "$FLAGS_generate_pkginfo" = "${FLAGS_TRUE}" ]] ; then
 fi
 
 info "Writing ${SYSEXTNAME}_packages.txt"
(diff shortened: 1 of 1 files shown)


krnowak (approved):
I was tempted to tell you to move the code, so that package generation happens when `${FLAGS_install_root_basename}` still has a valid portage config, but eh - it's not like profile config here has any importance at all here, since `equery` seemed to do just fine without it.

PR #3783, opened 2026-03-05
common.sh: redirect verbose debug output to file

t-lo:
common.sh produces *very* verbose debug output on failure. This leads to the actual error being masked as it is pushed out of the scrollback buffer by many pages of dmesg.

This change redirects this output to a FAILURE log file which is timestapmed.


krnowak:
Opened https://github.com/flatcar/scripts/pull/3984.

PR #3933, opened 2026-04-15
sbsign: update Azure key vault location

tormath1:
Everything is in the title.

---

As suggested by @danzatt we can move this to a CI parameter, but as we are releasing I don't want to introduce more bugs than we currently have.


chewi (approved):

tormath1:
Cherry-picked to:
* flatcar-4669
* flatcar-4628
* flatcar-4593
* flatcar-4459

PR #3954, opened 2026-04-21
Parameterize secure boot keys

jqueuniet:
# Parameterize Secure Boot keys

The current Secure Boot pipeline uses two different set of keys, one is public, hard-coded in the SDK container filesystem, and intended for test builds, the other is pulled from an Azure KMS and intended for release builds. Neither can be meaningfully parameterized to sign builds with custom keys.

This merge request attempts to parameterize the test keys to allow using custom keys instead. I tried to minimize actual changes to the codebase and as such opted for environment variables, at least to validate the initial approach. I feel they're a bit too "magic" and obscure though, and CLI arguments could be added to the necessary scripts if preferred.

Also, the generation, env preparation and artefact verification could be offloaded to proper scripts (ie, `verify_sbkey_images`, `generate_sbkeys` and `set_sbkeys_env`) to ease custom key workflow as well as act as documentation for the process.

## How to use

Generate keys:

```sh
OUTPUT_DIR="./sb-keys"
SUBJ_PREFIX="Flatcar Derivative"
DAYS=7300

openssl req -new -x509 -newkey rsa:2048 -sha256 -days "${DAYS}" \
    -subj "/CN=${SUBJ_PREFIX} Platform Key/" -nodes \
    -keyout "${OUTPUT_DIR}/PK.key" -out "${OUTPUT_DIR}/PK.crt"

openssl req -new -x509 -newkey rsa:2048 -sha256 -days "${DAYS}" \
    -subj "/CN=${SUBJ_PREFIX} Key Exchange Key/" -nodes \
    -keyout "${OUTPUT_DIR}/KEK.key" -out "${OUTPUT_DIR}/KEK.crt"

openssl req -new -x509 -newkey rsa:2048 -sha256 -days "${DAYS}" \
    -subj "/CN=${SUBJ_PREFIX} Secure Boot DB/" -nodes \
    -keyout "${OUTPUT_DIR}/DB.key" -out "${OUTPUT_DIR}/DB.crt"

openssl genrsa -out "${OUTPUT_DIR}/shim.key" 2048
openssl req -new -x509 -sha256 -days "${DAYS}" \
    -subj "/CN=${SUBJ_PREFIX} Shim Key/" \
    -key "${OUTPUT_DIR}/shim.key" -out "${OUTPUT_DIR}/shim.pem"
openssl x509 -in "${OUTPUT_DIR}/shim.pem" -inform PEM \
    -out "${OUTPUT_DIR}/shim.der" -outform DER
```

Set a proper environment:

```sh
CONTAINER_KEYS_DIR="/home/sdk/trunk/src/scripts/sb-keys"

export SBSIGN_KEY="${CONTAINER_KEYS_DIR}/shim.key"
export SBSIGN_CERT="${CONTAINER_KEYS_DIR}/shim.pem"
export SBSIGN_DB_KEY="${CONTAINER_KEYS_DIR}/DB.key"
export SBSIGN_DB_CERT="${CONTAINER_KEYS_DIR}/DB.crt"
export SHIM_SIGNING_CERTIFICATE="${CONTAINER_KEYS_DIR}/shim.der"
```

Build Flatcar as usual:

```sh
./build_packages
./build_image ...
./image_to_vm.sh ...
```

## Testing done

Verify artefacts have been signed with the correct keys:

```sh
sbverify __build__/images/images/amd64-usr/latest/flatcar_production_image.vmlinuz --cert sb-keys/shim.pem
Signature verification OK
```

### DIFF
--- a/sdk_lib/sdk_entry.sh
+++ b/sdk_lib/sdk_entry.sh
@@ -72,10 +72,14 @@ fi
 
 # Create key directory if not already configured in .bashrc
 if ! grep -q 'export MODULE_SIGNING_KEY_DIR=' /home/sdk/.bashrc; then
-    # For official builds, use ephemeral keys. For unofficial builds, use persistent directory
-    if [[ ${COREOS_OFFICIAL:-0} -eq 1 ]]; then
+    if [[ -n ${MODULE_SIGNING_KEY_DIR:-} ]]; then
+        # Pre-set via environment (e.g. .sdkenv) — use as-is
+        :
+    elif [[ ${COREOS_OFFICIAL:-0} -eq 1 ]]; then
+        # For official builds, use ephemeral keys
         MODULE_SIGNING_KEY_DIR=$(su sdk -c "mktemp -d")
     else
+        # For unofficial builds, use persistent directory
         MODULE_SIGNING_KEY_DIR="/home/sdk/.module-signing-keys"
         su sdk -c "mkdir -p ${MODULE_SIGNING_KEY_DIR@Q}"
     fi
@@ -97,7 +101,10 @@ if grep -q 'export SYSEXT_SIGNING_KEY_DIR' /home/sdk/.bashrc; then
     fi
 fi
 grep -q 'export SYSEXT_SIGNING_KEY_DIR' /home/sdk/.bashrc || {
-    if [[ ${COREOS_OFFICIAL:-0} -eq 1 ]]; then
+    if [[ -n ${SYSEXT_SIGNING_KEY_DIR:-} ]]; then
+        # Pre-set via environment (e.g. .sdkenv) — use as-is
+        :
+    elif [[ ${COREOS_OFFICIAL:-0} -eq 1 ]]; then
         SYSEXT_SIGNING_KEY_DIR=$(su sdk -c "mktemp -d")
     else
         SYSEXT_SIGNING_KEY_DIR="/home/sdk/.sysext-signing-keys"
--- a/build_library/grub_install.sh
+++ b/build_library/grub_install.sh
@@ -37,6 +37,9 @@ switch_to_strict_mode
 . "${BUILD_LIBRARY_DIR}/board_options.sh" || exit 1
 . "${BUILD_LIBRARY_DIR}/sbsign_util.sh" || exit 1
 
+SBSIGN_DB_KEY="${SBSIGN_DB_KEY:-/usr/share/sb_keys/DB.key}"
+SBSIGN_DB_CERT="${SBSIGN_DB_CERT:-/usr/share/sb_keys/DB.crt}"
+
 # Our GRUB lives under flatcar/grub so new pygrub versions cannot find grub.cfg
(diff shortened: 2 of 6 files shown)


sdk_lib/sdk_entry.sh
```diff
@@ -72,10 +72,14 @@ fi
 
 # Create key directory if not already configured in .bashrc
 if ! grep -q 'export MODULE_SIGNING_KEY_DIR=' /home/sdk/.bashrc; then
-    # For official builds, use ephemeral keys. For unofficial builds, use persistent directory
-    if [[ ${COREOS_OFFICIAL:-0} -eq 1 ]]; then
+    if [[ -n "${MODULE_SIGNING_KEY_DIR:-}" && -d "${MODULE_SIGNING_KEY_DIR}" ]]; then
```
chewi: I would drop the existence check and let it fail below. You don't want to specify `MODULE_SIGNING_KEY_DIR` and do a whole build, only to realise it used the wrong key at the end. Same below.

```suggestion
    if [[ -n ${MODULE_SIGNING_KEY_DIR:-} ]]; then
```

chewi (changes requested):
Thank you very much for this! Apart from the one suggestion and some overquoting (no quotes needed in `[[ ]]`), this looks fine. I think using environment variables is perfectly valid here, otherwise we'll need to add and remember to use arguments in several places. I ran this under GHA, which had some unrelated issues, so I tried it locally, including Kola's `cl.basic` test, and it just worked. If you can make that change, I'd be happy to merge without doing another build.

For us, we need to do things differently for official vs unofficial builds because our sources are public, and we obviously cannot publish the official private key. You won't have this problem, but I wondered whether you were satisfied with keeping the private key in a file, as opposed to using a safer PKCS11-based mechanism.

jqueuniet:
> Thank you very much for this! Apart from the one suggestion and some overquoting (no quotes needed in `[[ ]]`), this looks fine.

I see, I'm used to `[ ]` and did not know that `[[ ]]` did not require such agressive quoting. So it's more of a style/consistency issue.

> I think using environment variables is perfectly valid here, otherwise we'll need to add and remember to use arguments in several places. I ran this under GHA, which had some unrelated issues, so I tried it locally, including Kola's `cl.basic` test, and it just worked. If you can make that change, I'd be happy to merge without doing another build.

Sure, I made the change

> For us, we need to do things differently for official vs unofficial builds because our sources are public, and we obviously cannot publish the official private key. You won't have this problem, but I wondered whether you were satisfied with keeping the private key in a file, as opposed to using a safer PKCS11-based mechanism.

For now we're operating from my Github fork so we're subject to the same constraints, and I would certainly not commit such keys to a git repository, even a private one. You're right that we will definitely not be satisfied by files in the long run, government regulation pressure will very likely steer us toward PKCS11 or similar mechanisms eventually. But I'm not sure yet how that will take place, it could be a remote KMS, a standard smartcard, a full-fledged HSM, etc. The only thing I'm pretty sure of is that it will not be cloud-based. So I'd rather look at this through a separate PR at a later time, once the picture is clear.

chewi (approved):

PR #3957, opened 2026-04-22
common: Filter dmesg on failure to warnings and errors

zhijianli88:
When a script fails, the die_notrace function dumps the entire dmesg buffer for debugging purposes. On a busy system, this log can be excessively long, making it difficult to locate the actual kernel-level warnings or errors that may have caused the failure. The important messages are often buried in a flood of unrelated informational output.

Modify the dmesg call in the error handler to make debugging easier. Use the --level=err,warn option to limit the output to only warning and error messages, which significantly reduces noise and focuses on potential problems.

In addition, add the --ctime option to print human-readable timestamps for better event correlation and --color=never to ensure clean, parseable output in log files. The debug section header is also updated to reflect this change.

### DIFF
--- a/common.sh
+++ b/common.sh
@@ -177,9 +177,9 @@ die_notrace() {
       error "${DIE_PREFIX}========"
       error_command_output "${DIE_PREFIX}" df -h
       error
-      error "${DIE_PREFIX}== DMESG =="
-      error "${DIE_PREFIX}==========="
-      error_command_output "${DIE_PREFIX}" sudo dmesg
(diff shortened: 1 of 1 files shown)


chewi:
@krnowak, @t-lo, you were talking about this recently, so I'll leave this with you.

krnowak:
I have a branch that drops this stuff altogether, among other things - it turned out to be never of any use. I'll file a PR eventually.

krnowak:
Opened https://github.com/flatcar/scripts/pull/3984.

PR #3958, opened 2026-04-22
common: check binfmt_misc for cross builds

zhijianli88:
Fail early when cross builds lack the required binfmt_misc entry

in build_packages and build_image, matching SDK guidance.

Ref: https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-modifying-flatcar/#select-the-target-architecture

## How to use

## Testing done


chewi:
Thanks, but this won't actually be needed once #3678 is merged, which should happen this week. We're no longer going to rely on the implicit execution of QEMU. I have made improvements to Gentoo to explicitly execute it where necessary.

PR #3960, opened 2026-04-23
Upgrade Linux Kernel for flatcar-4628 from 6.12.81 to 6.12.83

flatcar-infra:
Subject says it all.


dongsupark:
See instead https://github.com/flatcar/scripts/pull/3977

PR #3964, opened 2026-04-24
Improve SDK container setup messages for new users

John15321:
## Improve SDK container setup messages for new users

When running run_sdk_container for the first time, several messages can confuse new users into thinking something is broken when the process is actually working correctly.

### Changes

- **Clarify gcloud credentials mount** - Added a message explaining that the mounted gcloud credentials are for optional artifact uploads and are not baked into any built image.
- **Improve registry fallback message** - Replaced `"Falling back to tar ball download..."` with a message explaining this is normal for nightly builds.
- **Show download progress interactively** - SDK tarballs are several GB but `curl` ran silently with no progress indication. Now shows a progress bar when a TTY is attached, stays silent in CI.

### Context

Multiple contributors were confused by all three of these when building Flatcar from `main` for the first time.

## Testing done

### DIFF
--- a/sdk_lib/sdk_container_common.sh
+++ b/sdk_lib/sdk_container_common.sh
@@ -303,6 +303,7 @@ function gnupg_ssh_gcloud_mount_opts() {
     if [[ -e ${GOOGLE_APPLICATION_CREDENTIALS:-} ]] ; then
         creds_dir=$(dirname "${GOOGLE_APPLICATION_CREDENTIALS}")
         if [[ -d ${creds_dir} ]] ; then
+            echo "Mounting gcloud credentials from ${creds_dir} (used for artifact uploads, safe to ignore if not needed, not baked into any image)"
             echo "-v $creds_dir:$creds_dir"
             args_ref+=( -v "${creds_dir}:${creds_dir}" )
         fi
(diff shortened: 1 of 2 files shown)


John15321:
Example of how it would look like now:

image

sdk_lib/sdk_container_common.sh
```diff
@@ -303,6 +303,7 @@ function gnupg_ssh_gcloud_mount_opts() {
     if [[ -e ${GOOGLE_APPLICATION_CREDENTIALS:-} ]] ; then
         creds_dir=$(dirname "${GOOGLE_APPLICATION_CREDENTIALS}")
         if [[ -d ${creds_dir} ]] ; then
+            echo "Mounting gcloud credentials from ${creds_dir} (used for artifact uploads, safe to ignore if not needed, not baked into any image)"
```
chewi, line 306: I'm going to kill this very soon anyway.

John15321: Yeah but for the time being maybe lets have it and then you remove it?

chewi: Hopefully will be in the next couple of weeks but okay.

chewi (approved):

PR #3983, opened 2026-04-28
build_image: Allow specifying which OEM sysexts to build

krnowak:
Useful if we want to locally generate one kind of image for testing.

Tested locally for building qemu OEM sysext only.

### DIFF
--- a/build_library/prod_image_util.sh
+++ b/build_library/prod_image_util.sh
@@ -276,13 +276,26 @@ create_prod_sysexts() {
 }
 
 create_oem_sysexts() {
-  local image_name="$1"
+  local image_name=${1}; shift
+  local requested_oem_sysexts_csv=${1}; shift
   local image_sysext_base="${image_name%.bin}_sysext.squashfs"
   local overlay_path
   overlay_path=$(portageq get_repo_path / coreos-overlay)
 
   local -a oem_sysexts
   get_oem_sysext_matrix "${ARCH}" oem_sysexts
+  if [[ ${requested_oem_sysexts_csv} != 'everything!' ]]; then
+      local -a all_oems requested_oems invalid_oems
+      all_oems=( "${oem_sysexts[@]}" )
+      all_oems=( "${all_oems[@]%%|*}" )
+      all_oems=( "${all_oems[@]#oem-}" )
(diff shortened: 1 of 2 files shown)


chewi (changes requested):
Handy, but I'm about to kill common-oem-files. This will still work, but it will check for the presence of `coreos-base/oem-*/*.ebuild` instead.

krnowak:
> Handy, but I'm about to kill common-oem-files. This will still work, but it will check for the presence of `coreos-base/oem-*/*.ebuild` instead.

Sure, whatever you do, it shouldn't affect my additions as long as `get_oem_sysext_matrix` works the same. I'll reword the help string to be less specific, though.

build_library/prod_image_util.sh
```diff
@@ -276,13 +276,24 @@ create_prod_sysexts() {
...
   get_oem_sysext_matrix "${ARCH}" oem_sysexts
+  if [[ ${requested_oem_sysexts_csv} != 'everything!' ]]; then
+      local -a all_oems requested_oems invalid_oems
+      all_oems=( "${oem_sysexts[@]##*|}" )
+      mapfile -t requested_oems <<<"${requested_oem_sysexts_csv//,/$'\n'}"
+      mapfile -t invalid_oems < <(comm -23 <(printf '%s\n' "${requested_oems[@]}" | sort -u) <(printf '%s\n' "${all_oems[@]}" | sort -u))
+      if [[ ${#invalid_oems[@]} -gt 0 ]]; then
+          die "Requested OEMs to build sysexts for are invalid: ${invalid_oems[*]}, valid OEMs are ${all_oems[*]}"
+      fi
+      mapfile -t oem_sysexts < <(printf '%s\n' "${oem_sysexts[@]}" | grep '|\('"${requested_oem_sysexts_csv//,/'\|'}"'\)$')
```
chewi: I think this is matching the last field? That's the USE flag field rather than the name field, which is first. I'm also killing OEM sysext USE flags, so this field will be empty.

chewi (changes requested):

build_library/prod_image_util.sh
krnowak: Yeah, I used the field that had the OEM name. I'll use the first field then.

krnowak:
Updated, tested locally.

chewi (approved):

PR #3985, opened 2026-04-30
overlay profiles: Try a more elegant solution to python version pinning

krnowak:
Got a suggestion via recent Gentoo news.

CI: https://jenkins.flatcar.org/job/container/job/sdk/51/cldsv/


krnowak:
That does not seem to be working, probably only usable in package.use in /etc/portage, not in profiles.

PR #4004, opened 2026-05-07
Convert OEM packages to sysext, drop several images, revamp Vagrant support

chewi:
Brace yourselves, this is a big one! The good news is the difference is about -1600 lines. 😁

Since having the OEM sysext packages built in the "packages" job, they have been ecleaned and then built again later. This fixes that issue and ensures they are always consumed from binary packages.

This drops the per-OEM USE flags. Having a USE flag per OEM is unnecessary when we only really need to know whether the package will be installed in an OEM sysext or not. A single USE flag could do this, but our existing `flatcar_target()` helper avoids the complexity of `binpkg-multi-instance`. The flag names were also prone to conflicts (e.g. `qemu`) and we don't currently make use of them anyway.

This populates the SYSEXT_ID/NAME/VERSION_ID/HOME_URL/BUG_REPORT_URL extension-relhease.d fields in each sysext. These are equivalent to their non-SYSEXT os-release counterparts. Most of these are taken from the sysext's package is it is built from just one, which is often the case. These fields are now used to write the oem-release file later by extracting them with systemd-dissect.

As per flatcar/bootengine#124, this adds the `vboxguest` kernel module, which allows the VirtualBox image to actually boot. It broke when upstream Ignition changed how the VirtualBox support works.

Also per flatcar/bootengine#124, the Ignition patch to handle unsupported platforms like Vagrant has been dropped because it is more useful to rely on systemd-detect-virt than to simply assume "metal", which does nothing.

As with my other recent PRs, this drops support for Equinix Metal (Packet) and Rackspace. It also drops the vmware_insecure image, which seems pointless, the vagrant_virtualbox image, which is unneeded when the vagrant image already targets VirtualBox, and the vagrant_vmware_fusion image, which we haven't actually built for years.

Dropping vagrant_virtualbox also meant dropping support for that combination in Afterburn. Now that the OEM ID is just `vagrant`, we can't distinguish it from Parallels. I highly doubt anyone cares, and it brings up closer to upstream.

The main point of this whole PR is to convert the remaining OEM packages to sysexts. This involves reworking how they are built. common-oem-files was a giant hack. oem-release is now generated from the sysexts themselves as mentioned above. grub.cfg is often just a couple of lines, so there's no point in having all this complex logic around generating it, especially when that logic has some exceptions anyway. It's far easier to just explicitly store each variant under build_library/oem. The list of OEMs is now generated by looking for `coreos-base/oem-*/*.ebuild`.

This totally revamps our Vagrant support, dropping custom code that was merged upstream years ago, adding support for Ignition and cloud-config within the single VirtualBox image, and sharing the JSON metadata between the VirtualBox and Parallels images. Check flatcar/flatcar-website#567 to see what this new support looks like in practise.

The VirtualBox OVF, which is also used by the Vagrant image, has been modernised for new hardware, including a VirtIO storage controller and UEFI. VirtualBox actually uses EDK2 for UEFI.

This drop the OEM sysext dependencies from coreos-devel/board-packages because these are no longer necessary now that they explicitly built with `--onlydeps` in `build_sysext_packages()`.

This needs to be merged in conjunction with flatcar/bootengine#124, flatcar/coreos-cloudinit#38, and flatcar/jenkins-os#442.

## How to use

There's obviously a lot to unpack here. You can try VirtualBox and Vagrant with [my build](https://bincache.flatcar-linux.net/images/amd64/9999.0.0+oem-rework-3/) using the new docs in flatcar/flatcar-website#567.

## Testing done

[Jenkins](https://jenkins.flatcar.org/job/container/job/packages_all_arches/147/cldsv/) completed successfully, except for Digital Ocean, which is unrelated. There are surprisingly few resulting image differences for such a large change. I have obviously done loads of manual testing around this, especially with VirtualBox and Vagrant. Unfortunately, I don't think we have the means to test the other OEMs being migrated, namely CloudSigma, CloudStack, and Exoscale, but I think these are all community supported.


.github/workflows/ci.yaml
```diff
@@ -189,25 +189,15 @@ jobs:
...
-            formats+=" vmware vmware_insecure vmware_ova vmware_raw"
+          printf -v formats "%s\n" ${formats}
+          if grep -q '^vmware' <<< "${formats}"; then
+              formats=$(grep -v '^vmware' <<< "${formats}")
+              printf -v formats "%s\n" ${formats} vmware vmware_ova vmware_raw
           fi
-          if echo "$formats" | tr ' ' '\n' | grep -q -P '^(ami|aws)'; then
-            formats=$(echo "$formats" | tr ' ' '\n' | sed '/ami.*/d' | sed '/aws/d')
-            formats+=" ami ami_vmdk"
+          if printf "%s\n" $formats | grep -q '^ami\|^aws'; then
```
krnowak: `formats` seems to be already a string newline-separated entries, so:

```suggestion
          if grep -q '^ami\|^aws' <<<"${formats}"; then
```

changelog/changes/2026-04-27-vagrant-virtualbox.md
```diff
@@ -0,0 +1 @@
+- Dropped the Vagrant VirtualBox image because the regular Vagrant image also targets VirtualBox. The only difference between them was that the former was geared for provisioning with Ignition and Afterburn rather than cloud-config and Vagrant itself. The Ignition support was broken when it was dropped by upsteam. A single image can handle Ignition, cloud-config, and Vagrant. The Afterburn support has been dropped entirely.
```
krnowak:
```suggestion
- Dropped the Vagrant VirtualBox image because the regular Vagrant image also targets VirtualBox. The only difference between them was that the former was geared for provisioning with Ignition and Afterburn rather than cloud-config and Vagrant itself. The Ignition support was broken when it was dropped by upstream. A single image can handle Ignition, cloud-config, and Vagrant. The Afterburn support has been dropped entirely.
```

sdk_container/src/third_party/coreos-overlay/README.md
```diff
@@ -23,13 +23,12 @@ gets built into a developer image and is not OEM specific.
...
-development or production image, plus any OEM specific packages.
+development or production image.
 
 `coreos-base/oem-*` are the OEM specific packages. They mostly install things
-that belong in the OEM partition. Any RDEPENDS from these packages should
-be copied to the RDEPENDS in `board-packages` to ensure they are built.
+that belong in the OEM partition.
 
-`coreos-base/coreos-oem-*` are metapackages for OEM specific ACIs. 
+`coreos-base/coreos-oem-*` are metapackages for OEM specific ACIs.
```
krnowak: Feel free to drop this paragraph. There are not coreos-base/coreos-oem-* packages and ACI is dead.

build_library/vm_image_util.sh
```diff
@@ -586,34 +407,27 @@ install_oem_sysext() {
...
-run_fs_hook() {
-    local fs_hook=$(_get_vm_opt FS_HOOK)
-    if [[ -n "${fs_hook}" ]]; then
-        info "Running ${fs_hook} fs hook"
-        _run_${fs_hook}_fs_hook "$@"
-    fi
-}
+    # Extract metadata from the sysext and write it to oem-release. Chop off the
+    # oem- prefix from SYSEXT_ID because that's how it has traditionally looked.
+    sudo systemd-dissect --json=short "${prebuilt_sysext_path}" | jq -r '.sysextRelease | .[]' |
```
krnowak: This could also be:

```suggestion
    sudo systemd-dissect --json=short "${prebuilt_sysext_path}" | jq -r '.sysextRelease[]' |
```

build_library/vm_image_util.sh
```diff
@@ -586,34 +407,27 @@ install_oem_sysext() {
...
+        sed -n -r -e "s/^SYSEXT_ID=oem-/SYSEXT_ID=/" -e "s/^SYSEXT_//p" |
```
krnowak: Consider using long flags for `-n` and `-r` (`--quiet` and `--extended-regexp`, respectively) as they are seldom used. `-e` is well known, I'd say, so this could stay as is.

build_library/vm_image_util.sh
```diff
@@ -1209,42 +954,55 @@ write_vm_bundle() {
...
         image="${VM_TMP_DIR}/tmp.pvm"
         mkdir -p "${image}"
         mv "${VM_DST_IMG}" "${image}"
         mv "$(_dst_path ".pvs")" "${image}"/config.pvs
     fi
 
-    cat > "${VM_TMP_DIR}"/box/metadata.json <<EOF
-{"provider": "${provider}"}
-EOF
+    jq -n '{ provider: env.provider, architecture: env.arch }' > "${VM_TMP_DIR}"/box/metadata.json
```
krnowak, line 975: Neat!

ci-automation/vms.sh
```diff
@@ -79,28 +79,16 @@ function _vm_build_impl() {
...
-    done
-
-    [[ ${has_packet} -eq 1 ]] && [[ ${has_pxe} -eq 0 ]] && set -- 'pxe' "${@}"
-
     # Convert platform names (also used to find the test scripts) to image formats they entail
-    formats="$*"
-    if echo "$formats" | tr ' ' '\n' | grep -q '^vmware'; then
-      formats=$(echo "$formats" | tr ' ' '\n' | sed '/vmware.*/d')
-      formats+=" vmware vmware_insecure vmware_ova vmware_raw"
+    printf -v formats "%s\n" ${formats}
```
krnowak: Shouldn't this be:
```suggestion
    printf -v formats "%s\n" "${@}"
```
? I can't see `formats` to be defined earlier in the function.

ci-automation/vms.sh
```diff
@@ -79,28 +79,16 @@ function _vm_build_impl() {
...
+    if grep -q '^vmware' <<< "${formats}"; then
+        formats=$(grep -v '^vmware' <<< "${formats}")
+        printf -v formats "%s\n" ${formats} vmware vmware_ova vmware_raw
     fi
-    if echo "$formats" | tr ' ' '\n' | grep -q -P '^(ami|aws)'; then
-      formats=$(echo "$formats" | tr ' ' '\n' | sed '/ami.*/d' | sed '/aws/d')
-      formats+=" ami ami_vmdk"
+    if printf "%s\n" $formats | grep -q '^ami\|^aws'; then
```
krnowak:
```suggestion
    if grep -q '^ami\|^aws' <<<"${formats}"; then
```

.github/workflows/ci.yaml
```diff
@@ -189,25 +189,15 @@ jobs:
...
-          for format in "${formats}";do
-            [[ "${format}" = 'packet' ]] || [[ "${format}" = 'equinix_metal' ]] && has_packet=1
-            [[ "${format}" = 'pxe' ]] && has_pxe=1
-          done
-
-          [[ ${has_packet} -eq 1 ]] && [[ ${has_pxe} -eq 0 ]] && set -- 'pxe' "${@}"
-          if echo "$formats" | tr ' ' '\n' | grep -q '^vmware'; then
-            formats=$(echo "$formats" | tr ' ' '\n' | sed '/vmware.*/d')
-            formats+=" vmware vmware_insecure vmware_ova vmware_raw"
+          printf -v formats "%s\n" ${formats}
```
krnowak: Should this be:
```suggestion
          printf -v formats "%s\n" ${IMAGE_FORMATS}
```
? I can't see `formats` defined earlier in the script.

build_sysext
```diff
@@ -232,6 +232,13 @@ for package; do
     "${package}"
 done
 
+# If there's only one package, grab its metadata for the sysext.
+unset PVR HOMEPAGE SYSEXT_NAME
+if [[ ${#} -eq 1 ]]; then
+  CPV=$(portageq best_version "${BUILD_DIR}/${FLAGS_install_root_basename}" "$1")
+  IFS=$'\n' read -r -d '' PVR HOMEPAGE SYSEXT_NAME < <(portageq metadata "${BUILD_DIR}/${FLAGS_install_root_basename}" installed "${CPV}" PVR HOMEPAGE SYSEXT_NAME; printf '\0')
```
krnowak, line 239: Huh? Is SYSEXT_NAME saved in metadata?

build_sysext
```diff
@@ -282,15 +289,18 @@ for entry in "${BUILD_DIR}/${FLAGS_install_root_basename}"/*; do
...
-version_field="${VERSION_FIELD_OVERRIDE:-VERSION_ID=${FLATCAR_VERSION_ID}}"
-all_fields=(
-  'ID=flatcar'
-  "${version_field}"
-  "ARCHITECTURE=${ARCH}"
-  "EXTENSION_RELOAD_MANAGER=1"
-)
-printf '%s\n' "${all_fields[@]}" >"${BUILD_DIR}/${FLAGS_install_root_basename}/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
+
+grep . <<EOF | install -m0644 -D /dev/stdin "${BUILD_DIR}/${FLAGS_install_root_basename}/usr/lib/extension-release.d/extension-release.${SYSEXT_ID}"
```
krnowak, line 293: Could `cat` be used instead of `grep .`?

krnowak:
Also, what is the reason behind using a separate license for metapackages?

build_sysext
chewi: This is to avoid blank lines. I did the same in Gentoo's baselayout.

chewi:
> Also, what is the reason behind using a separate license for metapackages?

That license comes from Gentoo and it is exactly for cases like this, packages with no actual code. Using Apache-2.0 or whatever here doesn't seem right.

build_sysext
chewi: Admittedly, the `portageq` help doesn't say it works for arbitrary variables in the ebuild, but it does. I'm happy to rely on this until it breaks.

.github/workflows/ci.yaml
chewi: Oops, you spotted how I worked on this incrementally. :flushed:

build_library/vm_image_util.sh
chewi: Bah, I do `sed` in my sleep. But fine. :stuck_out_tongue:

ci-automation/vms.sh
chewi: Well spotted! The perils of copy/paste.

.github/workflows/ci.yaml
chewi: Another good spot!

krnowak (approved):

PR #4019, opened 2026-05-13
Update amd64_defconfig with new kernel options

tylerbucher:
# Added CONFIG_DRM_I915_PXP, CONFIG_INTEL_MEI_GSC, and CONFIG_INTEL_MEI_PXP to the amd64_defconfig.

This change enables hardware video transcoding support for Intel Arc (DG2) discrete GPUs by adding three missing kernel config options to the amd64 defconfig.

Intel Arc GPUs require HuC firmware authentication before the GPU's dedicated video engine (VDBox/VEBox) can be used for hardware accelerated video decode/encode. On DG2, HuC loading is performed by the GSC via a PXP command this is the only mechanism the i915 driver implements for HuC authentication on DG2. Without mei_gsc and mei_pxp present, the driver has no path to authenticate HuC, and it fails unconditionally. Without it, HuC fails to initialize and all video transcoding falls back to CPU software encoding, making hardware-accelerated media server workloads (e.g. Jellyfin, Plex, FFmpeg) non-functional on Flatcar with Intel Arc hardware.

(all statements and information are mine, AI assisted with grammar and composition)

## How to use

```
# Confirm modules are present and loaded
modprobe mei_gsc && modprobe mei_pxp

# Verify HuC authenticated successfully
cat /sys/kernel/debug/dri/128/gt*/uc/huc_info
```
huc_info should show `status: RUNNING` and `HuC status: 0x00164001` (rather than INIT FAIL and 0x00164000).

## Testing done

No testing outside of github actions has been preformed. A conversation with @chewi "we can kick off a full build in Jenkins for you... but if the modules are there, I expect it will just work anyway. " has lead me to believe that this jenkins build should be sufficient for testing. If that is not the case, the only other testing plan that makes sense to me is to spin up a bare metal system with this os build and an arc gpu. (Let me know if yall have other thoughts on testing)


changelog/changes/2026-05-13-intel-mei-support.md
chewi, line 1: I'd prefer a more human entry that leaves out the config names in favour of what these features are actually used for.

tylerbucher: I wrote it to be as direct and concise as possible. But I will make it more readable as to what this change is supporting. Please let me know if the new entry is to long / wordy.

chewi: That's perfect, thanks! These entries appear on the web site for each release, and I think it's helpful for users to know what else they can do with Flatcar now. 😄

chewi:
I ran the GitHub Actions build (not Jenkins), which was enough to get an image with the modules and the change report. The change report is slightly broken (I suspected this before, now confirmed) as it doesn't show the size of new files. I grabbed the image and had a look though, and they're really small, so no issues there. Merging!

chewi (approved):

PR #4039, opened 2026-05-19
.github: Bump github actions

krnowak:
I don't know if we have self-hosted runners. If so, they would need to be updated before merging this PR.


chewi (approved):
I'm no GHA expert, but this looks harmless.

tormath1:
> I don't know if we have self-hosted runners. If so, they would need to be updated before merging this PR.

@krnowak why they would need to be updated? We have self-hosted runners but managed by CNCF.

(any chance to sign-off the commit as well?)

robinschneider:
Maybe we can add depedabot or renovate for this kind of updates?

PR #4040, opened 2026-05-20
Upgrade open-vm-tools in main from 13.0.10 to 13.1.0

flatcar-infra:
Subject says it all.


chewi:
I don't have time to look at this fully right now, but don't merge it as-is. GTK2 has been dropped and GTK4 has been added.

krnowak:
I think we should just use whatever Gentoo has - I think we already upstreamed all our modifications (in https://github.com/gentoo/gentoo/pull/41434), but it needs to be verified.

dongsupark:
All right, let's sync with Gentoo directly.
I have just manually disabled the open-vm-tools workflow to avoid further noise.

PR #4052, opened 2026-05-29
Deal with distfile requests being blocked by crates.io

chewi:
crates.io (on Amazon CloudFront) is now rejecting our requests with 403. This can be avoided by setting a custom user-agent for wget. I have made a similar change to upstream Portage, but let's use our own string for Flatcar.

While the user-agent change deals with the blocked requests, it turns out we should be using a different faster and less restricted endpoint anyway, so let's also take the (about to be merged upstream) cargo.eclass change as well.

## How to use

Try building a Rust-based package with `GENTOO_MIRRORS=""`.

## Testing done

[Jenkins](https://jenkins.flatcar.org/job/container/job/sdk/39/) got as far as the image job, which failed for an unrelated reason.

### DIFF
--- a/sdk_container/src/third_party/portage-stable/eclass/cargo.eclass
+++ b/sdk_container/src/third_party/portage-stable/eclass/cargo.eclass
@@ -280,7 +280,7 @@ _cargo_set_crate_uris() {
 			name="${BASH_REMATCH[1]}"
 			version="${BASH_REMATCH[2]}"
 		fi
-		url="https://crates.io/api/v1/crates/${name}/${version}/download -> ${name}-${version}.crate"
+		url="https://static.crates.io/crates/${name}/${name}-${version}.crate"
 		CARGO_CRATE_URIS+="${url} "
 
 		# when invoked by pkgbump, avoid fetching all the crates
--- a/setup_board
+++ b/setup_board
@@ -299,9 +299,7 @@ sudo_clobber "${BOARD_ETC}/portage/make.conf" <<EOF
 # Settings derived from the host environment
 CBUILD="$(portageq envvar CHOST)"
 HOSTCC="$(portageq envvar CHOST)-gcc"
-DISTDIR="$(portageq envvar DISTDIR)"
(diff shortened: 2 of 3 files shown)


John15321:
lgtm, i guess we should just make sure that the build goes thru and all good then

chewi:
> lgtm, i guess we should just make sure that the build goes thru and all good then

The relevant bit has passed. All the packages are built as binaries, so there are no more distfiles to download. If any of them had mismatched, it would have failed the hash checks.

John15321 (approved):
LGMT

John15321:
> > lgtm, i guess we should just make sure that the build goes thru and all good then
> 
> The relevant bit has passed. All the packages are built as binaries, so there are no more distfiles to download. If any of them had mismatched, it would have failed the hash checks.

Perfect then!

tormath1:
Cherry-picked to:
* flatcar-4694
* flatcar-4628
* flatcar-4593
* flatcar-4081

tormath1:
Hey @chewi I've backported this to LTS (flatcar-4081) but it [fails](https://jenkins.flatcar.org/job/container/job/image/154/consoleFull) with:
```
01:09:45  /build/amd64-usr/etc/portage/make.conf: line 9: FILE: unbound variable
```

Do you know who takes care of setting this `FILE` variable? Or we can we just avoid backporting the commit "make.conf: Set a custom user-agent for wget when fetching distfiles" ?

chewi:
I don't know what happened there, but you only need the cargo.eclass fix. The make.conf change is just an extra.

tormath1:
> I don't know what happened there, but you only need the cargo.eclass fix. The make.conf change is just an extra.

Ok good, thanks. I'll revert the "make.conf: Set a custom user-agent for wget when fetching distfiles" commit then. It's only for LTS, so that's "OK" in terms of consistency haha.

chewi:
For the record, I've realised it's probably because `portageq envvar -v` wasn't supported in the older version of Portage. I implemented that myself not so long ago.

PR #4071, opened 2026-06-03
Dockerfile.sdk-*: Disable Portage sandboxes during all image builds

chewi:
In #3955, I stopped disabling these sandboxes permanently, but they still need to be disabled temporarily during all the (unprivileged) image builds, not just Dockerfile.sdk-update.

## How to use

Check for `EPERM` warnings during the SDK build.

## Testing done

An SDK build in [Jenkins](https://jenkins.flatcar.org/job/container/job/sdk/50/cldsv/) made it to the packages job. I did find a single `EPERM` warning from Dockerfile.sdk-build when it breaks the dependency loops, but I stuck a `declare -p FEATURES` literally right before the emerge invocation, and it reported the expected value. It's not sudo masking the variable either. I can only guess that it's leftover state from /var/db/pkgs, which will go away with a further SDK build. One warning isn't a big deal anyway, and we were seeing a lot more than that before this change.

- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc. -- N/A

### DIFF
--- a/sdk_lib/Dockerfile.lean-arch
+++ b/sdk_lib/Dockerfile.lean-arch
@@ -5,7 +5,8 @@ ARG RMARCH
 ARG RMCROSS
 
 RUN if [ -n "$RMCROSS" ]; then \
-        sudo crossdev --clean --force "$RMCROSS"; \
+        FEATURES="-ipc-sandbox -network-sandbox -pid-sandbox" \
+            sudo crossdev --clean --force "$RMCROSS"; \
     fi
 
 RUN if [ -n "$RMARCH" ]; then \
--- a/sdk_lib/Dockerfile.sdk-build
+++ b/sdk_lib/Dockerfile.sdk-build
@@ -7,7 +7,8 @@ ARG OFFICIAL=0
 # mark build as official where appropriate
 RUN echo "export COREOS_OFFICIAL=$OFFICIAL" > /mnt/host/source/.env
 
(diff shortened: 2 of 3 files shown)


sdk_lib/Dockerfile.lean-arch
```diff
@@ -5,7 +5,8 @@ ARG RMARCH
 ARG RMCROSS
 
 RUN if [ -n "$RMCROSS" ]; then \
-        sudo crossdev --clean --force "$RMCROSS"; \
+        FEATURES="-ipc-sandbox -network-sandbox -pid-sandbox" \
+            sudo crossdev --clean --force "$RMCROSS"; \
```
krnowak, line 9: Won't sudo eat the env var set that way? For me, `FOO=bar sudo bash -c 'echo ${FOO}'` prints an empty line. But `sudo FOO=bar bash -c 'echo ${FOO}'` seems to do the trick. So should this rather be like this?


```suggestion
        sudo FEATURES="-ipc-sandbox -network-sandbox -pid-sandbox" \
            crossdev --clean --force "$RMCROSS"; \
```

chewi: We add `FEATURES` to `keep_env` in our sudoers.

chewi: If we didn't, this wouldn't be enough because the other instances call sudo within the scripts.

krnowak (approved):

PR #4074, opened 2026-06-04
Monthly GLSA metadata 2026-06-04

flatcar-infra:
Updated GLSA metadata


dongsupark (approved):
Looks good, as this PR is no-op in practice.
CI failures are not relevant.

PR #4077, opened 2026-06-08
sys-apps/ignition: Fix giving partition number 0 to get next available

chewi:
This also improves the earlier race condition fix. I have submitted the changes upstream in coreos/ignition#2234, but I have left out the partition deletion change as well as the test changes, since we don't run them.

## How to use

This resizes ROOT to make space for a fixed size FOO partition and a variable size VAR partition. VAR is mounted as /var. It fails on a current release. Unlike FCOS, we have to explicitly wipe ROOT here because although Flatcar attempts to resize ROOT, it only does this if the size can change, but it's already changed by the time that script runs. Maybe something to fix later, although I'm not sure how we can.

```yaml
variant: flatcar
version: 1.0.0
storage:
  disks:
  - device: /dev/vda
    wipe_table: false
    partitions:
    - number: 9
      label: ROOT
      size_mib: 512
      resize: true
    - number: 0
      label: FOO
      size_mib: 64
    - number: 0
      label: VAR
      size_mib: 0
  filesystems:
    - device: /dev/disk/by-partlabel/ROOT
      format: ext4
      wipe_filesystem: true
      label: ROOT
    - device: /dev/disk/by-partlabel/VAR
      format: ext4
      wipe_filesystem: true
      with_mount_unit: true
      path: /var
      label: VAR
```

## Testing done

A [Jenkins run](https://jenkins.flatcar.org/job/container/job/packages_all_arches/196/cldsv/) has...


krnowak (approved):
I reviewed the commits in the upstream PR, and stuff seems to be looking good.

PR #4083, opened 2026-06-10
run_local_tests: Clean up container after listing tests

zhijianli88:
The run_local_tests.sh script uses 'docker run' to execute 'kola list' and determine which tests to run. However, without the '--rm' flag, a stopped container instance is left behind after each execution.

Running the script multiple times leads to an accumulation of these exited containers, which consumes disk space and clutters the output of 'docker ps -a'.

Add the '--rm' flag to the 'docker run' command. This instructs Docker to automatically remove the container when it exits, ensuring no leftover artifacts. Since this container is only used temporarily to list tests, there is no need to preserve it.

## How to use

## Testing done


t-lo (approved):
Good catch - LGTM, thanks!

t-lo:
Only Devcontainer and OEM sysext tests failed; this is expected with PRs of forked repositories. Merging.

PR #3998, opened 2026-05-05
sys-apps/systemd: add upstream patch

tormath1:
This patch is a bit adapted from the upstream to remove kernel 6.15 limitation (`open_tree_attr`)

## Testing done

Locally built:
```
$ sudo emerge -v systemd
 * Applying 41875.patch ...                                                                                                                                                           [ ok ]
...
>>> Completed (1 of 1) sys-apps/systemd-257.9::portage-stable
```

- [x] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.
- [ ] CI: https://jenkins.flatcar.org/job/container/job/packages_all_arches/142/cldsv/

EDIT: For now it's failing:
```
May 05 12:56:08 localhost systemd[1]: Starting systemd-sysext.service - Merge System Extension Images into /usr/ and /opt/...
May 05 12:56:08 localhost (sd-merge)[1234]: Using extensions 'containerd-flatcar.raw', 'docker-flatcar.raw', 'oem-qemu.raw'.
May 05 12:56:08 localhost (sd-merge)[1234]: Failed to move mount /opt/foo to /run/systemd/sysext/submounts/opt/foo: Invalid argument
May 05 12:56:08 localhost systemd-sysext[1214]: Failed to merge hierarchies
May 05 12:56:08 localhost systemd[1]: systemd-sysext.service: Main process exited, code=exited, status=1/FAILURE
May 05 12:56:08 localhost systemd[1]: systemd-sysext.service: Failed with result 'exit-code'.
May 05 12:56:08 localhost systemd[1]: Failed to start systemd-sysext.service - Merge System Extension Images into /usr/ and /opt/.
```

Related to: https://github.com/flatcar/Flatcar/issues/2111

PR #4088, opened 2026-06-10
stable: bump to systemd-258

tormath1:
Right now, systemd-257 on Stable seems to create user issues with sysext + nested mounts. See: https://github.com/flatcar/Flatcar/issues/2111

An upstream patch has been created but once backported to systemd-257, it does not seem to solve the issue with systemd-257: https://github.com/flatcar/scripts/pull/3998#issue-4382900971

Current Beta, with systemd-258 seems to not have this issue but it ships the new confext feature which seems a bit young to be introduced on Stable so we can't yet promote a Beta with known issues. See: https://github.com/flatcar/Flatcar/issues/2166

Let's do a trade-off: keep current Stable (4593) but bump systemd to 258 with the upstream patch without the confext feature.

CI: https://jenkins.flatcar.org/job/container/job/packages_all_arches/182/cldsv/


tormath1:
Closing because not relevant as Beta will be promoted to Stable.

PR #4099, opened 2026-06-15
Update ca-certificates in main from 3.124 to 3.125

flatcar-infra:
Subject says it all.


chewi:
This is no longer needed. My direct push testing has already bumped it.

PR #4119, opened 2026-06-24
Bump edk2 to 202605, shim to 16.1, and refresh GRUB 2.12 against Fedora

chewi:
For GRUB, I am targeting Fedora 45 rather than 44 now. They don't seem to consistently publish their changes for upstream, and the 44 branch hasn't changed since December.

~~I had to fix an issue affecting a couple of Kola tests with Secure Boot enabled. I have submitted the fix [upstream](https://src.fedoraproject.org/rpms/grub2/pull-request/238).~~

Instead of applying the above fix, I have reverted the change that caused it, as recommended by upstream. We don't need the change anyway.

## How to use

See if it boots, particularly with Secure Boot enabled.

## Testing done

I've manually tested my EDK2 builds for all the supported architectures on Gentoo. I've also manually tested the Flatcar Secure Boot builds. A [Jenkins run](https://jenkins.flatcar.org/job/container/job/packages_all_arches/314/cldsv/) has totally passed.


sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-boot/grub/grub-2.12-verifier-null-file.patch
```diff
@@ -0,0 +1,36 @@
+https://src.fedoraproject.org/rpms/grub2/pull-request/238
+
+From db7782da011a39b06fe80282f608390ed76dae1c Mon Sep 17 00:00:00 2001
+From: James Le Cuirot <jlecuirot@microsoft.com>
+Date: Fri, 26 Jun 2026 14:08:05 +0100
+Subject: [PATCH] verifiers: Fix NULL pointer dereference when verification
+ fails on x64
+
```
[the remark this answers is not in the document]
chewi: This is nonsense.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
```diff
@@ -43,7 +43,7 @@ cros_post_src_install_sbat() {
 	insinto /usr/share/grub
 	newins - sbat.csv <<-EOF
 		sbat,1,SBAT Version,sbat,1,https://github.com/rhboot/shim/blob/main/SBAT.md
-		grub,4,Free Software Foundation,grub,${PV},https://www.gnu.org/software/grub/
+		grub,5,Free Software Foundation,grub,${PV},https://www.gnu.org/software/grub/
```
krnowak, line 46: Is this something we should be bumping on every update? If so, it may be a good idea to split it away into a separate variable that is defined at the top of the file (just like FLATCAR_VERSION).

sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-boot/grub/grub-2.12-00-redhat.patch
krnowak, line 1: Whoa, that patch blew up in size, didn't it?

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
```diff
@@ -7,6 +7,9 @@
 # Gentoo upstream package stabilisation
 # (the following packages are "unstable" upstream; we're stabilising these)
 
+# Chewi maintains edk2 in Gentoo, and we generally want the latest.
+sys-firmware/edk2-bin
+
```
krnowak, line 12: This is something I meant to ask you about - I normally keep entries in the file sorted alphabetically. Does keeping it sorted make sense? I guess that in Gentoo the order is more like date-based, no?

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
chewi: No, it's to do with fixed vulnerabilities, and there are strict rules around it. The short version is that we're copying [what Fedora has](https://src.fedoraproject.org/rpms/grub2/blob/rawhide/f/sbat.csv.in#_2).

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
chewi: Gentoo is date-based with newest at the top. I don't care too much. We can group some logically if that makes sense. I put this one at the top because it's permanent, and I don't want it getting in the way of the changing ones below.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
krnowak: Shouldn't then `grub.flatcar` below be bumped to 2? But I guess is something for revocation of some boot parts, so maybe not.

sdk_container/src/third_party/coreos-overlay/profiles/coreos/base/package.accept_keywords
krnowak: Alright, let's keep it like that. Maybe later I'll regroup these to permanent at the top and temporary below it.

krnowak (approved):

sdk_container/src/third_party/coreos-overlay/coreos/user-patches/sys-boot/grub/grub-2.12-00-redhat.patch
chewi: I must admit that I didn't look closely. I think Red Hat has backported a lot from 2.14 rather than just switching to 2.14. The ASN stuff looks rather big. We probably don't need much of this, but I don't want to start slicing and dicing here.

sdk_container/src/third_party/coreos-overlay/coreos/config/env/sys-boot/grub
chewi: [This](https://github.com/rhboot/shim/blob/main/SBAT.md) is how it works. I forget the details, but I think we only need to bump that if we introduce and then fix a vulnerability, either directly or through ingesting Red Hat's patches.

krnowak: Thanks for the link. So yeah, our line can stay as is, at 1.

PR #4134, opened 2026-07-07
build_library: Backport new initrd extraction script and adapt for zstd

chewi:
The older extraction script blew up on the latest kernel bump, seemingly because it is sensitive to cpio's handling of garbage data. The newer script handles this better, but I've had to adapt from xz back to zstd. For this latest bump, it now emits two rootfses when it should probably only emit one, but at least it doesn't blow up. It may be impossible to make this perfect, and it's only used for the difference report anyway.

CI: https://jenkins.flatcar.org/job/container/job/packages_all_arches/364/cldsv/


tormath1 (approved):
The CI job that was failing is now passing. Thanks!

PR #4136, opened 2026-07-08
bcachefs: enable in-tree module and add userspace-tools sysext

ananthb:
## Summary

Ships bcachefs on Flatcar as:

1. `bcachefs.ko` built in-tree with `coreos-modules` (`CONFIG_BCACHEFS_FS=m`, `CONFIG_BCACHEFS_POSIX_ACL=y`). The module is part of the base image and loads on demand via `modprobe`.
2. A `flatcar-bcachefs` sysext that carries the bcachefs userspace utilities (`bcachefs`, `mkfs.bcachefs`, `fsck.bcachefs`, `mount.bcachefs`, ...).

Users opt into the tools by writing `bcachefs` to `/etc/flatcar/enabled-sysext.conf`. The kernel module is available on every Flatcar image whether or not the sysext is merged.

## amd64-only

Keeping the sysext amd64-only until arm64 stabilizes. The kernel module itself is architecture-neutral in Kconfig, so it will also be available on arm64 images without any further config change.

## Test plan

- [ ] `emerge sys-kernel/coreos-modules` builds cleanly with the new config; `olddefconfig` doesn't downgrade `BCACHEFS_FS` to `n` (i.e., every `select`ed dep is satisfied by the existing config).
- [ ] The produced kernel image loads `bcachefs.ko` via `modprobe bcachefs` and `grep bcachefs /proc/filesystems` reports it.
- [ ] `USE=\"-modules\" emerge sys-fs/bcachefs-tools` produces `/usr/bin/bcachefs` and the compat symlinks without touching any kernel module tree.
- [ ] Merging the `flatcar-bcachefs` sysext exposes `mkfs.bcachefs`, `mount.bcachefs`, etc.
- [ ] `mkfs.bcachefs` on a loopback file + `mount -t bcachefs` round-trip.


ananthb:
Per the review request to add kola coverage like the ZFS sysext has, I opened flatcar/mantle#883 with `sysext.bcachefs.reboot` and `sysext.bcachefs.nfs` tests modeled on `kola/tests/sysext/zfs.go`.

Both PRs need to wait for Flatcar to promote Linux 6.18 as stable (so `bcachefs.ko` from `sys-kernel/coreos-modules` is in the base image). The mantle PR's `MinVersion` is a placeholder and should be bumped to whatever release first ships this.

PR #4171, opened 2026-08-02
ci-automation: verify and cache SDK image archives

satwiksps:
Prepared revision: [48a1cb551a](https://github.com/satwiksps/scripts/commit/48a1cb551a3b1b4727970bcfe171860305930895) on `repair/pr-4171`. The original branch is preserved so this closed PR can be reopened; its Files changed tab still shows the old revision. The description below covers the prepared repair.

Classic Docker and containerd image stores can report different IDs for the same saved image, so the SDK fallback downloads it again for each new container ([Flatcar#2086](https://github.com/flatcar/Flatcar/issues/2086)). Cache the archive's published SHA-512 together with the loaded local image ID, and check both before reusing it.

Verify the archive against its existing HTTPS `.DIGESTS` metadata before loading it. Keep metadata and archive on the same mirror, propagate load failures, and write the cache only after the short and registry-qualified tags point to the verified image. Gzip fallback applies when the zstd download is unavailable.

## How to use

From the repository root, run `bash test-container-download.sh` with Bash, jq, zstd, tar, sha512sum and flock available. The tests use local archive fixtures and mock Docker and HTTP operations.

## Testing done

- `bash -n ci-automation/ci_automation_common.sh test-container-download.sh` — passed.
- `git diff --check` — passed.
- `bash test-container-download.sh` on Ubuntu 24.04 — all 12 cases passed with real archive tools and mocked Docker/HTTP operations.
- `bash .repair-validation/sdk-live.sh "$PWD" "$RUNNER_TEMP/sdk-live"` — passed with Docker 29.8.0, using separate classic `overlay2` and containerd image stores in both directions. Checks covered repeated loads, missing or replaced tags, changed digests, corrupt archives, an injected load failure, retry, gzip fallback and concurrent calls.
- `bash .repair-validation/sdk-podman.sh "$PWD" "$RUNNER_TEMP/sdk-podman"` — Podman 4.9.3 archives loaded into both Docker stores, ran the expected payload, and skipped a second download.

The same classic-store archive produced ID `cc11ef4f4d7a...` on its producer and `42d1cddbb377...` on the containerd-store consumer. Reverse-direction IDs also differed. Both tags ran the expected payload after loading. These integration checks use real engines and local archive fixtures; HTTP transport and registry fallback discovery are stubbed.

[Validation logs and artifacts](https://github.com/satwiksps/scripts/actions/runs/34735477956/job/103666024402). The additional integration scripts are on a separate validation branch; the helper and regression-test blobs match this PR. A full SDK or Flatcar image build has not been run.

AI assistance was used for investigation, implementation and regression tests.

- [x] Changelog entry added.
- [ ] Inspected CI output for image differences.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,9 @@
 # Flatcar SDK tarballs
 *.tar.bz2
 
+# Record of the container images loaded from the buildcache
+*.id.local
+
 # SDK container env passing helpers
 sdk_container/.env
 sdk_container/.sdkenv
--- a/ci-automation/ci_automation_common.sh
+++ b/ci-automation/ci_automation_common.sh
@@ -216,36 +216,75 @@ function docker_commit_to_buildcache() {
 }
 # --
 
+# Prints the name:tag the image is available under locally, or nothing
+# if it is not present.
+function local_image_name() {
+    local name="$1"
+    local version="$2"
+
+    if image_exists_locally "${name}" "${version}" ; then
+        echo "${name}:${version}"
+    elif image_exists_locally "${CONTAINER_REGISTRY}/${name}" "${version}" ; then
+        echo "${CONTAINER_REGISTRY}/${name}:${version}"
+    fi
+}
+# --
+
+# Prints the ID of a local image, cutting the "sha256:" prefix that is
+# present in Docker but not in Podman.
+function local_image_id() {
+    $docker image inspect "${1}" | jq -r '.[].Id' | sed 's/^sha256://'
+}
+# --
+
+# Prints the image ID published in the buildcache, or "not found".
+function buildcache_image_id() {
+    local name="$1"
+    local version="$2"
+    local id_file="${name}-${version}.id"
+
+    curl --fail --silent --show-error --location --retry-delay 1 \
+        --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 \
+        "https://${BUILDCACHE_SERVER}/containers/${version}/${id_file}" \
+        || curl --fail --silent --show-error --location --retry-delay 1 \
+        --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20 \
+        "https://mirror.release.flatcar-linux.net/containers/${version}/${id_file}" \
+        || echo "not found"
+}
+# --
+
 function docker_image_from_buildcache() {
     local name="$1"
     local version="$2"
     local compr="${3:-zst}"
     local tgz="${name}-${version}.tar.${compr}"
-    local id_file="${name}-${version}.id"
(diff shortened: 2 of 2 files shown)


zhijianli88:
> But a docker save / docker load round-trip can produce a different image ID on the consumer's Docker daemon (e.g., due to different Docker versions or storage drivers between CI and the developer's machine)

Do you have any more concrete evidences? 

<.id> is used to verify the image integrity, if the <.id> is untrustable, I'd like to choose another method

John15321:
@zhijianli88 We are currently experiencing a lot of AI contributions, if you see anything like that just close it

satwiksps:
@zhijianli88 @John15321  this is documented upstream behavior: [moby/moby#51934](https://github.com/moby/moby/issues/51934).
few months ago I came across above docker issue and learnt this when a maintainer there said ["The ID will be persistent only if you make sure that both hosts have the same storage driver."](https://github.com/moby/moby/issues/51934#issuecomment-3805844114)

IMO since flatcar CI and developer machines can have different backends, the .`id `from bincache won't always match the local image after docker load, even though the content is the same. I thought I could fix it in this way, there could be other way too like comparing layer digests which are consistent across storage backends ig...so yeah I gave the issue flatcar/Flatcar#2086 a go, my fix caches the locally loaded ID after docker load.

EDIT: The second approach i.e. remote .id is fetched on every run again and stays what we compare against, I think this would still re-download sdk images after docker load. Hence closing this PR, would create another PR if I find a better working solution for this. 
I will reopen this PR if I find a better solution.

zhijianli88:
> @zhijianli88 @John15321 this is documented upstream behavior: [moby/moby#51934](https://github.com/moby/moby/issues/51934). few months ago I came across above docker issue and learnt this when a maintainer there said ["The ID will be persistent only if you make sure that both hosts have the same storage driver."]

Make sense.

However, the .id is used to ensure image integrity and your patch doesn't address this critical aspect

satwiksps:
> However, the .id is used to ensure image integrity and your patch doesn't address this critical aspect

My prev approach cached the loaded ID locally and returned early on it. I understood that once a local copy existed, the remote .id was never looked at again. Rebuilt image on bincache would have gone unnoticed

In my second approach the remote .id is fetched on every run again and stays what we compare against. After a load I record two values in <name>-<version>.id.local, the published ID we loaded from & the ID our own daemon gave the image. it only treated as current when both still match, so the published ID still has to agree (no silent rebuild slips through) and the local ID catches the image being replaced underneath us. image_id == remote_id still runs first, so nothing changes where the IDs already line up. Only assumption dropped is that docker load reproduces the ID docker image inspect printed in CI, which is the part that doesn't hold across storage drivers. I hope this covers the integrity side properly now

PR #4053, opened 2026-06-01
Add options to enable /dev/kfd on amdgpu

mrbojangles3:
# Add amdgpu /dev/kfd 
The amdgpu driver has two parts, one part is for the video use case,`/dev/dri/card0`, the other is for the compute use case,`/dev/kfd`. AMD has some [documentation](https://rocm.docs.amd.com/projects/install-on-linux/en/latest/how-to/docker.html#without-the-amd-container-toolkit) on it. 

Currently the `amdgpu` driver is only exposing the `/dev/dri/card0` file descriptor. 

This PR is related to https://github.com/flatcar/Flatcar/issues/1935 where the driver was introduced.

My specific hardware:
`ASUSTeK COMPUTER INC. MINIPC PN50/PN50, BIOS 0626 08/31/2023`
```
04:00.0 VGA compatible controller: Advanced Micro Devices, Inc. [AMD/ATI] Renoir [Radeon Vega Series / Radeon Vega Mobile Series] (rev c1) (prog-if 00 [VGA controller])
	Subsystem: ASUSTeK Computer Inc. Device 87e7
	Flags: bus master, fast devsel, latency 0, IRQ 34, IOMMU group 5
	Memory at d0000000 (64-bit, prefetchable) [size=256M]
	Memory at e0000000 (64-bit, prefetchable) [size=2M]
	I/O ports at e000 [size=256]
	Memory at fe600000 (32-bit, non-prefetchable) [size=512K]
	Capabilities: <access denied>
	Kernel driver in use: amdgpu
	Kernel modules: amdgpu
```

These changes will enable the compute functionality on all AMD GPUs, not mine specifically. 

## How to use

load the amdgpu driver and verify the /dev/kfd device is present. Confirm that the gpu is visible with rocminfo.

## Testing done

I have built an image and loaded it onto my machine. ROCm info reports the correct info. I feel that it should be noted that 

```
core@host ~ $ uname -a
Linux host.yallywood.com 6.12.87-flatcar #1 SMP PREEMPT_DYNAMIC Fri May 29 02:41:36 -00 2026 x86_64 AMD Ryzen 7 4800U with Radeon Graphics AuthenticAMD GNU/Linux
core@host ~ $ cat /etc/os-release 
NAME="Flatcar Container Linux by Kinvolk"
ID="flatcar"
ID_LIKE="coreos"
VERSION="4694.0.0+2026-05-29-1414"
VERSION_ID="4694.0.0"
BUILD_ID="2026-05-29-1414"
[...]

PR #4191, opened 2026-08-06
Update Linux to 6.18

sayanchowdhury:
## Testing done

CI: https://jenkins.flatcar.org/job/container/job/packages_all_arches/490/

This might help: https://github.com/flatcar/Flatcar/issues/2424


sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/commonconfig-6.18
```diff
@@ -128,6 +128,7 @@ CONFIG_CPU_FREQ_GOV_USERSPACE=m
 CONFIG_CPU_FREQ_STAT=y
 # CONFIG_CROSS_MEMORY_ATTACH is not set
 CONFIG_CRYPTO_AES=m
+CONFIG_CRYPTO_CRC32C=y
```
dongsupark, line 131: Why is the kernel config CRYPTO_CRC32C is needed for building old nvidia drivers?

sdk_container/src/third_party/coreos-overlay/x11-drivers/old-nvidia-drivers/files/0001-mm-use-vm_flags_reset-to-avoid-GPL-only-vma_start_wr.patch
```diff
@@ -0,0 +1,155 @@
+From d1154945f8b2d22392d30a5e522612dfbcf17cff Mon Sep 17 00:00:00 2001
```
dongsupark, line 1: Huge thanks for working on old nvidia drivers.
When I was recently touching those, I was wondering if it would worth keeping this one, because of a huge amount of build issues with newer Kernel versions. Ideally we should deprecate them...? ;-)

dongsupark (commented):
I think we should run Jenkins CI for all providers, especially for such a big Kernel bump.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.18
```diff
@@ -0,0 +1,140 @@
+# CONFIG_ACPI_AC is not set
```
ader1990, line 1: the *-6.12 files from this folder need to be removed, same with files from the coreos-sources folder.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/files/6.18/z0007-tools-hv-fix-cross-compilation-for-ARM64.patch
```diff
@@ -0,0 +1,35 @@
...
+Date: Thu, 19 Sep 2024 07:59:59 +0000
+Subject: [PATCH 7/8] tools: hv: fix cross-compilation for ARM64
+
+---
+ tools/hv/Makefile | 4 +++-
+ 1 file changed, 3 insertions(+), 1 deletion(-)
+
+diff --git a/tools/hv/Makefile b/tools/hv/Makefile
+index 34ffcec264ab..7a7ece3cdf03 100644
+--- a/tools/hv/Makefile
```
ader1990, line 12: There were some new patches recently, maybe we can reuse the official one instead of this one: https://github.com/torvalds/linux/commit/ca5ee0e918115fb5cf626d75461d9fca06e06caf

sdk_container/src/third_party/coreos-overlay/x11-drivers/old-nvidia-drivers/files/0001-mm-use-vm_flags_reset-to-avoid-GPL-only-vma_start_wr.patch
sayanchowdhury: Yes, The problem is we don't have a deprecation plan with the older-sysext. We should urgently start discussing this as because we will start piling old versions on our end and it quite difficult to maintain.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-sources/coreos-sources-6.18.45.ebuild
```diff
@@ -0,0 +1,45 @@
...
+
+# -rc releases should be versioned L.M_rcN
+# Final releases should be versioned L.M.N, even for N == 0
+
+# Only needed for RCs
+K_BASE_VER="5.15"
+
+inherit kernel-2
+EXTRAVERSION="-flatcar"
+detect_version
```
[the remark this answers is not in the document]
chewi: Please keep this as it was. This stuff is hairy. I have a branch that renames the kernel from "coreos" to "flatcar" if we eventually decide to do that.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.18
```diff
@@ -119,7 +105,6 @@ CONFIG_TCG_INFINEON=m
 CONFIG_TCG_NSC=m
 CONFIG_UPROBE_EVENTS=y
 CONFIG_USB_OHCI_HCD_SSB=y
-CONFIG_VBOXGUEST=m
```
chewi, line 122: Why? We explicitly include that in bootengine.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/amd64_defconfig-6.18
chewi, line 1: Most of the options dropped here were added recently in #4053.

sdk_container/src/third_party/coreos-overlay/x11-drivers/old-nvidia-drivers/old-nvidia-drivers-550.163.01-r2.ebuild
```diff
@@ -148,12 +153,36 @@ src_prepare() {
 	rm nvidia-xconfig && mv nvidia-xconfig{-${PV},} || die
 	mv NVIDIA-kernel-module-source-${PV} kernel-module-source || die
 
+	# Linux 6.15 removed EXTRA_CFLAGS support for out-of-tree kernel modules
+	# (upstream commit b2c885b9). NVIDIA 550 uses EXTRA_CFLAGS in its Kbuild
+	# files to pass -I$(src)/common/inc, so headers like os-interface.h,
+	# nv-firmware.h, and nv-pci-types.h are never found at compile time on
+	# kernels >= 6.15. Replace with ccflags-y, the correct variable.
+	find "${S}" \( -name 'Kbuild' -o -name 'Makefile' \) \
+		-exec sed -i 's/\bEXTRA_CFLAGS\b/ccflags-y/g' {} + || die
```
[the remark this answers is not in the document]
chewi: Nonsense. I've tested it, either works.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/commonconfig-6.18
```diff
@@ -168,16 +171,15 @@ CONFIG_DM_UEVENT=y
 CONFIG_DM_VERITY=m
 CONFIG_DM_ZERO=m
 CONFIG_DNS_RESOLVER=y
-CONFIG_DRM=y
+CONFIG_DRM=m
```
chewi: I wondered why this, `CONFIG_DRM_VIRTIO_GPU`, and `CONFIG_VIRTIO_CONSOLE` had changed. @ader1990 changed them from `m` to `y` back in #2359. He didn't explicitly say why, but I guess having them in the initrd leaves you blind if the system fails to boot that far.

chewi (changes requested):

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/commonconfig-6.18
ader1990: These were changed to y so that the initrd stage can produce output - See https://github.com/flatcar/Flatcar/issues/1553. We could have manually loaded the modules from the bootengine, as we now have access to all the kernel .ko-s, I suppose it is a matter of preference maybe.

ader1990: If you put them back to m, the issue needs to be again reopened, unless we manually install / load them from the bootengine code?

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/commonconfig-6.18
```diff
@@ -251,7 +248,7 @@ CONFIG_HWMON=m
 CONFIG_HW_RANDOM=y
 CONFIG_HW_RANDOM_TIMERIOMEM=m
 CONFIG_HW_RANDOM_VIRTIO=y
-CONFIG_HYPERV=m
+CONFIG_HYPERV=y
```
ader1990: Does this have a reason to be enabled? asking as a curiosity - if this is enabled, why not enable (Y) all of the Hyper-V modules?

We can get rid of this line as well https://github.com/flatcar/bootengine/blob/7727ec78da72e700e8fa1ce2144cb2476448d186/dracut/53ignition/module-setup.sh#L173.

sdk_container/src/third_party/coreos-overlay/sys-kernel/coreos-modules/files/commonconfig-6.18
chewi: I _think_ they would automatically load as modules, but as I said, if it fails to get that far, you won't see anything.

PR #4197, opened 2026-08-09
coreos-modules: enable the Bluetooth subsystem

ananthb:
This PR enables the Bluetooth stack as modules.

## What's enabled

Core stack (BR/EDR + LE) plus some transports.

## Userspace

Pulling BlueZ into the base image would grow every Flatcar image for a feature many users won't use. I've added a bluez sysext instead: [flatcar/sysext-bakery#253](https://github.com/flatcar/sysext-bakery/pull/253).


tormath1 (commented):
I'm curious, what would be the use case for this?

ananthb:
I've got a few automation devices in my homelab datacenter that can be controlled over BLE like the AC units and access control locks. I've been migrating my servers over to Flatcar and I'd like to control those with services running on Flatcar.

robinschneider:
> I've got a few automation devices in my homelab datacenter that can be controlled over BLE like the AC units and access control locks. I've been migrating my servers over to Flatcar and I'd like to control those with services running on Flatcar.

You can also use one or more Bluetooth proxies for that, this also allows you to be closer to the Bluetooth devices.
For Home assistant there is the ESPHome Bluetooth Proxy for this for example.

ananthb:
Yeah that is what I'm doing now. I can close this PR if flatcar doesn't want to support bluetooth.

tormath1:
> Yeah that is what I'm doing now. I can close this PR if flatcar doesn't want to support bluetooth.

I'm fine with this as long as it does not increase a lot the kernel size. 

Right now it fails to build with:
```
 * ERROR: sys-kernel/coreos-modules-6.12.102::coreos-overlay failed (configure phase):
 *   Requested options not enabled in build:
 *     CONFIG_BT_HCIUART_BCM
 * 
 * Call stack:
 *     ebuild.sh, line  143:  Called src_configure
 *   environment, line 1768:  Called coreos-kernel_src_configure
 *   environment, line  822:  Called die
 * The specific snippet of code:
 *           die "Requested options not enabled in build:
 * ${missing}";
```

ananthb:
It won't meaningfully move the size of the kernel by much. I'll take a look at that build failure.

PR #4198, opened 2026-08-09
coreos/sysext/containerd: configure registry certs.d

### LINKED ISSUE #583: [RFE] Include certs.d registry plugin in containerd config.toml
## Current situation

I currently customize my containerd config.toml just to add the following settings:

```
[plugins."io.containerd.grpc.v1.cri".registry]
config_path = "/etc/containerd/certs.d"
```

My testing shows that containerd does not care if the directory does not exist, but I certainly would welcome further testing.

## Impact

I have to add a custom containerd config that might not keep compatibility with upstream, simply so I can add custom registries or proxies. 

## Ideal future situation

To have a way of customizing a section of containerd's config.toml without replacing the whole file, which then also requires a dropin for the containerd service.

Either to have the plugin registry snippit added to the main config, or 
## **Implementation options
(tested) add the following snippet to config.toml, which does not seem to care if the directory does not exist:

```
[plugins."io.containerd.grpc.v1.cri".registry]
config_path = "/etc/containerd/certs.d"
```

(untested) to have an imports section that points to a custom config area such as:

```
imports = ["/etc/containerd/runtime_*.toml"]`
```

DUKartik:
Fixes flatcar/Flatcar#583
## What does this PR do?

Configure containerd to load registry configuration from
`/etc/containerd/certs.d`.

This adds the following registry configuration to both Flatcar
containerd configuration variants:

```toml
[plugins."io.containerd.grpc.v1.cri".registry]
config_path = "/etc/containerd/certs.d"


DUKartik:
Hi @sayanchowdhury ,
When you have a moment, could you kindly review and approve this PR so the workflows can proceed? I’d really appreciate your time and support.

PR #4210, opened 2026-08-12
fix: defer variable expansion in gc-sdk aliases

Harsh-2005d:
# FIX
Wrap aliases in single quotes to evaluate $PWD and $HOME at execution time rather than when the script is sourced.

## How to use

## Testing done


chewi (approved):
Good catch. I think `HOME` is unlikely to be an issue in practise, but it should be quoted. Much more importantly though, `PWD` will almost certainly be wrong without this fix.

chewi (changes requested):
Wait, a stray ` has snuck in at the end.

Harsh-2005d:
Sorry for that. Fixed.

chewi (approved):

PR #4212, opened 2026-08-12
fix: inner loop variable in ci-automation

Harsh-2005d:
# Fix

Renamed the inner loop iterator from `file` to `tmp_file`. 

## How to use

## Testing done


krnowak:
Thanks for the PR, but this isn't fixing anything really. The file variable gets reused in the inner loop, because its value from outer loop isn't important anymore.

Harsh-2005d:
I thought it would be a good practice to not reuse it.

PR #4216, opened 2026-08-14
sys-libs/glibc: bump to 2.44 (CVE-2026-5435, 5450, 5928, 6238, 6368, 6791)

imshubham22apr-gif:
This pull request updates the GNU C Library (sys-libs/glibc) in our portage-stable overlay from version 2.43-r2 to 2.44. Bumping to 2.44 is a security-focused update that resolves six critical vulnerabilities, including a high-severity heap buffer overflow in the scanf parser, a stack exhaustion bug during tilde expansions, a wide-character memory leak in ungetwc, and various out-of-bounds read/write flaws in resolver packet handling routines. This change introduces the new ebuild file and adds the verified source archive checksums to the Manifest file, keeping our system aligned with Gentoo upstream stability. All architecture keywords for Flatcar (~alpha, amd64, arm, arm64, etc.) are maintained, and the pristine source package configuration runs cleanly within our containerized Portage SDK environment.

To review and validate this PR, reviewers need to perform a package rebuild within the Flatcar SDK container. Enter the container shell and run the emerge command "emerge-gitclean sys-libs/glibc" to verify that the ebuild parses correctly, downloads the validated source tarball from the mirrors, compiles cleanly, and installs into the build root.

We have done local testing before submitting this PR. First, we ran a Python command to fetch the source tarball from the GNU upstream server and compute its sizes and checksums. The script output confirmed the size is 20,620,544 bytes, the SHA512 hash matches the official release, and the BLAKE2b hash starts with f943e3. Second, we ran "git diff HEAD~1..HEAD --name-status" to inspect the workspace changes, and the command verified that only the Manifest was modified and the new glibc-2.44.ebuild was added. Finally, we confirmed that all changelog entries have been added to the respective changelog directory, and the CI output is ready to be inspected for any boot and use package size differences or missing binaries.


chewi:
No. Gentoo hasn't even keyworded 2.44 yet, let alone stabilised it. We would never make that jump ahead of them. I expect sufficiently serious security fixes will be backported by Gentoo or glibc upstream.

chewi:
Also, portage-stable needs to stay synced to Gentoo. Any changes here would be lost.

PR #4218, opened 2026-08-14
sysexts: Drop some more cruft from sysext images

krnowak:
GCE and AMI OEM images still had debug directories. Incus is about to get an update where two packages (blake3 and xdelta) are installing some cmake stuff.

CI: https://jenkins.flatcar.org/job/container/job/packages_all_arches/543/


chewi (approved):
Cool, although we should probably unify this stuff. It's almost the same in every case.

krnowak:
> Cool, although we should probably unify this stuff. It's almost the same in every case.

Some day. ;)

PR #4221, opened 2026-08-19
Add better handling for the azure marketplace updates

sayanchowdhury:
This commit fixes:
- Add better parsing for the LTS, earlier this was broken and the LTS release was not found in the list of supported channels.
- Update the account_name to flatcar0001
- We hitting the 100 limit for the Azure images no on a regular basis so added a threshold to remove the images when the number is higher than 95
- Add the option to delete the draft versions which is quite difficult to do manually as it involves time and effort and the script works a bit differently where it handles channel across plans where if any thing needs to rectified then all the plans needs to be visited and handled manually. The commit eases the process a bit
- Add a dry run mode to check if everything will be well run
- Formatting using black

AI Usage:
- Used to find the endpoints of the Microsoft API-
- Used to add comments & helpers.
- Implement the `dry-run` feature.

### DIFF
--- a/ci-automation/release/azure_marketplace_publish.py
+++ b/ci-automation/release/azure_marketplace_publish.py
@@ -6,6 +6,7 @@
 # ]
 # ///
 import os
+import sys
 import copy
 import json
 import argparse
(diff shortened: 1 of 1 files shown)


chewi (commented):
Seems okay, but shouldn't it deprecate more than one at a time? You might still be over the threshold after running it.

sayanchowdhury:
I'm just deprecating one version at a time, because that's how much we need and what I've been doing manually.

I've just replicated what I've been doing manually

chewi (approved):
I think Copilot may have raised some valid concerns, but I trust your judgement here.

PR #4159, opened 2026-07-29
Drop net-misc/bridge-utils

Alivestars04:
`bridge-utils` has been marked deprecated by its upstream maintainers in favor of the `bridge` utility from iproute2, which Flatcar already ships via `sys-apps/iproute2` (currently 6.16.0). Keeping a deprecated package in the image means carrying two Gentoo patches (a musl fix and an AR-variable fix) and an ebuild for tooling that upstream no longer recommends.

This removes the package from `portage-stable`, drops it from the `coreos-base/coreos` package list, and removes it from the `portage-stable-packages-list` used by the automated package update workflow, so the updater stops tracking it. The `coreos-base/coreos` revision symlink is bumped from `-r320` to `-r321` so the package is rebuilt.

This is a user-visible change: `brctl` will no longer be present in the image. Users invoking `brctl` in scripts or Ignition configs need to switch to `ip link` / `bridge`, e.g. `ip link add br0 type bridge` instead of `brctl addbr br0`. Given that, maintainers may prefer to time this against a specific release — happy to hold or re-target the branch.

Closes flatcar/Flatcar#638

## How to use

Reviewers can confirm the removal is complete and correctly scoped:

1. No live references to the package remain in the tree (only the historical 2022 changelog entry, intentionally left as a release record):
   ```
   git grep -n bridge-utils
   ```
2. The replacement is present in the same image package list:
   ```
   git grep -n 'sys-apps/iproute2' sdk_container/src/third_party/coreos-overlay/coreos-base/coreos/coreos-0.0.1.ebuild
   ```
3. On a built image, verify `brctl` is gone and `bridge`/`ip link` still create bridges:
   ```
   command -v brctl            # expected: not found
   ip link add br0 type bridge && bridge link show && ip link del br0
   ```

## Testing done

Verified statically in the tree; I have not built an SDK image locally, so the image-level checks below are left to CI.

- Confirmed the package existed only at `sdk_container/src/third_party/portage-stable/net-misc/bridge-utils/` and was referenced in exactly two live places (`coreos-base/coreos/coreos-0.0.1.ebuild` and `.github/workflows/portage-stable-packages-list`):

  ```
  $ git grep -n bridge-utils
  sdk_container/src/third_party/portage-stable/changelog/updates/2022-02-23-update-pkgs-2015.md:1:- bridge-utils ([1.7.1](...))
  sdk_container/src/third_party/portage-stable/net-misc/bridge-utils/metadata.xml:12: ...
[...]

PR #4241, opened 2026-08-27
changelog: add links to bridge-utils, fix file name of weekly updates

dongsupark:
Add a link to the original issue https://github.com/flatcar/Flatcar/issues/638 in the changelog missing in from https://github.com/flatcar/scripts/pull/4159.

Fix file name of weekly updates changelog from https://github.com/flatcar/scripts/pull/3174.


dongsupark:
Oh, the previous Copilot [comment](https://github.com/flatcar/scripts/pull/4241#pullrequestreview-5039568349) is not valid any more, because that was based on the commit including my file name mistake.
Now I have fixed the changelog file name from 2026 to 2025. (Yes, 2025 is the correct one)

changelog/updates/2025-08-06-weekly-updates.md
tormath1, line 1: Thanks for the heads-up, to me we were still in 3035.

tormath1 (approved):

krnowak:
Won't the rename cause the release automation to pick up the entries from the renamed file, because the timestamp of the file has changed?

dongsupark:
> Won't the rename cause the release automation to pick up the entries from the renamed file, because the timestamp of the file has changed?

Interesting question, so I have tested this branch with [show-changes](https://github.com/flatcar/flatcar-build-scripts/blob/main/show-changes), but show nothing. That has something to do with how the script is implemented.

dongsupark:
TL;DR we do not have to worry about such a regression in release notes.

What `show-changes` calls is not a normal `git diff` but a slightly different one, `git difftool ...`. 
You can test it with `git difftool --no-prompt --extcmd='sh -c "if [ \"$LOCAL\" = /dev/null ] ; then cat \"$REMOTE\" ; fi"' main..HEAD -- "changelog/updates/"`, then it shows nothing.
On the other hand, `git diff main..HEAD changelog/updates` would print out a file rename change.
See also the long comment in `show-change`, https://github.com/flatcar/flatcar-build-scripts/blob/4dd31990ce7d60fcc1d8829997b0b290c0865e06/show-changes#L80.

PR #4252, opened 2026-09-01
 Build sysext DDIs with a 4K sector size

danzatt:
# Build sysext DDIs with a 4K sector size

systemd-repart's --make-ddi auto-sizes the image to the exact minimum. Since systemd v261.2 the tail GPT reserve is (16KiB + one sector) and is no longer rounded up, so with the default 512-byte sector the total image size is not a multiple of 4096 which breaks update_engine's delta_generator. Pass an explicit --sector-size=4096 so that repart
produces 4k-aligned sysexts.

## How to use 

Rebuild flatcar and check all sysexts (especially overlaybd) have sizes aligned to 4K

## Testing done

/cc @tormath1

### DIFF
--- a/build_sysext
+++ b/build_sysext
@@ -331,6 +331,7 @@ else
 fi
 
 systemd-repart \
+  --sector-size=4096 \
   --private-key="${SYSEXT_SIGNING_KEY_DIR}/sysexts.key" \
   --certificate="${SYSEXT_SIGNING_KEY_DIR}/sysexts.crt" \
   --make-ddi=sysext \


tormath1 (approved):
LGTM. It has been tested as part of a 261.2 upgrade (https://github.com/flatcar/scripts/commit/d4baddf125cd25c81112ec8085737c5640b59cd1) - and it built correctly: https://jenkins.flatcar.org/job/container/job/packages_all_arches/608/cldsv/

chewi (approved):
We're lucky it even has such an option!

danzatt:
I just checked if we shouldn't also set `--grain-size` as in @tormath1's dev build, but it seems it defaults to `MAX(4096, sector_size)`, so there's no need to set it explicitly (unless repart decides to change something under the hood again 😆).

tormath1:
> I just checked if we shouldn't also set `--grain-size` as in @tormath1's dev build, but it seems it defaults to `MAX(4096, sector_size)`, so there's no need to set it explicitly (unless repart decides to change something under the hood again 😆).

Yeah, I was curious to see if it was fixing the issue. :)

PR #4264, opened 2026-09-07
changelog: Add security entry to the Ignition 2.27.0 update

chewi:
Just a changelog entry for flatcar/Flatcar#2167.


krnowak (approved):
I wonder if the clause below will work as a part of the review message.

Closes https://github.com/flatcar/Flatcar/issues/2167

